PR #3, opened 2019-01-28
Prototyping

DavideD:
Should we merge this?
This way if somebody is interested can start to play with it


.gitignore
```diff
@@ -4,6 +4,9 @@
 # Gradle work directory
 .gradle
 
+# Gradle cache
+.gradletasknamecache
```
Sanne: interesting, I never had this file. What creates this ? (hint: not Gradle)

DavideD: I think it's an intellij file. But I haven't checked.

PR #4, opened 2019-01-28
Introduce a draft for a Query API

Sanne:


DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/4

gavinking:
I know this is a draft but just wanted to note down somewhere that we still need (at least)

- `get()`, 
- `lock()`, and
- `refresh()`.

Also, I think we will *eventually* want to provide a way to batch up multiple ids, eg:

- `remove(id1, id2)`
- `persist(id1, id2)`
- `load(id1, id2)`

Which can all be optimized to a single SQL query with `where id in ...`.

Sanne:
Hi @gavinking ! Yes absolutely. I just picked some simple ones so that we can start focusing on the implementation first.

Granted queries aren't "simple" but I decided to include them already as they are a bit special in regards to loading the stream and controlling the flow of data.

Multi-load is something `Session` provides nowadays; +1 to have it too, and we'll likely want to mimick the existing API.

gavinking:
@Sanne yes yes, of course.

PR #5, opened 2019-01-30
Update the API and, add licenses and split in submodules

DavideD:
I've applied the changes we discussed.

@Sanne I left the remove method so that it returns a `CompletionStage<Void>` and the test are partially commented for now because I haven't to added vertx-unit yet


Sanne:
great! thanks

PR #6, opened 2019-01-31
Reactive testing using Junit 5 and Vert.x Junit5

DavideD:
Initial experiments testing with Junit5 + Vert.x Junit5


Sanne:
looks nice, and works fine in command line. I have some problems in IDEA though; in particular it looks like unable to resolve `io.vertx:vertx-junit5:3.6.2` .. not sure why.

DavideD:
I think it happens because Intellij Caches somewhere the previous version of Junit when you launched the test before.

Try this in IntelliJ: Run -> Edit Configurations -> Remove all the entries under JUnit on the left

After that the issue should disappear

Sanne:
thanks, that worked!

PR #7, opened 2019-09-19
Find, insert, update, delete

DavideD:
Initial work for CRUD operations using the RxSession.

Still a prototype and the code and commits are not that clean (sorry) but should be a good starting point for discussing how to move on.
It only works for a simple entity (no associations, no transactions, single table, no cascading, ... )

Tests require a PostgreSQL running,  this is the command using podman (should work with docker as well):
```
sudo podman run --ulimit memlock=-1:-1 -it --rm=true --memory-swappiness=0 --name PostgreSQL -e POSTGRES_USER=hibernate-rx -e POSTGRES_PASSWORD=hibernate-rx -e POSTGRES_DB=hibernate-rx -p 5432:5432 postgres:10.5
```

@FroMage Could you have a look at ReactiveSessionTest and the code in general and check that I didn't break any of your changes (or mostly that what I've changed make some sense), please?

@Sanne If you want to have a look I think this might be a good starting point.

Thanks


DavideD:
By the way, I think there is a transaction that doesn't get closed somewhere (I can see it form the Postgre Server log). If you have any idea on when that happens, it would be nice.

hibernate-rx-core/src/test/java/org/hibernate/rx/ReactiveSessionTest.java
```diff
@@ -102,12 +101,14 @@ public void init() {
 		session = sessionFactory.unwrap( RxHibernateSessionFactory.class ).openRxSession();
 	}
 
+	@After
+	public void tearDown() {
+		dropTable().whenComplete( (r, e) -> { sessionFactory.close(); } );
```
FroMage: If this is async you probably need to tell the test context about it.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/impl/RxSingleTableEntityPersister.java
```diff
@@ -301,7 +298,7 @@ private void preInsertInMemoryValueGeneration(
 		return insertStage;
 	}
 
-	public CompletionStage<RowSet> insertRx(
+	public CompletionStage<?> insertRx(
```
FroMage: If this returns a `CS<Integer>` why use `?`?

DavideD: I'm still not sure it should return an integer or a specific class. I didn't want to commit to something at this stage. Mostly because it makes it easier to change thing at this stage.

hibernate-rx-core/src/test/java/org/hibernate/rx/ReactiveSessionTest.java
DavideD: The thing is that this is not a test. It doesn't seem to work if I use TestContext like in the other tests. Is there a proper way to do it in the `@After` methods?

FroMage:
Transactions are automatically closed when you use the query methods on the pool. if you access the query method on the connection then you're responsible for closing it.

BTW, if you ever find that you have an async operation that's giving you trouble and you don't know who started it, use https://github.com/FroMage/quarkus/blob/panache-rx/extensions/panache/rx-panache/runtime/src/main/java/io/quarkus/panache/rx/runtime/RxOperations.java#L93

In hibernate-rx you will want to work not on the pool, but on a client, which can be either a connection or a transaction.

In Panache RX I use `PgClient` as given to me by the container: https://github.com/FroMage/quarkus/blob/panache-rx/extensions/panache/rx-panache/runtime/src/main/java/io/quarkus/panache/rx/runtime/RxOperations.java#L147

And then this class gives me a `PgClient` that is either transaction-scoped, and a PG transaction and part of the JTA transaction, or a request-scoped `PgConnection`: https://github.com/FroMage/quarkus/blob/panache-rx/extensions/panache/rx-panache/runtime/src/main/java/io/quarkus/panache/rx/runtime/client/TransactionScopedPgClient.java

This is really a copy of that the ORM guys did for Quarkus Hibernate ORM.

hibernate-rx-core/src/test/java/org/hibernate/rx/ReactiveSessionTest.java
FroMage: @vietj says it should work, use the `TestContext` like in the other test methods.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/impl/RxSingleTableEntityPersister.java
FroMage: OK

hibernate-rx-core/src/test/java/org/hibernate/rx/ReactiveSessionTest.java
vietj: it should work given you provide a `TestContext` argument in the method

FroMage: This is a vertx-junit test, so it _is_ supported:

```java
@After
    public void after(TestContext context) {
      vertx.close(context.asyncAssertSuccess());
    }
```

DavideD: Thanks, I can see that it works now. Probably, there was something else wrong when I tried before.

DavideD: I've added a fix. Thanks

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/7

Issue #47, opened 2020-01-21
hi/lo id generation


gavinking:
Our table and sequence id generators don't support the hi/lo optimization.

gavinking:
I think this is the issue we should work on next. (It's a potential performance bottleneck.)

gavinking:
This is now done in #318.

PR #57, opened 2020-02-13
#47 : Add "no-op" and "hilo" optimizers for sequences

gbadner:
Here is my attempt at adding support for "no-op" and "hilo" optimizers used with a sequence for https://github.com/hibernate/hibernate-rx/issues/47 .

This fix relies on the following change to hibernate-orm:
https://github.com/gbadner/hibernate-core/commit/fc9afeb0e816b0bca50825f8a54495b396e0b141

I hesitate to create an HHH jira with the above commit until the code in this PR is at least sanity-checked.

Also, this fix updates the hibernate-orm dependencies to 5.4.12-SNAPSHOT.

Please be kind.

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
@@ -1,0 +1,56 @@
+package org.hibernate.rx.persister.entity.impl;
+
+import org.hibernate.AssertionFailure;
+import org.hibernate.id.IdentifierGeneratorHelper;
+import org.hibernate.id.enhanced.HiLoOptimizer;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public class HiLoRxOptimizer extends HiLoOptimizer implements RxOptimizer<Long> {
+
+	public HiLoRxOptimizer(Class<?> integerType, int incrementSize) {
+		super( integerType, incrementSize );
+	}
+
+	@Override
+	public synchronized CompletionStage<Optional<Long>> generate(RxAccessCallback callback) {
+
+		final GenerationState generationState = locateGenerationState( callback.getTenantIdentifier() );
+
+		if ( !generationState.isInitialized() ) {
+			return getValidFirstValueFromDatabase( callback )
+					.thenCompose( validFirstValue -> getActualNextValue( generationState, validFirstValue ) );
+		}
+		else if ( generationState.requiresNewSourceValue() ) {
+			return callback.getNextValue()
+					.thenCompose( valueFromDatabase -> getActualNextValue( generationState, valueFromDatabase ) );
+		}
+		else {
+			return CompletableFuture.completedFuture(Optional.of(generationState.makeValueThenIncrement().longValue()));
+		}
+	}
+
+	private CompletionStage<Optional<Long>> getValidFirstValueFromDatabase(RxAccessCallback callback) {
+		// TODO: increment if it is < 1
+		// while ( generationState.lastSourceValue.lt( 1 ) ) {
+		//	generationState.lastSourceValue = callback.getNextValue();
+		//}
+
+		return callback.getNextValue();
+	}
+
+	private CompletionStage<Optional<Long>> getActualNextValue(GenerationState generationState, Optional<Long> valueFromDataBase) {
+
+		if (!valueFromDataBase.isPresent()) {
+			throw new AssertionFailure("No sequence value present!");
+		}
+
+		generationState.updateFromNewSourceValue(
+				IdentifierGeneratorHelper.getIntegralDataTypeHolder( Long.class ).initialize( valueFromDataBase.get() )
+		);
+
+		return CompletableFuture.completedFuture( Optional.of(generationState.makeValueThenIncrement().longValue()));
+	}
+}
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/IdentifierGeneration.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/IdentifierGeneration.java
@@ -1,5 +1,6 @@
 package org.hibernate.rx.persister.entity.impl;
 
+import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
 import org.hibernate.boot.model.naming.Identifier;
 import org.hibernate.boot.model.relational.Database;
@@ -21,6 +22,7 @@ import org.hibernate.mapping.SimpleValue;
 import org.hibernate.persister.spi.PersisterCreationContext;
 import org.hibernate.rx.util.impl.RxUtil;
 
+import javax.naming.OperationNotSupportedException;
 import java.util.Optional;
 import java.util.Properties;
 
@@ -81,7 +83,13 @@ public class IdentifierGeneration {
 				return new TableRxIdentifierGenerator(persistentClass, creationContext);
 			}
 			else if (structure instanceof SequenceStructure) {
-				return new SequenceRxIdentifierGenerator(persistentClass, creationContext);
+				return new SequenceRxIdentifierGenerator(
+						persistentClass,
+						creationContext,
+						createRxOptimizer(
+								((SequenceStyleGenerator) identifierGenerator).getOptimizer()
+						)
+				);
 			}
 			throw new IllegalStateException("unknown structure type");
 		}
@@ -89,7 +97,8 @@ public class IdentifierGeneration {
 			return new TableRxIdentifierGenerator(persistentClass, creationContext);
 		}
 		else if (identifierGenerator instanceof SequenceGenerator) {
-			return new SequenceRxIdentifierGenerator(persistentClass, creationContext);
+			throw new UnsupportedOperationException(
+					identifierGenerator.getClass().getName() + " is no longer supported. Use " + SequenceStyleGenerator.class.getName() + " instead.");
 		}
 		else if (identifierGenerator instanceof SelectGenerator) {
 			throw new HibernateException("SelectGenerator is not yet supported");
@@ -195,5 +204,17 @@ public class IdentifierGeneration {
 			);
 		}
 		return qualifiedTableName;
+	}
+
+	static RxOptimizer createRxOptimizer(Optimizer optimizer) {
+		if ( optimizer instanceof NoopOptimizer ) {
+			return new NoopRxOptimizer();
+		}
+		else if ( optimizer instanceof HiLoOptimizer ) {
+			return new HiLoRxOptimizer( ((HiLoOptimizer) optimizer).getReturnClass(), optimizer.getIncrementSize() );
+		}
+		else {
+			throw new HibernateException( optimizer.getClass().getName() + " is not supported yet" );
+		}
 	}
 }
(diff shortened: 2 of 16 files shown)


hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/IdentifierGeneration.java
```diff
@@ -81,15 +83,22 @@ static Properties identifierGeneratorProperties(
...
 			}
 			throw new IllegalStateException("unknown structure type");
 		}
 		else if (identifierGenerator instanceof TableGenerator) {
 			return new TableRxIdentifierGenerator(persistentClass, creationContext);
 		}
 		else if (identifierGenerator instanceof SequenceGenerator) {
-			return new SequenceRxIdentifierGenerator(persistentClass, creationContext);
+			throw new UnsupportedOperationException(
+					identifierGenerator.getClass().getName() + " is no longer supported. Use " + SequenceStyleGenerator.class.getName() + " instead.");
```
gbadner: {{SequenceGenerator}} is deprecated. Do we really need to support it here? If so, I'll implement as appropriate.

gavinking:
Oh, excellent, thanks @gbadner!

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.CompletionStage;
+
+public class HiLoRxOptimizer extends HiLoOptimizer implements RxOptimizer<Long> {
+
+	public HiLoRxOptimizer(Class<?> integerType, int incrementSize) {
+		super( integerType, incrementSize );
+	}
+
+	@Override
+	public synchronized CompletionStage<Optional<Long>> generate(RxAccessCallback callback) {
```
DavideD: This looks correct to me but I'm wondering if we could test it under stress to have some confidence that it's actually thread safe.

This is an example of what we did for OGM: https://github.com/hibernate/hibernate-ogm/blob/master/core/src/test/java/org/hibernate/ogm/backendtck/id/SequenceNextValueGenerationTest.java

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/IdentifierGeneration.java
DavideD: I'm fine with not implementing it

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+
+		if (!valueFromDataBase.isPresent()) {
+			throw new AssertionFailure("No sequence value present!");
+		}
+
+		generationState.updateFromNewSourceValue(
+				IdentifierGeneratorHelper.getIntegralDataTypeHolder( Long.class ).initialize( valueFromDataBase.get() )
+		);
+
+		return CompletableFuture.completedFuture( Optional.of(generationState.makeValueThenIncrement().longValue()));
```
DavideD: For consistency. should be `RxUtil.completedFuture(...`

hibernate-rx-core/src/test/java/org/hibernate/rx/SequenceHiLoGeneratorTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+		sessionStage = sessionStage.thenCompose(s -> s.flush());
+		for (int i = 0; i < 7; i++) {
+			final int id = i + 4;
+			sessionStage = sessionStage.thenCompose(s2 ->
+					findAndCheck(s2, context, id, id + 1, 1 )
+			);
+		}
+
+		sessionStage = sessionStage.thenCompose(s3 -> openSession()  );
```
DavideD: Double spaces after   openSessions()

gbadner:
Hibernate ORM PR is https://github.com/hibernate/hibernate-orm/pull/3236

gbadner:
@Sanne has concerns with the ORM PR:
https://github.com/hibernate/hibernate-orm/pull/3236#issuecomment-586172410

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+		if ( !generationState.isInitialized() ) {
+			return getValidFirstValueFromDatabase( callback )
+					.thenCompose( validFirstValue -> getActualNextValue( generationState, validFirstValue ) );
+		}
+		else if ( generationState.requiresNewSourceValue() ) {
+			return callback.getNextValue()
+					.thenCompose( valueFromDatabase -> getActualNextValue( generationState, valueFromDatabase ) );
+		}
+		else {
+			return CompletableFuture.completedFuture(Optional.of(generationState.makeValueThenIncrement().longValue()));
```
Sanne: When `GenerationState` is in such a state that clients needing a new ID reach this line, they will all be creating a `CompletableFuture` wrapping `makeValueThenIncrement`: so you risk having thousands of threads hitting the DB to allocate a new block.

In Hibernate ORM this is prevented by the `synchronized` keyword: all threads need to wait for the first increment to have completed; in this case we're by-passing the lock, which is going to flood the DB with increase requests - that's perhaps "correct" but it's side-stepping the point of the optimiser :-/

I'm afraid this needs to be modelled in a different way. If there's one operation of `makeValueThenIncrement` being scheduled or in progress, the other operations need to wait on that same operation, and then check again for the state as it's possible that a single increment operation didn't allocate enough IDs for all clients.

gavinking:
@Sanne @gbadner So casting my mind back ten years or so, I'm pretty sure I recall that the earliest hi/lo generator in Hibernate (which no longer exists in the codebase) didn't attempt to minimize id wastage, and so if you had two txns arrive at the same time, they would both simultaneously allocate a new block, thereby wasting almost a whole block. The reasoning behind this was precisely that we went out of our way to avoid the possibility of one thread blocking another, *especially* over IO.

So what I understand from Sanne's comments is that in the current design of Hibernate, we attempt to avoid this behavior by blocking all other threads while a new block of ids is being allocated. And so of course this behavior is no good from the point of RX where we an even stronger rule that not only can threads not block each other, but that they can't even block themselves!

Now it seems to me (and perhaps this is a little naïve) that the obvious solution at least for now is to just go back to the ancient behavior and simply accept that blocks can be wasted in the case of a lot of threads looking for a new block all at once.

Or is this so terrible that we need to look for another option like querying the DB first to see if the new block was already allocated before allocating the new block? I have not tried to think this through.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
gavinking: @Sanne is is really the case that you're likely to have thousands of inserts happening simultaneously?

I mean, a system which does 1000 inserts a second is adding almost 100 million rows to the table each day. That can't possibly be the common case can it?

gavinking:
FTR: I've never really understood why people are bothered about "wasting" IDs. There are more than enough positive integers. I assume that it is because they're not really pure surrogate keys and that they're actually keys that are displayed to the user.

OTOH, I do understand that in the non-RX world it can be more efficient to just block a thread than to make it do an unnecessary round trip to the database.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
Sanne: Fair enough, it might be a bit pathological - but that's why I'm saying that I'd rather see a "dumb" optimiser implemented first, as the whole point of the current HiLoOptimiser is to optimise for such pathological cases?

But also consider, that a system that _happens to_ issue 1000 inserts in a second in a specific point in time, might be having a momentary spike; it's totally possible that there won't be any other insert for the rest of the day - so while doing 100 million inserts a day might be a stretch, we still need to handle spikes graciously.  I'd argue such spikes are not far fetched, there might be the "daily upload" of data and processes like that - just saying let's not look at average expected rates. (But if we did, 100M inserts a day are not unlikely either!)

Consider what happens when the in-memory pool is exhausted - which will happen periodically and quite often - and you have a non trivial amount of operations "on the stack" of the various threads running this RX system (which supposedly is good for highly concurrent scenarios): you'll have _ALL_ operations that hit this method also hit the database, for quite a long span of time:

the T1 starting point of time is when we first exhausted the pool
the T2 end point is when the first completable stage actually returned from the DB (and replenished the pool)

_ALL_ operations between T2-T1 will have caused an increse in the "page" on the DB. So let's focus on that a second: we're not skipping individual IDs, but whole pages of {sizeIncrement}, multipled by the amount of parallel ops that this reactive system is able to take... Worst aspect is that soon after, we'll repeat this again: I'm afraid we'll exhaust any sequence rather fast.

In terms of design, I believe the goal of such optimisers is to prevent such a picture - however I can agree that this is not a level of optimisations we need for a first MVP of the project: just explaining why I believe current Optimisers can't be taken as-is.

Sanne:
> I've never really understood why people are bothered about "wasting" IDs. 

Sure, I don't mind us having some "holes" in the sequence of IDs. But the context I describe in https://github.com/hibernate/hibernate-rx/pull/57#discussion_r380336911 would waste many "pages" of IDs at once, and repeatedly.  I hope that's clearer now - wouldn't want to nitpick otherwise :)

I don't think it's overly complex to do this the better way - but I hope I managed to explain why it's not a straight forward adaptation of the existing code.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/HiLoRxOptimizer.java
gavinking: > But also consider, that a system that happens to issue 1000 inserts in a second in a specific point in time, might be having a momentary spike; it's totally possible that there won't be any other insert for the rest of the day

Sure sure, I know that's possible. I'm just saying that even to get a spike like that I think you would have to be dealing with something pretty specific, and it's reasonable that such general purpose optimizers might not be the right thing to use for something so specific.

I mean, sure, you *can* get such an avalanche by chance, but only with [extremely low probability](https://en.wikipedia.org/wiki/Poisson_distribution).

> I'd argue such spikes are not far fetched, there might be the "daily upload" of data and processes like that

Yes of course, but a "daily upload" is more likely to happen in a single thread, not distributed across 1000 threads.

> In terms of design, I believe the goal of such optimisers is to prevent such a picture - however I can agree that this is not a level of optimisations we need for a first MVP of the project: just explaining why I believe current Optimisers can't be taken as-is.

Look, TBH I don't have a strong feeling about how optimizers should work *at all*. I would just like it if we made some sort of basic no-frills best effort to respect the `allocationSize` our users have explicitly specified in `@TableGenerator` or `@SequenceGenerator`, so we can cross that off the list of Limitations, i.e. of unsupported features of JPA.

Sure, sure, our initial implementation might not exhibit ideal behavior for some particularly hardcore workloads, but at least for the most common case where inserts arrive in a steady flow, according to a typical Poisson distribution, rather than in sudden avalanches, it would cut the wait time in half, by avoiding an additional roundtrip on every insert.

(I don't think we're really disagreeing over very much here.)

Sanne: right we're not disagreeing: let's forget about Hi/Lo optimisers for now and focus on the no-frills implementations we really need.

:+1:

gavinking:
@Sanne When I opened issue #47 all I was really looking for was the simple possible implementation of `allocationSize` because we currently pretty much just ignore it. When @gbadner came up with an implementation that reused code in core I was even happier because I figured that we would get some extra nice behavior for free.

DavideD:
I'm going to close this issue for now. We can get back to this later on and think of a better design

Issue #40, opened 2020-01-20
Hibernate RX alongside regular Hibernate


gavinking:
While I admit I have not actually *tried*, from my understanding of how the code works, I conclude that it's currently impossible to have reactive sessions alongside regular sessions for the same `SessionFactory`. It might even be impossible to have them coexist in the same *program*.

Right, @DavideD?

I think that's something we eventually need to fix.

DavideD:
I think you are correct, but I will check.

gavinking:
Alright, so a major roadblock to resolving this problem is going to be the fact that we're hijacking Hibernate's regular `PERSIST`, `FLUSH`, etc, event types to do implement our own differently-semantic'd operations.

So today I tried to fix this by breaking us away from Hibernate's `EventType`s and event processing and, surprisingly, I failed! It looked, naively, like there would be about two or three different ways to do this but not a single one of them panned out, and not even with the sort of "adjustments" I made in #36, could I do it this way. Without some sort of a more major change in core, this one is actually much harder than it looks.

The problems relate to the whole interrelated lifecycles of `EventType`s, the `EventListenerRegistry` and `CallbackRegistry`. You might think you can implement this stuff without even using events at all, but that's not, AFAICT, the case, since you've got no other way of getting access to the  `CallbackRegistry`. And you need it!

So to solve this issue we would need to work with the Hibernate core team to add a well-defined way to augment the set of event types.

gavinking:
Take note ☝️@Sanne  :-)

Sanne:
I don't think there's any desire to use one single SessionFactory both in reactive and in non-reactive way?  IMO either you're the kind of developer liking reactive APIs, or not ...

If there's some 10% of people that are say migrating an application and need both APIs available "for some time" I think it would be perfectly fine for them to start two differently configured instances. 

Also, assuming we don't develop some kind of sick adapter, they'll need two differently configured datasources anyway.

gavinking:
> I don't think there's any desire to use one single SessionFactory both in reactive and in non-reactive way?

*Perhaps* not, (it's not clear to me) but I'm pretty sure our current implementation won't even let you have one factory for reactive and one for normal.

gavinking:
> Also, assuming we don't develop some kind of sick adapter, they'll need two differently configured datasources anyway.

I don't think that's right. We already use regular JDBC for schema export right alongside Vert.x for DML.

DavideD:
[...]

PR #66, opened 2020-02-26
Fix master

DavideD:
This PR contains all the changes we discussed here: https://github.com/hibernate/hibernate-rx/pull/58
It fixes the build so I'm going to merge it, we can revert it later.

I've changed the credentials and database name to `hibernaterx` so that it doesn't cause problem to MySQL. Make sure to update your dbs.

Cheers

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/collection/RxPersistentList.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/collection/RxPersistentList.java
@@ -1,0 +1,42 @@
+package org.hibernate.rx.collection;
+
+import org.hibernate.HibernateException;
+import org.hibernate.collection.internal.PersistentList;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.loader.CollectionAliases;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.rx.util.impl.RxUtil;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * A {@link PersistentList} that makes sure that {@link CompletionStage} values complete before adding them
+ * to the underlying list;
+ */
+public class RxPersistentList extends PersistentList {
+
+	public RxPersistentList(SharedSessionContractImplementor session, List delegate ) {
+		super( session, delegate);
+	}
+
+	@Override
+	public Object readFrom(ResultSet rs, CollectionPersister persister, CollectionAliases descriptor, Object owner) throws HibernateException, SQLException {
+		// FIXME: Create a method for reactive reading. For example: rxReadElement
+		final Object element = persister.readElement( rs, owner, descriptor.getSuffixedElementAliases(), getSession() );
+		if ( element instanceof CompletionStage ) {
+			CompletionStage<?> elementStage = (CompletionStage<?>) element;
+			return elementStage.thenApply( obj ->  {
+				if ( obj != null) {
+					add( obj );
+				}
+				return obj;
+			} );
+		}
+		else {
+			return super.readFrom( rs, persister, descriptor, owner );
+		}
+	}
+}
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
@@ -1,0 +1,511 @@
+package org.hibernate.rx.engine.impl;
+
+import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.MappingException;
+import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.loading.internal.LoadContexts;
+import org.hibernate.engine.spi.*;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.rx.engine.loading.impl.RxLoadContexts;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+public class RxPersistenceContext implements PersistenceContext {
+
+	private final PersistenceContext delegate;
+
+	public RxPersistenceContext(PersistenceContext context) {
+		this.delegate = context;
+	}
+
+	@Override
+	public LoadContexts getLoadContexts() {
+		return new RxLoadContexts( delegate.getLoadContexts() );
+	}
+
+	//////////////////
+	// DELEGATES
+	//////////////////
+
+	@Override
+	public boolean isStateless() {
+		return delegate.isStateless();
+	}
+
+	@Override
+	public SharedSessionContractImplementor getSession() {
+		return delegate.getSession();
+	}
+
+	@Override
+	public void addUnownedCollection(CollectionKey key, PersistentCollection collection) {
+		delegate.addUnownedCollection(key, collection);
+	}
+
+	@Override
+	public PersistentCollection useUnownedCollection(CollectionKey key) {
+		return delegate.useUnownedCollection(key);
+	}
+
+	@Override
+	public BatchFetchQueue getBatchFetchQueue() {
+		return delegate.getBatchFetchQueue();
+	}
+
+	@Override
+	public void clear() {
+		delegate.clear();
+	}
+
+	@Override
+	public boolean hasNonReadOnlyEntities() {
+		return delegate.hasNonReadOnlyEntities();
+	}
+
+	@Override
+	public void setEntryStatus(EntityEntry entry, Status status) {
+		delegate.setEntryStatus(entry, status);
+	}
+
+	@Override
+	public void afterTransactionCompletion() {
+		delegate.afterTransactionCompletion();
+	}
+
+	@Override
+	public Object[] getDatabaseSnapshot(Serializable id, EntityPersister persister) {
+		return delegate.getDatabaseSnapshot(id, persister);
+	}
+
+	@Override
+	public Object[] getCachedDatabaseSnapshot(EntityKey key) {
+		return delegate.getCachedDatabaseSnapshot(key);
+	}
+
+	@Override
+	public Object[] getNaturalIdSnapshot(Serializable id, EntityPersister persister) {
+		return delegate.getNaturalIdSnapshot(id, persister);
+	}
+
+	@Override
+	public void addEntity(EntityKey key, Object entity) {
+		delegate.addEntity(key, entity);
+	}
+
+	@Override
+	public Object getEntity(EntityKey key) {
+		return delegate.getEntity(key);
+	}
+
+	@Override
+	public boolean containsEntity(EntityKey key) {
+		return delegate.containsEntity(key);
+	}
+
+	@Override
+	public Object removeEntity(EntityKey key) {
+		return delegate.removeEntity(key);
+	}
+
+	@Override
+	public void addEntity(EntityUniqueKey euk, Object entity) {
+		delegate.addEntity(euk, entity);
+	}
+
+	@Override
+	public Object getEntity(EntityUniqueKey euk) {
+		return delegate.getEntity(euk);
+	}
+
+	@Override
+	public EntityEntry getEntry(Object entity) {
+		return delegate.getEntry(entity);
+	}
+
+	@Override
+	public EntityEntry removeEntry(Object entity) {
+		return delegate.removeEntry(entity);
+	}
+
+	@Override
+	public boolean isEntryFor(Object entity) {
+		return delegate.isEntryFor(entity);
+	}
+
+	@Override
+	public CollectionEntry getCollectionEntry(PersistentCollection coll) {
+		return delegate.getCollectionEntry(coll);
+	}
+
+	@Override
+	public EntityEntry addEntity(Object entity, Status status, Object[] loadedState, EntityKey entityKey, Object version, LockMode lockMode, boolean existsInDatabase, EntityPersister persister, boolean disableVersionIncrement) {
+		return delegate.addEntity(entity, status, loadedState, entityKey, version, lockMode, existsInDatabase, persister, disableVersionIncrement);
+	}
+
+	@Override
+	public EntityEntry addEntry(Object entity, Status status, Object[] loadedState, Object rowId, Serializable id, Object version, LockMode lockMode, boolean existsInDatabase, EntityPersister persister, boolean disableVersionIncrement) {
+		return delegate.addEntry(entity, status, loadedState, rowId, id, version, lockMode, existsInDatabase, persister, disableVersionIncrement);
+	}
+
+	@Override
+	public boolean containsCollection(PersistentCollection collection) {
+		return delegate.containsCollection(collection);
+	}
+
+	@Override
+	public boolean containsProxy(Object proxy) {
+		return delegate.containsProxy(proxy);
+	}
+
+	@Override
+	public boolean reassociateIfUninitializedProxy(Object value) throws MappingException {
+		return delegate.reassociateIfUninitializedProxy(value);
+	}
+
+	@Override
+	public void reassociateProxy(Object value, Serializable id) throws MappingException {
+		delegate.reassociateProxy(value, id);
+	}
+
+	@Override
+	public Object unproxy(Object maybeProxy) throws HibernateException {
+		return delegate.unproxy(maybeProxy);
+	}
+
+	@Override
+	public Object unproxyAndReassociate(Object maybeProxy) throws HibernateException {
+		return delegate.unproxyAndReassociate(maybeProxy);
+	}
+
+	@Override
+	public void checkUniqueness(EntityKey key, Object object) throws HibernateException {
+		delegate.checkUniqueness(key, object);
+	}
+
+	@Override
+	public Object narrowProxy(Object proxy, EntityPersister persister, EntityKey key, Object object) throws HibernateException {
+		return delegate.narrowProxy(proxy, persister, key, object);
+	}
+
+	@Override
+	public Object proxyFor(EntityPersister persister, EntityKey key, Object impl) throws HibernateException {
+		return delegate.proxyFor(persister, key, impl);
+	}
+
+	@Override
+	public Object proxyFor(Object impl) throws HibernateException {
+		return delegate.proxyFor(impl);
+	}
+
+	@Override
+	public void addEnhancedProxy(EntityKey key, PersistentAttributeInterceptable entity) {
+		delegate.addEnhancedProxy(key, entity);
+	}
+
+	@Override
+	public Object getCollectionOwner(Serializable key, CollectionPersister collectionPersister) throws MappingException {
+		return delegate.getCollectionOwner(key, collectionPersister);
+	}
+
+	@Override
+	public Object getLoadedCollectionOwnerOrNull(PersistentCollection collection) {
+		return delegate.getLoadedCollectionOwnerOrNull(collection);
+	}
+
+	@Override
+	public Serializable getLoadedCollectionOwnerIdOrNull(PersistentCollection collection) {
+		return delegate.getLoadedCollectionOwnerIdOrNull(collection);
+	}
+
+	@Override
+	public void addUninitializedCollection(CollectionPersister persister, PersistentCollection collection, Serializable id) {
+		delegate.addUninitializedCollection(persister, collection, id);
+	}
+
+	@Override
+	public void addUninitializedDetachedCollection(CollectionPersister persister, PersistentCollection collection) {
+		delegate.addUninitializedDetachedCollection(persister, collection);
+	}
+
+	@Override
+	public void addNewCollection(CollectionPersister persister, PersistentCollection collection) throws HibernateException {
+		delegate.addNewCollection(persister, collection);
+	}
+
+	@Override
+	public void addInitializedDetachedCollection(CollectionPersister collectionPersister, PersistentCollection collection) throws HibernateException {
+		delegate.addInitializedDetachedCollection(collectionPersister, collection);
+	}
+
+	@Override
+	public CollectionEntry addInitializedCollection(CollectionPersister persister, PersistentCollection collection, Serializable id) throws HibernateException {
+		return delegate.addInitializedCollection(persister, collection, id);
+	}
+
+	@Override
+	public PersistentCollection getCollection(CollectionKey collectionKey) {
+		return delegate.getCollection(collectionKey);
+	}
+
+	@Override
+	public void addNonLazyCollection(PersistentCollection collection) {
+		delegate.addNonLazyCollection(collection);
+	}
+
+	@Override
+	public void initializeNonLazyCollections() throws HibernateException {
+		delegate.initializeNonLazyCollections();
+	}
+
+	@Override
+	public PersistentCollection getCollectionHolder(Object array) {
+		return delegate.getCollectionHolder(array);
+	}
+
+	@Override
+	public void addCollectionHolder(PersistentCollection holder) {
+		delegate.addCollectionHolder(holder);
+	}
+
+	@Override
+	public PersistentCollection removeCollectionHolder(Object array) {
+		return delegate.removeCollectionHolder(array);
+	}
+
+	@Override
+	public Serializable getSnapshot(PersistentCollection coll) {
+		return delegate.getSnapshot(coll);
+	}
+
+	@Override
+	public CollectionEntry getCollectionEntryOrNull(Object collection) {
+		return delegate.getCollectionEntryOrNull(collection);
+	}
+
+	@Override
+	public Object getProxy(EntityKey key) {
+		return delegate.getProxy(key);
+	}
+
+	@Override
+	public void addProxy(EntityKey key, Object proxy) {
+		delegate.addProxy(key, proxy);
+	}
+
+	@Override
+	public Object removeProxy(EntityKey key) {
+		return delegate.removeProxy(key);
+	}
+
+	@Override
+	@Deprecated
+	public HashSet getNullifiableEntityKeys() {
+		return delegate.getNullifiableEntityKeys();
+	}
+
+	@Override
+	@Deprecated
+	public Map getEntitiesByKey() {
+		return delegate.getEntitiesByKey();
+	}
+
+	@Override
+	public Map.Entry<Object, EntityEntry>[] reentrantSafeEntityEntries() {
+		return delegate.reentrantSafeEntityEntries();
+	}
+
+	@Override
+	@Deprecated
+	public Map getEntityEntries() {
+		return delegate.getEntityEntries();
+	}
+
+	@Override
+	public int getNumberOfManagedEntities() {
+		return delegate.getNumberOfManagedEntities();
+	}
+
+	@Override
+	@Deprecated
+	public Map getCollectionEntries() {
+		return delegate.getCollectionEntries();
+	}
+
+	@Override
+	public void forEachCollectionEntry(BiConsumer<PersistentCollection, CollectionEntry> action, boolean concurrent) {
+		delegate.forEachCollectionEntry(action, concurrent);
+	}
+
+	@Override
+	@Deprecated
+	public Map getCollectionsByKey() {
+		return delegate.getCollectionsByKey();
+	}
+
+	@Override
+	public int getCascadeLevel() {
+		return delegate.getCascadeLevel();
+	}
+
+	@Override
+	public int incrementCascadeLevel() {
+		return delegate.incrementCascadeLevel();
+	}
+
+	@Override
+	public int decrementCascadeLevel() {
+		return delegate.decrementCascadeLevel();
+	}
+
+	@Override
+	public boolean isFlushing() {
+		return delegate.isFlushing();
+	}
+
+	@Override
+	public void setFlushing(boolean flushing) {
+		delegate.setFlushing(flushing);
+	}
+
+	@Override
+	public void beforeLoad() {
+		delegate.beforeLoad();
+	}
+
+	@Override
+	public void afterLoad() {
+		delegate.afterLoad();
+	}
+
+	@Override
+	public boolean isLoadFinished() {
+		return delegate.isLoadFinished();
+	}
+
+	@Override
+	public String toString() {
+		return delegate.toString();
+	}
+
+	@Override
+	public Serializable getOwnerId(String entityName, String propertyName, Object childEntity, Map mergeMap) {
+		return delegate.getOwnerId(entityName, propertyName, childEntity, mergeMap);
+	}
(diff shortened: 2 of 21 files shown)


gavinking:
@DavideD I disagree with this course of action.

This is a huge patch which does all sorts of stuff which can't possibly be related to the problem we're trying to fix. It just *can't* be the case that we need all the code in this PR to fix whatever presumably small change in behavior was made in hibernate-core.

And apparently, despite its size, it's an *incomplete* fix, since it doesn't fix `CascadeTest`. So I think we need to properly investigate the true cause of the breakage, rather than just pushing partial fixes for something we don't understand. I mean, hell, perhaps this is the result of a *regression* in core!

Now, I'm not saying that your all your changes are bad or wrong; it seems to me that most of the changes in your commit bceefd8d10c48498c1915a74be3917a1ee3230d2 must be doing other unrelated stuff to do with collections, and aren't directly about fixing this problem. If that's correct, I think that commit needs to describe what it's really doing in its commit message.

gavinking:
> I've changed the credentials and database name to hibernaterx so that it doesn't cause problem to MySQL. 

See 9122e4dba8d7915e8512ba0770a5d576b7e0e233 where I already change it to use `hibernate-rx` as the database name on MySQL.

gavinking:
(If I understand correctly, @gbadner is already hard at work investigating the cause of the breakage, and I think we should give her a chance to figure it out before we decide what action to take.)

gbadner:
@gavinking, using this [commit](https://github.com/hibernate/hibernate-orm/commit/252ca9e98b3691728697a96a7dddcae32e666f8f) from Hibernate ORM master, I am getting the following failures:

org.hibernate.rx.MySQLBasicTest > testBasicTypes FAILED
    org.hibernate.service.spi.ServiceException
        Caused by: org.hibernate.exception.GenericJDBCException
            Caused by: java.sql.SQLException
    java.lang.NullPointerException

org.hibernate.rx.MySQLAutoincrementTest > testBasicTypes FAILED
    org.hibernate.service.spi.ServiceException
        Caused by: org.hibernate.exception.GenericJDBCException
            Caused by: java.sql.SQLException
    java.lang.NullPointerException

org.hibernate.rx.LazyManyToOneAssociationTest > fetchWithTwoAuthors FAILED
    java.lang.IllegalStateException
        Caused by: io.netty.channel.ChannelException
            Caused by: java.io.IOException

org.hibernate.rx.LazyManyToOneAssociationTest > fetchProfileWithOneAuthor FAILED
    org.hibernate.service.spi.ServiceException
        Caused by: java.lang.IllegalStateException
            Caused by: io.netty.channel.ChannelException
                Caused by: java.io.IOException

I don't have MySQL set up. Would that be the reason for MySQL failures?

For now, I'll assume I've found a "good" commit.

gbadner:
@DavideD , @gavinking, this is the commit that broke things:
https://github.com/hibernate/hibernate-orm/commit/5c1cc35f9842222b74901459245c1bf387e398aa

This is a wild guess as I have not even looked at the related code -- I am guessing that this commit resulted in hibernate-rx listeners replacing ORM ones? 

If so, it seems like it would make sense that the hibernate-rx listeners would need to be refined.

If that is the case, then I will assume that this was an expected breaking change, and I will switch to using this PR, and start investigating why CascadeTest is failing.

gavinking:
Is there an issue related to this commit?

gbadner:
@gavinking, https://hibernate.atlassian.net/browse/HHH-13831

gavinking:
OK, so thanks @gbadner, this hardens my opinion that *it can't possibly be the case* that the fix to this problem requires 2300 lines of new code in Hibernate RX. Do you guys understand what I'm saying here?

DavideD:
> _it can't possibly be the case_

Pretty bold statement... anyway we are looking into it

gavinking:
OK, then, so I surmise, *but am not certain*, that what is going on here is that previously some operations in Hibernate core could get hold of their own old regular `DefaultXxxxxEventListener`s, which were still available via calling `fireEventOnEachListener()`. Whereas Hibernate RX, which always calls `listeners()`, always got its own `DefaultRxXxxxxEventListener`s.

Now, of course, there is functionality in RX that depends upon plain vanilla functionality of Hibernate core, and so the effect of the fix it listener replacement was to break that functionality of core. This would explain why cascade persistence got broken, I imagine. I suppose it also breaks lifecycle callbacks, but I have not checked. In general, there's probably extra broken stuff that we haven't noticed because we don't have explicit tests for it in Hibernate RX. (All this assuming that my understanding of the problem is correct.)

Now, as I've argued in #40, what we're doing here with replacing Hibernate's built-in listeners is in fact completely conceptually *wrong*! The operations of a reactive session are *different* operations with different semantics, and it's an abuse to hijack the events of Hibernate core which have their own well-defined semantics. Indeed, it's worse than an abuse since it means that we simply break Hibernate for anyone who introduces Hibernate RX into their program even to do one simple little thing!

Now, I had already attempted to fix this problem, and failed, because when I tried to introduce our own `RxEvent`s I ran up against the fact that I simply couldn't obtain the `CallbackRegistry` without being a Hibernate core built-in event type.

So look, it's pretty clear to me that the *correct* fix to this issue is to make it possible either:

- obtain the Hibernate core `CallbackRegistry` without hoop-jumping, or
- generalize Hibernate core's event listener registry just a little in order to allow extensions to define their own event types.

It seems to me that the second option is completely reasonable and surely can't be that difficult to implement.

So I think that's what we should do.

gavinking:
> anyway we are looking into it

OK, let me know if my above analysis is correct.

gbadner:
@DavideD, the version needs to be changed to 

This change will get CascadeTest to succeed up until the test removes `node` : https://pastebin.com/LFjuNYkK

The new failure is due to cascade trying to delete `node.elements`. The problem is that the collection elements are of type `CompletableFuture`, when `Element` is expected.

gbadner:
@gavinking, I'm still looking into why that change to ORM broke things...

DavideD:
> OK, then, so I surmise, but am not certain, that what is going on here is that previously some operations in Hibernate core could get hold of their own old regular DefaultXxxxxEventListeners, which were still available via calling fireEventOnEachListener(). Whereas Hibernate RX, which always calls listeners(), always got its own DefaultRxXxxxxEventListeners.

My understanding of what's happening here is this:
SessionImpl#internalLoad calls DefaultLoadEventListener instead of DefaultRxLoadEventListener.
This happened because of a bug in ORM that didn't replace the listeners everywhere. Because the ORM default returns an object and not a CompletionStage we didn't notice and the code worked well anyway. When I made the bug fix in ORM the right listener got called returning a CompletionStage that we use as id without first waiting for its completion.

What my patch does is make sure that the value applied is the one returned by the CompletionStage and not the stage itself. This means changing the signatures of some methods.
I now have a better understanding of what's going on and I think I can reduce the changes.

> it's worse than an abuse since it means that we simply break Hibernate for anyone who introduces Hibernate RX into their program even to do one simple little thing!

I don't disagree that what we are doing is not great, but how is this related to this issue? This seems more a design approach that we can implement later.

> generalize Hibernate core's event listener registry just a little in order to allow extensions to define their own event types.

I don't know about this, even if you say "just a little", that part seems done that way for a reason. I guess we need to ask @Sanne or @sebersole. @gbadner, any idea?

I'm not saying we don't have to do this but, even with this change, don't we still have to implement similar listeners to what we have now? They still have to return CompletionStage and assign the values once the CompletionStage is completed.

@gavinking Am I missing something?

gavinking:
> My understanding of what's happening here is this:
`SessionImpl#internalLoad` calls `DefaultLoadEventListener` instead of `DefaultRxLoadEventListener`.

OK, but are you sure this is wrong? My memory is a little fuzzy, but I feel like I recall that this is the way it was supposed to work. Isn't it the job of `internalLoad()` to return proxies?

> Because the ORM default returns an object and not a CompletionStage we didn't notice and the code worked well anyway. 

Yeah but are you certain this isn't the behavior we want? We still have a need for an operation which returns proxies:

- for stubbing out unfetched associations, and
- for `RxSession.getReference()`.

(And perhaps for other things I have not thought of.)

> I don't disagree that what we are doing is not great, but how is this related to this issue?

Well, like I said, I'm *not* certain it's the cause of this issue, I'm hypothesizing that completely replacing Hibernate's default listeners broken functionality of core *that we actually want*. But sure, I could easily be wrong.

> I don't know about this, even if you say "just a little", that part seems done that way for a reason. 

Well, AFAIR I'm the original designer of the event listener stuff, which was introduced in Hibernate3, IIRC. It's designed around hardcoded event types because they correspond pretty much one-to-one with operations of the `Session`, which were never anticipated to be extensible. Now that we *are* extending the set of operations of a session, it makes sense to extend the set of event types.

I don't think it's right to hijack the pre-existing event types to implement different operations with fundamentally different semantics.

> They still have to return` CompletionStage` and assign the values once the `CompletionStage` is completed.

We *do* have a need for that *sometimes* (e.g., `RxSession.fetch()` and `RxSession.find()`) but it seems to me that we also probably have a need for the other regular operations in other contexts (e.g. `RxSession.getReference()`). Note that these might not be the very best examples of what I'm talking about.

> I'm not saying we don't have to do this but, even with this change, don't we still have to implement similar listeners to what we have now? 

Yes, we would. But we wouldn't replace the default Hibernate listeners with them, so we wouldn't break functionality that we want to get for free from core.

gbadner:
@DavideD, thanks for figuring out about `DefaultLoadEventListener` being called. I've verified it and it totally makes sense.

Regarding the fix I mentioned for `CascadeTest` -- I realized what a subtle bug it is. In a case like this, we need to make a defensive copy; in other cases, doing so will break things. 

This is probably obvious to you, but this reactive programming is very new to me.

```CascadeTest``` is a good example to show how copying ORM's cascade process to hibernater-rx is not the best way to do things.

hibernate-orm "cascade-before-save" ensures that an entity that is a foreign key target (e.g., many-to-one associated entity) is inserted before the entity that owns the foreign key.

This doesn't happen with hibernate-rx. It's easier to explain why that happens with an example.

(This may be completely obvious to you. I'm illustrating it for myself, to make sure I understand. I hope this is helpful to you as well.)

For `CascadeTest`, we have:

`Node basik = new Node("Child");`
`basik.parent = new Node("Parent");`
`...`

hibernate-orm cascade-before-save process inserts `basik.parent` before `basik`:

Insert `basik.parent` with `id == 2, parent_id == null, version == 0`: 
insert into Node (parent_id, string, version, id) values (?, ?, ?, ?)

Insert `basik` entity with `id == 1, parent_id == 2, version == 0`: 
insert into Node (parent_id, string, version, id) values (?, ?, ?, ?)

hibernate-rx cascade-before-save process does not complete before adding an insert action for `basik`:

Insert `basik` with `id == 1, parent_id == null, version == 0`: 
insert into Node (parent_id, string, version, id) values ($1, $2, $3, $4)

Insert `basik.parent` with `id == 2, parent_id == null, version == 0`: 
insert into Node (parent_id, string, version, id) values ($1, $2, $3, $4)

After the fix I mentioned [above](https://github.com/hibernate/hibernate-rx/pull/66#issuecomment-592315817), hibernate-rx needs to update `basik`'s `parent_id` as an extra step:

Update `basik` with `parent_id == 2, version == 1`:
update Node set parent_id=$1, string=$2, version=$3 where id=$4 and version=$5

Notice that with hibernate-orm, `basic.version == 0`; with hibernate-rx, `basic.version == 1` (due to the extra update).

The reason why this happens is because generation of the ID for `basik.parent` is delayed (in order to not block on the database operation), and the addition of the insert action for `basik` proceeds before cascade of`basik.parent` adds its insert action.

At this point, I can't say how this should be fixed. I agree these expose good reasons why the approach for hibernate-rx needs to be adjusted.

@gavinking, I am not clear on your suggestion about generalizing Hibernate core's event listener registry in order to allow extensions to define their own event types. I am thinking about it though.

**Comment by gavinking** [comment]
> The new failure is due to cascade trying to delete `node.elements`. The problem is that the collection elements are of type `CompletableFuture`, when `Element` is expected.

That makes sense. Originally we were using regular Hibernate proxies to sub out unfetched objects. And then it's the job of `RxSessionInternalImpl.immediateLoad()` to blow up if you do something that would require initialization of the proxy.

But if now `internalLoad()` returns a `CompletionStage` instead of a proxy, then all the places where we're expecting a proxy are now going to receive a `CompletionStage`. I've don't see how that is going to work, since `internalLoad()` is used from `EntityType`, for example.

**Comment by gavinking** [comment]
@gbadner 

> hibernate-rx cascade-before-save process does not complete before adding an insert action for `basik`:

Hrm, this is a consequence of what I've done in `AbstractRxSaveEventListener.rxPerformSaveOrReplicate()`. Instead reading the snapshot of the entity *after* the cascade operation has completed, I go ahead and do it synchronously. But as you point out, it's the cascade operation that results in generation of the parent id. So I need to consider changing this.

 > After the fix I mentioned above, hibernate-rx needs to update `basik`'s `parent_id` as an extra step:

I'm pretty sure this bug predates that fix, and is present in my original implementation of cascade for hibernate-rx.

**Comment by gbadner** [comment]
@gavinking, 

> I'm pretty sure this bug predates that fix, and is present in my original implementation of cascade for hibernate-rx

Could be. Is there anything I can help look into today?

**Comment by gavinking** [comment]
> So I need to consider changing this.

See https://github.com/hibernate/hibernate-rx/pull/71.

**Comment by gbadner** [comment]
@gavinking, OK, I'll take a look at https://github.com/hibernate/hibernate-rx/pull/71

@gavinking , @DavideD:  

Before switching gears, I just want to mention that `find()` returns an `Optional<Node>' with an `elements` collection that is a `RxPersistentList` containing an `ArrayList` of `Element` elements  (not `CompletableFuture<Element>` elements`).

`merge()` returns a `Node` with an `elements` collection that is a `RxPersistentList` containing  an `ArrayList` of CompletableFuture<Element>` elements`. Each `Element` is initialized to have `node.parent.elements` set to an empty initialized `PersistentBag`.

Definitely inconsistent. I'm not sure how it should be though.

**Comment by gbadner** [comment]
@gavinking, @Sanne, @sebersole, I'm looking into how Hibernate ORM can accommodate adding a new `EventType` and corresponding event listeners. I hope to have something to show you on Monday.

**Comment by gavinking** [comment]
@gbadner 👍awesome

**Comment by DavideD** [comment]
> OK, but are you sure this is wrong? My memory is a little fuzzy, but I feel like I recall that this is the way it was supposed to work. Isn't it the job of internalLoad() to return proxies?

Aren't there cases where it could return something from the db? That was my impression looking at the code. To be fair, I didn't understand that internalLoad was calling the load listener underneath (or I forgot). It seemed to me that the assumption was to call the same listener as in the other places. If that doesn't happen when you replace it with a custom listener, it's a bug in ORM for me.
If we always expect a proxy I can change the code patch accordingly and maybe simplify it (I need to think about it).

> it's true that I'm merely hypothesizing that completely replacing Hibernate's default listeners broke functionality of core that we actually want.

I think the issue here is more about what we expect as return value in the different use cases. Because I don't have the same clarity on the internal of ORM that you all have, it takes longer for me to understand the right workflow. If our goal is to reuse as much as possible from ORM, I don't think using different events will solve this. Still, it would be nice to have it and I'm glad @gbadner  is looking into it.

**Comment by gavinking** [comment]
> Aren't there cases where it could return something from the db? 

I believe we've assuming (for now) that when you set up your domain model you have a choose between:

- eager join fetching, and
- lazy fetching via `RxSession.fetch()` (i.e. an exception when you access the proxy without calling `fetch()` first).

That is to say, it's user error to config an association that would result in an immediate sequential `select`.

NOTE: To all of the above I need to add the caveat that eventually we're going to do #49, but speaking approximately, all that means is that you have a wrapper object that implicitly calls `fetch()` for you. Anyway, we're not doing #49 just yet.

> If we always expect a proxy

I think that's the model we were going for, yeah.

*That said,* perhaps it's the case that your code adds new functionality, namely, the ability to do sequential `select` fetching by composing `CompletionStage`s. That's something we didn't have until now, and if it works, then it's definitely useful!

Is that what you've implemented?

If it is, however:

1. I would like to see new functionality like that separated from "fixing master", because it seems to me it's a separate thing, and I would prefer for us to be able to look at it and test it out on its own terms.
2. The comment from @gbadner [here](https://github.com/hibernate/hibernate-rx/pull/66#issuecomment-592728201) suggests that perhaps this isn't totally 100% working. (I'm not sure because until this moment it had not occurred to me that this might be what you were trying to do, and so I have not tried to test it.)

Of course, if you've really implemented that, it would perfectly explain why we have 2000+ lines of new code here!

**Comment by gavinking** [comment]
@DavideD Please note that we have the open issue #39 that talks about adding support for eager `select` fetching. If that's what you've implemented here then: great!!

**Comment by DavideD** [comment]
> @DavideD Please note that we have the open issue #39 that talks about adding support for eager select fetching. If that's what you've implemented here then: great!!

It's not what I planned but I will test it out of curiosity

**Comment by gavinking** [comment]
> It's not what I planned but I will test it out of curiosity

👍

**Comment by gbadner** [comment]
@gavinking, @DavideD, FYI, in ORM, there are cases where an eager association will be loaded in a separate SELECT statement.

In ORM master, it will definitely happen in this case: https://hibernate.atlassian.net/browse/HHH-13085. I'm not sure if this affects hibernate-rx though.

**Comment by gbadner** [comment]
HHH-13085 is on my radar to get fixed; it's kind of complicated though. It affects queries also.

**Comment by gavinking** [comment]
@gbadner 

> FYI, in ORM, there are cases where an eager association will be loaded in a separate SELECT statement.

Yes, of course, indeed that's the basis for functionality like batch fetching, subselect fetching, and it's the best way to handle immutable "reference data" stored in the database (but cached).

But we discussed this at the meeting in Barcelona and I think we concluded that we didn't need to worry having this functionality just yet. Of course that conclusion was based on the assumption that it would be hard to implement. If @DavideD has already essentially implement it, then its an unsound conclusion.

**Comment by DavideD** [comment]
I've added this testcase and it seems to work fine: https://github.com/hibernate/hibernate-rx/pull/66/commits/f5a77ea15b22091060feae2c7c6ed0c695ff92e6
You need ORM 13-SNAPSHOT

**Comment by gavinking** [comment]
> I've added this testcase and it seems to work fine

Cool!

Then let me go through the rest of the changes and ask some questions....

File: hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/AbstractRxSaveEventListener.java [hunk 1]
```diff
@@ -277,7 +277,8 @@ else if (identifierType == IntegerType.INSTANCE) {
 		}
 
 		if ( substitute ) {
-			persister.setPropertyValues( entity, values );
+			cascadeBeforeSave = cascadeBeforeSave
+					.thenCompose( v ->  ((RxEntityPersister) persister).setRxPropertyValues(entity, values) );
 		}
```
gavinking: @DavideD what was the reason for this, and how does it relate the change I proposed here: https://github.com/hibernate/hibernate-rx/pull/71

DavideD:
@gavinking Let me clean up a bit the PR first
It' is a bit big for a single commit

hibernate-rx-core/src/main/java/org/hibernate/rx/collection/RxPersistentList.java
```diff
@@ -0,0 +1,42 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * A {@link PersistentList} that makes sure that {@link CompletionStage} values complete before adding them
+ * to the underlying list;
+ */
+public class RxPersistentList extends PersistentList {
```
gavinking: @DavideD what was the motivation for introducing this collection-related stuff in this work, bearing in mind that we currently document the following limitations:

> At this time, Hibernate RX does _not_ support the following features:
>
> - `@ElementCollection` and `@ManyToMany`
> - `@OneToMany` without `mappedBy` 

I was surprised to see a bunch of persistent collection-related stuff in here, because I didn't think we had tests that would exercise these features.

DavideD:
MAybe you can have a look at the test and see if it's correct first

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
```diff
@@ -0,0 +1,511 @@
...
+import org.hibernate.rx.engine.loading.impl.RxLoadContexts;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+public class RxPersistenceContext implements PersistenceContext {
```
gavinking: @DavideD What is the purpose of this class? It has no Javadoc? Am I right in surmising it exists just to plug in an `RxLoadContexts`? Is there a good reason to not just extend the public class `StatefulPersistenceContext`?

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxTwoPhaseLoad.java
```diff
@@ -0,0 +1,543 @@
...
+import org.hibernate.stat.spi.StatisticsImplementor;
+import org.hibernate.type.Type;
+import org.hibernate.type.TypeHelper;
+
+import org.jboss.logging.Logger;
+
+/**
+ * @see org.hibernate.engine.internal.TwoPhaseLoad
+ */
+public class RxTwoPhaseLoad {
```
gavinking: @DavideD I *hate hate hate* the idea of copy/pasting this very complex functionality into Hibernate RX. Is it necessary? If so, we need documentation explaining why it's necessary, and exactly what it does differently to `TwoPhaseLoad`.

I recently put a lot of work into reducing code duplication and adding JavaDoc to things in the Hibernate RX codebase, and this is precisely the sort of code duplication that I was trying to eliminate and/or properly document.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/loading/impl/RxLoadContexts.java
```diff
@@ -0,0 +1,64 @@
...
+import org.hibernate.engine.loading.internal.CollectionLoadContext;
+import org.hibernate.engine.loading.internal.EntityLoadContext;
+import org.hibernate.engine.loading.internal.LoadContexts;
+import org.hibernate.engine.spi.CollectionKey;
+import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.persister.collection.CollectionPersister;
+
+import java.sql.ResultSet;
+
+public class RxLoadContexts extends LoadContexts {
```
gavinking: @DavideD What is the purpose of this class? It's not documented, and it doesn't appear to do anything interesting.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxFlushEntityEventListener.java
```diff
@@ -139,53 +152,80 @@ public void onFlushEntity(FlushEntityEvent event) throws HibernateException {
...
-		boolean substitute = wrapCollections( session, persister, types, values );
+			if ( isUpdateNecessary( event, mightBeDirty ) ) {
+				substitute = scheduleUpdate( event ) || substitute;
+			}
 
-		if ( isUpdateNecessary( event, mightBeDirty ) ) {
-			substitute = scheduleUpdate( event ) || substitute;
-		}
+			if ( status != Status.DELETED ) {
+				// now update the object .. has to be outside the main if block above (because of collections)
```
gavinking: @DavideD Again I'm surprised to see collection-related changes, just because I thought we had such extremely limited support for collections.

In particular, we *don't* support making changes to the database via modifying a collection, so I'm very surprised to see collection-related changes to the *flush* functionality. What's going on here, exactly? Is it related to cascades or something?

gavinking:
@DavideD I added a bunch of comments. I speculate that what is going on with the patch is that you've had to jump through hoops to replace a bunch of collection-related functionality because it used to be that `internalLoad()` directly returned entity objects, whereas now it returns `CompletionStage`s.

If that's the case, then we're back where we started this conversation: I think it's the wrong fix. We should leave `internalLoad()` alone to do what it's always done. And we shouldn't need all this new code.

DavideD:
if `internalLoad` isn't suppose to always return a proxy, I don't understand why it should return something different from a `CompletionStage`. I will try to change the PR so that we can worry about this later.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
DavideD: 1. I don't like to extend if I can avoid it
2. This way we could potentially reuse this for other `PersistenceContext`s

hibernate-rx-core/src/main/java/org/hibernate/rx/collection/RxPersistentList.java
DavideD: See [RxPersistentList#readFrom](https://github.com/hibernate/hibernate-rx/pull/66/files/299835cf057732e91bc02ba2aabaa8e238b7fbd5#diff-ae92aa6395093fca5730bf179aa0a9ffR26). I needed to resolve the CompletionStage before setting the value in the List. That's the only way I've found to achieve this.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/loading/impl/RxLoadContexts.java
DavideD: Sorry, it's not needed. Just something I forgot to delete

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/AbstractRxSaveEventListener.java
DavideD: When associations are involved the value is returned as a CompletionStage. I'm trying to make sure that the value returned by the CompletionStage is set before using the collection anywherelse.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
DavideD: Sorry, I've just realized it's not needed in this PR.

gavinking:
> if `internalLoad` isn't suppose to always return a proxy, I don't understand why it should return something different from a `CompletionStage`.

Well what is actually being called (via a somewhat indirect route involving `EntityType`) is `AbstractEntityPersister.loadByUniqueKey()` and then `AbstractEntityLoader.load()`. The issue, I surmise is that this operation used to return things of the entity type (either the entity itself or a proxy). But after the change to event listeners, this operation now returns a `CompletionStage`.

I think that rather than having to introduce `RxPersistentList` and then `RxPersistentSet`, `RxPersistentMap` and then eventually `RxPersistentSortedSet`, etc, we need to figure out a solution where the pre-existing operations of `AbstractEntityPersister` return the entity type, as the did before, and you only get a `CompletionStage` in the operations which explicitly need one. (Namely, `RxSession.find()` and `RxSession.fetch()`.)

Remember: it's *only* on explicit calls to `RxSession.find()` and `RxSession.fetch()` that we want a `CompletionStage`. In all other cases we neither need nor want it.

Remember that it's not just persistent collections that are going to run into this problem: with this approach it seems to me that we're setting ourselves up for more pain when we start processing query result sets, which go via the same path of `EntityType` and `TwoPhaseLoad`.

hibernate-rx-core/src/main/java/org/hibernate/rx/collection/RxPersistentList.java
gavinking: Well I hope there's a better way to solve it.

Because note that it's not going to be enough to just introduce `RxPersistentList` without `RxPersistentSet`, `RxPersistentSortedSet`, etc, etc.

Would *really* prefer to not go down that path.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
gavinking: > 1. I don't like to extend if I can avoid it

Well by `extend`ing a class and overriding just one method, instead of implementing an interface and overriding tens of methods, we limit our exposure to changes in hibernate-core. We don't want to have hibernate-rx break every time there is a totally unrelated change to a method signature in core.

> 2. This way we could potentially reuse this for other `PersistenceContext`s

Are there other implementations that are relevant to Hibernate RX?

> Sorry, I've just realized it's not needed in this PR.

Even better.

DavideD:
Of course, but there are not so many of them and they are all similar.
Anyway, if I can avoid them I will.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/AbstractRxSaveEventListener.java
gavinking: OK, I see it now. Thanks.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
gavinking: > Well by extending a class and overriding just one method, instead of implementing an interface and overriding tens of methods, we limit our exposure to changes in hibernate-core. 

Note that there's an ancient bit of folklore from back in the 90s that holds that class extension represents (supposedly bad) "implementation inheritance" whereas interface implementation being pure type inheritance is something that is somehow more maintainable. This was always mostly nonsense, since adding an abstract operation to an interface is a change that breaks implementations of the interface. This bit of folklore was implicitly rejected by the gods of Java a few years ago by when support for `default` interface methods was added. But most of us had figured out that it was rubbish a bit before then.

gbadner:
@gavinking , @DavideD , I've created a POC for adding custom `EventType`s and associated event listeners at:
https://github.com/gbadner/hibernate-core/commit/2d368a2ce1da36468ab7f965467711bc59f439cf

The `CallbackRegistry` is available from any event listener that implements `CallbackRegistryConsumer`. Would this work for hibernate-rx purposes?

I've asked for feedback about this on hibernate-dev mailing list.

gavinking:
>  I've created a POC for adding custom `EventType`s and associated event listeners at:

Awesome!

> The `CallbackRegistry` is available from any event listener that implements `CallbackRegistryConsumer`. Would this work for hibernate-rx purposes?

I don't see why not.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContext.java
DavideD: In this case it made sense to me to add a delegator because I didn't really care about the specific implementation of the interface, PersistenceContext doesn't change that often and when it does it's easy to notice it and update the delegator. In practice,SatefulPersistenceContext seems to be the only implementation so I could have extended it.

DavideD:
See https://github.com/hibernate/hibernate-rx/pull/74

PR #67, opened 2020-02-27
Extract the Hibernate ORM version into a build parameter

Sanne:
For two reasons:
 1- ensure the versions are aligned among components (e.g. same version of `hibernate-core` as `hibernate-testing`
 2- allows to easily check compatibility with a different version

In particular, I'm thinking to setup a CI job which will monitor both existing ORM tags and next upcoming snapshots. 

    ./gradlew clean build -PhibernateOrmVersion=5.4.13-SNAPSHOT

That's not to say we need to strive to have master pass with multiple versions of ORM, but to be aware and informed: we can then decide ad-hoc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,15 @@ subprojects {
     }
 }
 
+// Versions which need to be aligned across modules;
+// this also allows to override the build from a parameter,
+// which can be useful to monitor compatibility for upcoming versions on CI:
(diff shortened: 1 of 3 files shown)


DavideD:
I've created an issue for the CI job: https://github.com/hibernate/hibernate-rx/issues/55
Feel free to update it as required

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/67

Thanks a lot

PR #71, opened 2020-02-28
fix problem in cascade persist

gavinking:
As reported by @gbadner [here](https://github.com/hibernate/hibernate-rx/pull/66#issuecomment-592692186), the way I implemented cascading operations wasn't quite perfect for cascade persist.

When we cascade the persist operation from a child to a parent, we should wait until the parent persist operation has completed *and its id has been generated* before capturing the snapshot of the child, because we're going to need the parent id as a foreign key.

This commit fixes that. However, I don't totally love this solution since it means cascade persist functions a bit differently from other cascading operations.

An alternative fix would be to cascade the persist operation in two phases, where the first phase only generates and assigns ids. But that's more work.

I need to think about this further.

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/AbstractRxSaveEventListener.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/AbstractRxSaveEventListener.java
@@ -267,45 +267,49 @@ abstract class AbstractRxSaveEventListener<C>
 
 		CompletionStage<Void> cascadeBeforeSave = rxCascadeBeforeSave(source, persister, entity, context);
 
-		Object[] values = persister.getPropertyValuesToInsert( entity, getMergeMap( context ), source );
-		Type[] types = persister.getPropertyTypes();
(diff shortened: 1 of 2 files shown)


gbadner:
@gavinking, your branch has Hibernate ORM 5.4.12.Final as a dependency.

I'm getting 42 failures with that version.

Do you have some other change on your local copy?

gavinking:
@gbadner Sure, I manually edited the dependency to `hibernateOrmVersion = '5.4.11-SNAPSHOT'`, i.e. the version of Hibernate before the change that broke us.

gbadner:
OK, thanks. I also had to change the user, password, and database back to "hibernate-rx".

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/71

DavideD:
I've merged it on master, I'm going to delete the branch from upstream

DavideD:
By the way @gavinking  could you fork the project and not create branches for fixes upstream, please? We usually reserved branches here for important ongoing ones.

PR #74, opened 2020-03-06
65 Fix master build

DavideD:
I left commit https://github.com/hibernate/hibernate-rx/commit/cc126844ee9d5c1b75e38826a651b9f0a11d8e30 to show what's the piece of code causing issues. But it requires the use of a blocking method.

The final fix is more complicated but I think it highlights better what's going on.

This patch is simpler than  https://github.com/hibernate/hibernate-rx/pull/66 and doesn't require the ORM dependency update.

The assumption I'm making here is that we only return a CompletionStage when we load the entity from the datasource and in all the other cases we return a proxy.

I removed the comment about checkIdClass because I think it was a mistake the way we implemented it. I will create an issue to make sure that we test that use case.

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
@@ -1,10 +1,23 @@
 package org.hibernate.rx.event.impl;
 
-import org.hibernate.*;
+import java.io.Serializable;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.NonUniqueObjectException;
+import org.hibernate.PersistentObjectException;
+import org.hibernate.TypeMismatchException;
 import org.hibernate.action.internal.DelayedPostInsertIdentifier;
 import org.hibernate.cache.spi.access.EntityDataAccess;
 import org.hibernate.cache.spi.access.SoftLock;
-import org.hibernate.engine.spi.*;
+import org.hibernate.engine.spi.EntityEntry;
+import org.hibernate.engine.spi.EntityKey;
+import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.engine.spi.Status;
 import org.hibernate.event.spi.EventSource;
 import org.hibernate.event.spi.LoadEvent;
 import org.hibernate.event.spi.LoadEventListener;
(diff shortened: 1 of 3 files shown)


gavinking:
@DavideD At first glance this looks *much* better. I'm busy right now but I will take a much closer look tonight or tomorrow morning, OK?

DavideD:
Sure

DavideD:
I've merged this one but feel free to review it when you have time and we can revert/fix what's necessary later.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
```diff
@@ -511,7 +595,7 @@ private Object createProxy(
 	 *
 	 * @return The loaded entity, or null.
 	 */
-	private CompletionStage<Optional<Object>> doLoad(
```
gavinking, line 514: @DavideD sorry if I'm a bit late in noticing this change, but I wonder why you made it?

Ultimately I think the pattern of returning `CompletionStage`s from methods of type `Object` that usually return the entity itself is something we'll eventually want to move away from (though it's OK for now).

So it feels like the return type was more correct before you made this change.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
```diff
@@ -240,12 +322,7 @@ else if ( isOptionalInstance && optional != event.getInstanceToLoad() ) {
 	 *
 	 * @return The result of the proxy/load operation.
 	 */
-	private CompletionStage<Object> proxyOrLoad(
-			final LoadEvent event,
-			final EntityPersister persister,
-			final EntityKey keyToLoad,
-			final LoadEventListener.LoadType options) {
-
+	private Object proxyOrLoad(final LoadEvent event, final EntityPersister persister, final EntityKey keyToLoad, final LoadEventListener.LoadType options) {
```
gavinking, line 325: @DavideD same comment here: I liked the previous return type a lot better.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
gavinking: @DavideD And I see a similar change was made to three other methods. So I suppose you had a good reason for wanting to do that, but I wonder what it was.

DavideD: I think the problem is that sometimes ORM calls `org.hibernate.internal.SessionImpl#internalLoad` that `callsDefaultRxLoadEventListener#onLoad` instead of the reactive `rxOnLoad`.
rxOnLoad and onLoad do basically the same thing but because onLoad is not reactive I cannot handle a CompletionStage withtout waiting for its completion first. My assumption is that a CompletionStage will only be returned when the method doLoad needs to execute a query on the db. In all the other cases it returns a proxy and the call to onLoad is not a problem.

That's why I've changed it the doLoad and the other methods to return an object, this way I can then check if it's a CompletionStage or not and throw an exception if I'm in the onLoad method. Basically, it is to avoid calling the .get() method and shows a blocking point in the code.

These are all private methods so I didn't expect it to be that important. Plus, I like that the final CompletionStage is created just before returning from rxOnLoad.

DavideD: But maybe there was a simple way to do it.

gavinking: I'm workin on it now.

gavinking: @DavideD See https://github.com/hibernate/hibernate-rx/pull/88.

Issue #55, opened 2020-02-06
Create a job on CI


DavideD:
We need to make sure that the project builds

aguibert:
Closing this issue since we now have a TravisCI pipeline in place.

CC @DavideD

DavideD:
Thanks @aguibert, Can you link on the right the correct PR under `Linked Pull Requests`, please?

PR #81, opened 2020-04-09
Create CI pipeline and add Testcontainers

aguibert:
Redelivery of PR #78 and #80

### DIFF
--- a/hibernate-rx-core/src/test/java/org/hibernate/rx/containers/PostgreSQLDatabase.java
+++ b/hibernate-rx-core/src/test/java/org/hibernate/rx/containers/PostgreSQLDatabase.java
@@ -1,0 +1,39 @@
+package org.hibernate.rx.containers;
+
+import org.testcontainers.containers.PostgreSQLContainer;
+
+public class PostgreSQLDatabase {
+
+	private static final boolean USE_DOCKER = Boolean.getBoolean( "docker" );
+
+	/**
+	 * Holds configuration for the PostgreSQL database contianer. If the build is run with <code>-Pdocker</code> then
+	 * Testcontianers+Docker will be used.
+	 * <p>
+	 * TIP: To reuse the same containers across multiple runs, set `testcontainers.reuse.enable=true` in a file located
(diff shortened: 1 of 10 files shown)


hibernate-rx-core/src/test/java/org/hibernate/rx/containers/PostgreSQLDatabase.java
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	public static final PostgreSQLContainer<?> postgresql = new PostgreSQLContainer<>()
+			.withUsername( "hibernate-rx" )
+			.withPassword( "hibernate-rx" )
+			.withDatabaseName( "hibernate-rx" )
+			.withReuse( true );
+
+	public static String getJdbcUrl() {
+		if ( USE_DOCKER ) {
+			postgresql.start();
```
DavideD, line 24: I think you should add a comment to explain that `postgresql.start();` is required or  `postgresql.getJdbcUrl()` will throw an exception.

aguibert:
@DavideD review comments are now addressed so this PR should be ready for final approval now

aguibert:
Closes #55

gavinking:
OK, I tried this out and it worked great. With `-Pdocker`, the tests run quite a lot slower on my machine, but I suppose that's to be expected. The important thing is that they still run great without Docker.

Thanks for this @aguibert!

aguibert:
Thanks for trying it out and reviewing @gavinking! 

By default, it will start containers for each run and then stop them automatically in a JVM shutdown hook. This is great for CI environments, but not so great for local development (because you are constantly starting/stopping containers on each test run).

To fix this, there is a notion of "reusable containers" which will allow to the containers to be reused across multiple runs. Since this is dangerous for CI environments and should only be enabled on local dev workstations, it needs to be configured as described here:
https://github.com/hibernate/hibernate-rx/pull/81/files#diff-04c6e90faac2675aa89e2176d2eec7d8R75
If you do that config step, then the test runtime should be on-par with manually starting the DBs yourself (after the 1st run)

PR #82, opened 2020-04-09
35 Fetch bidirectional lazy one-to-many associations

DavideD:
The code could be simplified by changing some methods in ORM but I would prefer to do it as a separate issue and first make sure that this solution makes sense.

The caveat is that this patch skip the path that should resolve collection of elements. But we agreed that it's not necessary to begin with so, hopefully, it should be fine.

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
@@ -395,17 +395,18 @@ public final class Cascade<C> {
 			cascadePoint = CascadePoint.AFTER_INSERT_BEFORE_DELETE_VIA_COLLECTION;
 		}
 
+		// TODO: Collection of elements is currently not supported
 		//cascade to current collection elements
-		if ( elemType.isEntityType() || elemType.isAnyType() || elemType.isComponentType() ) {
-			cascadeCollectionElements(
-				componentPathStackDepth,
-				child,
-				type,
-				style,
-				elemType,
-				persister.isCascadeDeleteEnabled()
-			);
-		}
+//		if ( elemType.isEntityType() || elemType.isAnyType() || elemType.isComponentType() ) {
+//			cascadeCollectionElements(
+//				componentPathStackDepth,
+//				child,
+//				type,
+//				style,
+//				elemType,
+//				persister.isCascadeDeleteEnabled()
+//			);
+//		}
 
 		cascadePoint = elementsCascadePoint;
 	}
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
@@ -1,13 +1,42 @@
 package org.hibernate.rx.impl;
 
-import org.hibernate.*;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import javax.persistence.EntityNotFoundException;
+import javax.persistence.LockModeType;
+
+import org.hibernate.CacheMode;
+import org.hibernate.HibernateException;
+import org.hibernate.JDBCException;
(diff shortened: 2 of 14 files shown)


gavinking:
@DavideD Nice! I'm looking at this now.

gavinking:
> The caveat is that this patch skip the path that should resolve collection of elements.

Sure, that's totally fine: we definitely don't need `@ElementCollection` for now.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
```diff
@@ -395,17 +395,18 @@ private void cascadeCollection(
...
+//		if ( elemType.isEntityType() || elemType.isAnyType() || elemType.isComponentType() ) {
+//			cascadeCollectionElements(
+//				componentPathStackDepth,
+//				child,
+//				type,
+//				style,
+//				elemType,
+//				persister.isCascadeDeleteEnabled()
+//			);
+//		}
```
gavinking, line 409: Why is this code disabled? Cascade persist for `@OneToMany` is pretty popular.

gavinking: OK, I see, so here's the thing: that code isn't only for `@ElementCollection`, it's actually for all kinds of collections including `@OneToMany`s.

Unfortunately, `CascadeTest` was missing an assertion, which means the test didn't fail when it really should have.

I've pushed the missing assertion [here](54ccaa07397d59d94a847b90f8d74a13fd2b20da).

gavinking: Now, if I uncomment that code, it *almost* works, except that `CascadeTest` fails when it tries to cascade delete an unfetched many-to-one association. That looks reasonably fixable and I'm looking into it. For now you can make the test pass by adding:

    .thenCompose(v-> s3.fetch(node.parent.elements))

 at line 98 of `CascadeTest`.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
```diff
@@ -88,12 +123,42 @@ public Object immediateLoad(String entityName, Serializable id) throws Hibernate
 					} );
 		}
 		if ( association instanceof PersistentCollection ) {
-			//TODO: handle PersistentCollection (raise InitializeCollectionEvent)
-			throw new UnsupportedOperationException("fetch() is not yet implemented for collections");
+			PersistentCollection persistentCollection = (PersistentCollection) association;
+			return rxInitializeCollection( persistentCollection, false )
+					.thenApply( l -> Optional.ofNullable( (T) l ) );
```
gavinking, line 128: Very nice to see this working, @DavideD!

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
DavideD: Thanks, I will look into it

gavinking: @DavideD note that the problem is avoided for to-one associations via the call to `fetch()` at line 52 of `CascadingActions`. We just need something like that for collections as well. The only problem is I'm not quite sure _where_ to add it.

DavideD: > That looks reasonably fixable and I'm looking into it.

If you are working on cascade delete of an unfetched many-to-one association. could you create a new issue, please?

gavinking: Well I think we need it in this changeset, because I added a (currently-passing) test to master that will probably fail with this changeset.

gavinking: @DavideD try [this commit](https://github.com/gavinking/hibernate-rx/commit/6aa11cae058a9082b371de588dd048e4ae304cee), it might be all you need.

DavideD:
I've sent a new PR with the changes we discussed: https://github.com/hibernate/hibernate-rx/pull/91

PR #85, opened 2020-04-09
Convert from Smallrye Axle to Smallrye Mutiny

### LINKED ISSUE #84: Convert from Axle to Mutiny
According to @Sanne the strategy going forward should be to use Mutiny instead of Axle, so we should convert dependencies over and also upgrade to the newest version of Mutiny (0.0.14 as of writing this).

aguibert:


aguibert:
Currently blocked by PR #81, after it is merged I will rebase this one

gavinking:
Wait wait wait. Can you please explain exactly what's going on here?

The motivation for migration to Mutiny was to replace the use of `CompletionStage` with `Uni`/`Multi`, which AFAICS this patch doesn't seem to do at all.

Instead, it seems to incorporate a bunch of completely different changes for containerizing this tests, and as I've argued strongly before, I'm completely against requiring a container to run the tests.

If what I'm seeing is correct, and much of the code in this patch is not strictly necessary for migration away from the `CompletionStage` API, then I would like to see that part of it resubmitted as a separate PR that fully explains the purpose and nature of the changes, and I would like to see that changes which are _strictly required_ in order to migrate away from `CompletionStage` in standalone form.

README.md
```diff
@@ -61,13 +66,27 @@ and then type the following:
     create database `hibernate-rx`;
     create user `hibernate-rx` identified by 'hibernate-rx';
     grant all on `hibernate-rx`.* to `hibernate-rx`;
+    
+Finally, run `./gradlew test` from the `hibernate-rx` directory.
```
gavinking, line 70: Why has this option been removed?

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/PoolConnection.java
```diff
@@ -58,25 +58,25 @@ public PoolConnection(Pool pool, boolean showSQL) {
...
+		return pool.preparedQuery( sql, parameters ).convert().toCompletionStage();
 	}
 
 	@Override
 	public CompletionStage<RowSet<Row>> preparedQuery(String sql) {
 		if (showSQL) {
 			System.out.println(sql);
 		}
-		return pool.preparedQuery( sql );
+		return pool.preparedQuery( sql ).convert().toCompletionStage();
```
gavinking, line 79: What's the point of this change if we're just going to immediately convert everything to `CompletionStage`?

gavinking:
> Instead, it seems to incorporate a bunch of completely different changes for containerizing this tests, and as I've argued strongly before, I'm completely against requiring a container to run the tests.

OK, I see now that there are commits in this PR that are duplicates of commits in #81.

So setting that aside, I don't see the point of making this change: if I understand correctly, it adds an additional unnecessary dependency to Hibernate RX (Mutiny) without actually using anything that Mutiny offers.

DavideD:
Like @gavinking said. the purpose of this issue was to expose Mutiny. I agree that it would be nice to expose both options with a Facade.

What I was thinking was having Mutiny in all our APIs and then expose CompletionStage as a Facade in a different interface, though. This way we would be able to see if we can benefit form Mutiny and, if I remember correctly, it has methods to switch to CompletionStage quite easily

aguibert:
Closing this PR so we can discuss further on issue #84 instead. I'll raise a new PR with a proper set of changes once we discuss this properly on #84 first

PR #87, opened 2020-04-11
86 Reduce test logging during build

DavideD:
Issue: https://github.com/hibernate/hibernate-rx/issues/86

I think there is too much information on the console right now during the build and it hides which test has failed. I guess on CI it would be handy but we can find an alternative to enable it if we need it.

A bit unrelated but I've noticed that the execution time for each test doesn't appear. Is there a way to enable it?

Note that running `gradle build -i` will produce the same output when nothing is cached.


DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/87

PR #88, opened 2020-04-11
Cleanup DefaultRxLoadEventListener

gavinking:
This patch cleans up the typesafety problems in `DefaultRxLoadEventListener`, where `CompletionStage`s were assigned to the type `Object` and passed around.

See [this discussion](https://github.com/hibernate/hibernate-rx/pull/74#pullrequestreview-391098287).

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
@@ -4,6 +4,7 @@ import java.io.Serializable;
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
+import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
 import org.hibernate.LockMode;
 import org.hibernate.NonUniqueObjectException;
@@ -75,26 +76,30 @@ public class DefaultRxLoadEventListener implements LoadEventListener, RxLoadEven
 			throw new HibernateException( "Unable to locate persister: " + event.getEntityClassName() );
 		}
 
-		Object checkId = checkId( event, loadType, persister );
-		if ( checkId instanceof CompletionStage) {
+		CompletionStage<Void> checkId = checkId( event, loadType, persister );
+		if ( !checkId.toCompletableFuture().isDone() ) {
 			// This only happens if the object is loaded from the db
-			throw new UnsupportedOperationException( "Unexpected access to the db" );
+			throw new AssertionFailure( "Unexpected access to the database" );
 		}
 
 		try {
-			Object loaded = doOnLoad( persister, event, loadType );
-			if ( loaded instanceof CompletionStage ) {
+			CompletionStage<Optional<Object>> loaded = doOnLoad( persister, event, loadType );
+			if ( !loaded.toCompletableFuture().isDone() ) {
 				// This only happens if the object is loaded from the db
-				throw new UnsupportedOperationException( "Unexpected access to the db" );
+				throw new AssertionFailure( "Unexpected access to the database" );
 			}
 			else {
 				// Proxy
-				event.setResult( loaded );
+				loaded.toCompletableFuture().getNow( Optional.empty() ).ifPresent( event::setResult );
 			}
 		}
 		catch (HibernateException e) {
 			LOG.unableToLoadCommand( e );
 			throw e;
+		}
+
+		if ( event.getResult() instanceof CompletionStage ) {
+			throw new AssertionFailure( "Unexpected CompletionStage" );
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
```diff
@@ -74,27 +76,31 @@ public void onLoad(
 			throw new HibernateException( "Unable to locate persister: " + event.getEntityClassName() );
 		}
 
-		Object checkId = checkId( event, loadType, persister );
-		if ( checkId instanceof CompletionStage) {
+		CompletionStage<Void> checkId = checkId( event, loadType, persister );
+		if ( !checkId.toCompletableFuture().isDone() ) {
```
DavideD, line 80: @FroMage Mentioned in the past that there might be some problems with this operation (`toCompletableFuture()`) and some implementations could throw an exception. @FroMage Do I remember correctly?

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
```diff
@@ -74,27 +76,31 @@ public void onLoad(
...
 			}
 		}
 		catch (HibernateException e) {
 			LOG.unableToLoadCommand( e );
 			throw e;
 		}
+
+		if ( event.getResult() instanceof CompletionStage ) {
+			throw new AssertionFailure( "Unexpected CompletionStage" );
+		}
```
DavideD, line 103: Can this happen or is it a safety check?

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
```diff
@@ -74,27 +76,31 @@ public void onLoad(
...
 			// This only happens if the object is loaded from the db
-			throw new UnsupportedOperationException( "Unexpected access to the db" );
+			throw new AssertionFailure( "Unexpected access to the database" );
 		}
 
 		try {
-			Object loaded = doOnLoad( persister, event, loadType );
-			if ( loaded instanceof CompletionStage ) {
+			CompletionStage<Optional<Object>> loaded = doOnLoad( persister, event, loadType );
+			if ( !loaded.toCompletableFuture().isDone() ) {
```
DavideD, line 87: Out of curiosity, this would work if the db is super quick and the stage completes before the check. isn't it?

DavideD:
I left a couple of comments for some clarifications but it looks good.
Let's wait for @FroMage  to confirm that I'm wrong about the `toCompletableFuture` and then we can merge it.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
gavinking: I think it's OK, since we're only going to call it on a `CompletionStage` that we're created ourselves using `RxUtil.completedFuture()` or `RxUtil.nullFuture()`. We're never going to call it on a "real" `CompletionStage` that's coming from the JDBC driver.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
gavinking: It's just an assertion, to help detect bugs. It should never ever occur. (That's why I had it throw `AssertionFailure`.)

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
gavinking: Assuming there is no bug in our code, this should never be called on a `CompletionStage` coming from the DB. If it *is*, the worst that can possibly happen is that we don't notice the bug in our code.

DavideD:
Thanks a lot. I'm fine with merging this after has been rebased to the latest changes @gavinking
Maybe a  comment or two to highlight the assumptions we are making would be helpful (especially for new contributors).

gavinking:
Have to close this one due to the conflicts with Gail's later PR.

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java
FroMage: I don't remember mentioning that warning. I do that sort of stuff in several places and don't know of any better way to see if a `CS` is done. IMO that's OK.

Issue #75, opened 2020-03-10
Test IdClass


DavideD:
We need to add tests for when @IdClass is used. In particular, there is a line of code that we converted from ORM that I'm not sure it's correct: https://github.com/hibernate/hibernate-rx/blob/master/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxLoadEventListener.java#L141

The mapping to test this should be (but I haven't tried it yet) something like:
```
@Entity
@IdClass(AnEntityId)
public class AnEntity {
@Id
@OneToOne
private OtherEntity otherEntity;
}

public class AnEntityId implements Serializable {
private int otherEntityId;
// getter/setter
}

@Entity
public class OtherEntity {
@Id
private int id;
}
```
and/or
```
@Entity
public class AnEntity {
@Id
@OneToOne
private OtherEntity otherEntity;
}

@Entity
public class OtherEntity {
@Id
private int id;
}
```

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/89

Thanks a lot!

PR #89, opened 2020-04-13
#75 Test IdClass

gbadner:
#75  Added test cases, and fixed a bug.


gavinking:
@gbadner cool, thanks!

Looks like this conflicts with my #88, but easy to resolve, no problem.

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/89

I've inverted the positions of the two commits so that tests appears after the fix. I know that this is not how Test driven developments works but, this way, the build works for each commit

PR #91, opened 2020-04-15
35 Fetch bidirectional lazy one-to-many associations v2

DavideD:
Follows https://github.com/hibernate/hibernate-rx/pull/82

@gavinking I had to change the commit for `fetchLazyAssociationBeforeCascade` because the stages weren't executed in the right order: https://github.com/hibernate/hibernate-rx/compare/master...DavideD:35-fetch-one-to-many-v2?expand=1#diff-1e7a64ba50be1d2c2ccc008faf2e4883R86

It seems that now all the tests are succesfull

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
@@ -1,13 +1,42 @@
 package org.hibernate.rx.impl;
 
-import org.hibernate.*;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import javax.persistence.EntityNotFoundException;
+import javax.persistence.LockModeType;
+
+import org.hibernate.CacheMode;
+import org.hibernate.HibernateException;
+import org.hibernate.JDBCException;
+import org.hibernate.LazyInitializationException;
+import org.hibernate.LockOptions;
+import org.hibernate.MappingException;
+import org.hibernate.ObjectDeletedException;
+import org.hibernate.ObjectNotFoundException;
+import org.hibernate.TypeMismatchException;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.internal.StatefulPersistenceContext;
-import org.hibernate.engine.spi.EffectiveEntityGraph;
+import org.hibernate.engine.spi.PersistenceContext;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.event.internal.MergeContext;
 import org.hibernate.event.service.spi.EventListenerRegistry;
-import org.hibernate.event.spi.*;
+import org.hibernate.event.spi.DeleteEvent;
+import org.hibernate.event.spi.EventSource;
+import org.hibernate.event.spi.EventType;
+import org.hibernate.event.spi.FlushEvent;
+import org.hibernate.event.spi.InitializeCollectionEvent;
+import org.hibernate.event.spi.InitializeCollectionEventListener;
+import org.hibernate.event.spi.LoadEvent;
+import org.hibernate.event.spi.LoadEventListener;
+import org.hibernate.event.spi.MergeEvent;
+import org.hibernate.event.spi.PersistEvent;
+import org.hibernate.event.spi.RefreshEvent;
 import org.hibernate.graph.GraphSemantic;
 import org.hibernate.graph.RootGraph;
 import org.hibernate.graph.spi.RootGraphImplementor;
@@ -24,21 +53,17 @@ import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.proxy.LazyInitializer;
 import org.hibernate.rx.RxSession;
 import org.hibernate.rx.RxSessionInternal;
+import org.hibernate.rx.engine.impl.RxPersistenceContextAdapter;
 import org.hibernate.rx.engine.spi.RxActionQueue;
-import org.hibernate.rx.event.spi.*;
+import org.hibernate.rx.event.impl.DefaultRxInitializeCollectionEventListener;
+import org.hibernate.rx.event.spi.RxDeleteEventListener;
+import org.hibernate.rx.event.spi.RxFlushEventListener;
+import org.hibernate.rx.event.spi.RxLoadEventListener;
+import org.hibernate.rx.event.spi.RxMergeEventListener;
+import org.hibernate.rx.event.spi.RxPersistEventListener;
+import org.hibernate.rx.event.spi.RxRefreshEventListener;
 import org.hibernate.rx.persister.entity.impl.RxEntityPersister;
 import org.hibernate.rx.util.impl.RxUtil;
-
-import javax.persistence.EntityNotFoundException;
-import javax.persistence.LockModeType;
-import java.io.Serializable;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.CompletionStage;
-import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.function.Supplier;
 
 /**
  * An {@link RxSessionInternal} implemented by extension of
@@ -53,6 +78,16 @@ public class RxSessionInternalImpl extends SessionImpl implements RxSessionInter
 
 	public RxSessionInternalImpl(SessionFactoryImpl delegate, SessionCreationOptions options) {
 		super( delegate, options );
+	}
+
+	@Override
+	public PersistenceContext getPersistenceContext() {
+		return new RxPersistenceContextAdapter( super.getPersistenceContext() );
+	}
+
+	@Override
+	public PersistenceContext getPersistenceContextInternal() {
+		return new RxPersistenceContextAdapter( super.getPersistenceContextInternal() );
 	}
 
 	@Override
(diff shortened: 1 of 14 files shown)


gavinking:
> @gavinking I had to change the commit for `fetchLazyAssociationBeforeCascade` because the stages weren't executed in the right order: https://github.com/hibernate/hibernate-rx/compare/master...DavideD:35-fetch-one-to-many-v2?expand=1#diff-1e7a64ba50be1d2c2ccc008faf2e4883R86

Oh, OK, WDYM, I didn't expect that the order would matter.

But of course I was not really able to test it properly.

hibernate-rx-core/src/main/java/org/hibernate/rx/util/impl/RxUtil.java
```diff
@@ -22,7 +22,7 @@
 		return state.thenApply( v -> zipper.apply( results ) );
 	}
 
-	public static CompletionStage<Void> nullFuture() {
+	public static <T> CompletionStage<T> nullFuture() {
```
gavinking, line 25: Cool, I noticed that this was annoying.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
```diff
@@ -88,12 +123,46 @@ public Object immediateLoad(String entityName, Serializable id) throws Hibernate
 					} );
 		}
 		if ( association instanceof PersistentCollection ) {
-			//TODO: handle PersistentCollection (raise InitializeCollectionEvent)
-			throw new UnsupportedOperationException("fetch() is not yet implemented for collections");
+			PersistentCollection persistentCollection = (PersistentCollection) association;
+			return rxInitializeCollection( persistentCollection, false )
+					.thenApply( l -> Optional.ofNullable( (T) l ) );
```
gavinking, line 128: @DavideD this can be written more cleanly as `.thenApply(Optional::ofNullable))`, I think.

But no matter, I will make that change later if you like.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
```diff
@@ -88,12 +123,46 @@ public Object immediateLoad(String entityName, Serializable id) throws Hibernate
...
+					if ( e instanceof MappingException ) {
+						throw getExceptionConverter().convert( new IllegalArgumentException( e.getMessage() ) );
+					}
+					else if ( e instanceof RuntimeException ) {
+						throw getExceptionConverter().convert( (RuntimeException) e );
+					}
+					else if ( e != null ) {
+						return RxUtil.rethrow( e );
+					}
+					return v;
```
gavinking, line 159: I think we have a bunch of places where we do something quite similar to this. I need to review that and see if we can just factor out a util function. (But of course that doesn't affect this PR.)

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
```diff
@@ -189,7 +258,7 @@ private Boolean getReadOnlyFromLoadQueryInfluencers() {
 						null,
 						object,
 						isCascadeDeleteEnabled,
-						((StatefulPersistenceContext) getPersistenceContextInternal())
+						((RxPersistenceContextAdapter) getPersistenceContextInternal())
 								.isRemovingOrphanBeforeUpates(),
```
gavinking, line 262: We should promote `.isRemovingOrphanBeforeUpates()` to `public` in Hibernate core.

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
```diff
@@ -0,0 +1,383 @@
...
+		} catch (SQLException sqle) {
+			throw getFactory().getJdbcServices().getSqlExceptionHelper().convert(
+					sqle,
+					"could not load an entity batch: " + MessageHelper.infoString(
+							getEntityPersisters()[0],
+							queryParameters.getOptionalId(),
+							session.getFactory()
+					),
+					sql
+			);
```
gavinking, line 160: We have a teensy bit of code duplication here. Let's factor this `try`/`catch` into a function. I can do it after you merge.

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
```diff
@@ -0,0 +1,383 @@
...
+
+					if ( returnProxies ) {
+						// now get an existing proxy for each row element (if there is one)
+						final PersistenceContext persistenceContext = session.getPersistenceContextInternal();
+						for ( int i = 0; i < entitySpan; i++ ) {
+							Object entity = row[i];
+							Object proxy = persistenceContext.proxyFor( persisters[i], keys[i], entity );
+							if ( entity != proxy ) {
+								// force the proxy to resolve itself
+								( (HibernateProxy) proxy ).getHibernateLazyInitializer().setImplementation( entity );
```
gavinking, line 196: @DavideD how come this isn't just:

    ( (HibernateProxy) proxy ).getHibernateLazyInitializer().initialize();

gavinking:
@DavideD This all looks great! Excellent work!

My comments are very minor things: more of a todo list for myself for after you merge, OK?

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
DavideD: I'm still not used to this kind of syntax :)

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
DavideD: It think there is already a method in RxUtil that does that.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
DavideD: I'm preparing a PR for ORM to add another change of this type. I will include this

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
DavideD: I don't know, that's how it's done in Loader#getRowFromResultSet in ORM

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
gavinking: Oops, I already implemented it and was about to send a PR. But I guess it's better if you roll it into your PR. So I will scrap mine.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
gavinking: No, but I mean the whole flow of exception conversion. Don't worry, I'll have a look and see if we really have duplication or if they're all subtly different.

DavideD:
> Oh, OK, WDYM, I didn't expect that the order would matter.

With the original code:
```
CompletionStage<Void> cascade = new Cascade<>(CascadingActions.DELETE,CascadePoint.AFTER_INSERT_BEFORE_DELETE,
					persister, entity, transientEntities, session)
					.cascade();
return beforeDelete.thenCompose( v -> cascade );
```

This was working (I guess it didn't matter) for initializing the collection but caused an exception towards the end of CascadeTest. Here to be exact: https://github.com/hibernate/hibernate-rx/blob/master/hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeTest.java#L98 (when the listener for the remove action was called).

If you replace the commit with the original one you can see the failure: https://github.com/gavinking/hibernate-rx/commit/6aa11cae058a9082b371de588dd048e4ae304cee

I notice that something was off in that line because `cascade`  can run and complete before the `beforeDelete` stage. But what we want (I think) is to run the cascade stage after the beforeDelete. Otherwise, what's the point of using` thenCompose`?

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/91

Thanks a lot. I will let the clean up of the code to you. In the meanwhile I'm preparing a PR for ORM where I will :

* Change the visibility isRemovingOrphanBeforeUpates(): https://github.com/hibernate/hibernate-rx/pull/91/files#diff-2d0d2708490b345ca3b7b6ede5d6c1edR261
* Change visibility of registerNonExists: https://github.com/hibernate/hibernate-rx/pull/91/files#diff-9ee6d08b506d43ce038d371918166946R231
* Create a constructor in StatefulPersistenceContext: https://github.com/hibernate/hibernate-rx/pull/91/files#diff-39d9aba0451f34abefbb597ff88be189R47

Please, let me know if there is something else that needs to be included.

Thanks

DavideD:
Ah, I see.

gavinking:
> But what we want (I think) is to run the cascade stage after the `beforeDelete`. Otherwise, what's the point of using `thenCompose`?

Hrmph. Really? But -to-many associated elements get cascade-deleted *before* the deletion of the parent. So I think the fetches need to happen before the cascade.

gavinking:
oooooooh waaaayt I see ... OK OK OK now I understand what you're saying: because `Cascade` actually does most of its work *synchronously*, it's not enough to chain the completion stages, we have to make its work happen actually asynchronously.

OK, but now I think that probably means that we need to make `cascadeAfterDelete()` more asynchronous also.

DavideD:
> OK, but now I think that probably means that we need to make cascadeAfterDelete() more asynchronous also.

Yes, I didn't notice it before.

DavideD:
> OK, but now I think that probably means that we need to make cascadeAfterDelete() more asynchronous also.

Are you going to have a look at it? I think it deserves an issue. I would be nice to have a way to test it

gavinking:
Yes I will look at it, but sure, go ahead and open an issue.

DavideD:
https://github.com/hibernate/hibernate-rx/issues/92

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
gavinking: Fixed in 91446b6fcc4893ed3b152f60aaef0acebd47d825

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
gavinking: Actually fixing this properly is sorta harder than it looks. Not worth the effort.

hibernate-rx-core/src/main/java/org/hibernate/rx/impl/RxSessionInternalImpl.java
gavinking: Oh well, gee, there are a *lot* of places which do very-similar looking conversion, but some of them are subtly different (differences that are inherited from core `SessionImpl`) so I'm scared to touch this.

Issue #52, opened 2020-01-22
Optional decision


gavinking:
So I want to know are we really going to force `Optional<Entity>` down everyone's throat? I've been using it the last few day, and yew, it really is just as crappy as I figured it was, *nothing* like a proper option or nullable type.

But if that's where the whole Java world is going ... well, I dunno, I guess we should be consistent.

Thoughts?

gavinking:
@FroMage?

FroMage:
Don't ask me I hate it. But a lot of people like it. IMO it should be opt-in with both options.

gavinking:
Pfff. Well currently it's on methods like `find()` and `fetch()`, so pretty hard to opt out...

gavinking:
The thing about it is that it not only doesn't protect you from bugs, since almost all of the time all you're going to do is just call `get()` on it (calling `isPresent()` first is just a worse way to write `!=null`), but it actually can be a *source* of bugs since `Optional<Object>` is assignable to `Object`!

gavinking:
You can even write shit like this:

```if ( thing.isPresent() ) {
    ...
    if (thing == otherThing) { ... }
}`

From that point of view, you were safer *without it*.

gavinking:
I think I would lean towards removing it `Optional` from our APIs, unless someone speaks up in its favor.

gavinking:
OK, given the votes in favor from @vietj, @FroMage, @aguibert, and the lack of strong objection from anybody else, I think I'll go ahead and merge #93.

DavideD:
OK, sounds good to me

DavideD:
Plus, if we realize that it makes sense to have Optional somewhere we can always add the method later

gavinking:
Done.

PR #93, opened 2020-04-15
Get rid of Optional (#52)

gavinking:
As proposed in #52, this change kicks `Optional` out of our code base.

Justification:

- it doesn't really protect you from NPEs because nothing stops you from calling `get()` on an uninitialized one (unlike other langs, Java doesn't have sum or union types) ... even worse, nothing stops you from assigning `null` to the type `Optional`
- indeed, it actually causes more bugs than it avoids, since `Optional` is assignable to `Object`
- it makes you create extra local variables to assign the result of `get()` to (unlike Ceylon, Java doesn't have intersection types)
- its comes with a performance overhead (Java doesn't have value types or union types)
- it makes method signatures uglier (unlike other langs, Java doesn't have a syntax sugar for it)

Does anybody want to step up and defend this 'orrible thing?

### DIFF
--- a/hibernate-rx-api/src/test/java/org/hibernate/rx/RxSessionApiTest.java
+++ b/hibernate-rx-api/src/test/java/org/hibernate/rx/RxSessionApiTest.java
@@ -2,7 +2,6 @@ package org.hibernate.rx;
 
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.junit.jupiter.api.Test;
@@ -36,7 +35,7 @@ public class RxSessionApiTest {
 		session.find( GuineaPig.class, 1 )
 				.whenComplete( (result, err) -> {
 					try {
-						assertThat( (Optional) result ).isNotPresent();
+						assertThat( result ).isNull();
 						testContext.completeNow();
 					}
 					catch (Throwable e) {
@@ -53,7 +52,7 @@ public class RxSessionApiTest {
 		session.find( GuineaPig.class, pig.getId() )
 				.whenComplete( (result, err) -> {
 					try {
-						assertThat( (Optional) result ).isPresent().hasValue( pig );
+						assertThat( result ).isNotNull().isEqualTo( pig );
 						testContext.completeNow();
 					}
 					catch (Throwable e) {
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/RxAbstractEntityPersister.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/RxAbstractEntityPersister.java
@@ -30,7 +30,6 @@ import java.io.Serializable;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
 /**
@@ -271,7 +270,7 @@ public interface RxAbstractEntityPersister extends RxEntityPersister, OuterJoinL
 				sql = sql + " returning " + identifierColumnName;
 			}
 			return queryExecutor().updateReturning( sql, insert.getParametersAsArray(), factory )
-					.thenApply(Optional::get);
+					.thenApply( id -> id );
 		}
 		else {
 			//use an extra round trip to fetch the id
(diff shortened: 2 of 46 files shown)


hibernate-rx-api/src/test/java/org/hibernate/rx/RxSessionApiTest.java
```diff
@@ -35,7 +35,7 @@ public void testLoadNull(VertxTestContext testContext) throws Throwable {
 		session.find( GuineaPig.class, 1 )
 				.whenComplete( (result, err) -> {
 					try {
-						assertThat( result ).isNotNull();
+						assertThat( result ).isNull();
```
DavideD, line 38: Can you squash this in the previous commit, please?

DavideD:
Isn't the idea of Optional to allow a more functional approach to code (it has functions like `map`, `flatMap`, `ifPresent`  and so on)?

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/RxAbstractEntityPersister.java
```diff
@@ -281,9 +280,9 @@ default AbstractEntityPersister delegate() {
 							identifierColumnName,
 							Types.INTEGER
 					);
-			return queryExecutor().update( sql, insert.getParametersAsArray(), factory)
+			return queryExecutor().update( sql, insert.getParametersAsArray(), factory )
 					.thenCompose( v -> queryExecutor().selectLong(selectIdSql, new Object[0], factory) )
-					.thenApply(Optional::get);
+					.thenApply( id -> id );
```
DavideD, line 285: Do you need this last `thenApply`?

gavinking:
> Isn't the idea of Optional to allow a more functional approach to code (it has functions like `map`, `flatMap`, `ifPresent`  and so on)?

That's not what "functional" means. (Not for any _meaningful_ definition of "functional" anyhow.)

Look, we have a reasonable-sized codebase that uses `Optional` throughout. We don't call `map()` or `flatMap()` on it anywhere. And we have no reason to do so: the whole reason those operations have those names in ML or Haskell is that there is an abstraction (a type class) called `Functor` that lets you operate polymorphically over `List`s and `Maybe`s and all sorts of other things, letting you write a special sort of highly generic code that is utterly impossible to write in Java, without or without `Optional`.

So this is just cargo-cult stuff. The only people who think that calling the monomorphic function `Optional.flatMap()` is making their code more "functional" are people who don't actually know a single thing about functional programming.

hibernate-rx-core/src/main/java/org/hibernate/rx/persister/entity/impl/RxAbstractEntityPersister.java
gavinking: Yup, it does a widening cast.

gavinking: (This wouldn't be necessary if Java had declaration-site variance *grumble*.)

gavinking:
P.S. @DavideD If you like, I can write you a function `applyWhenNotNull()` that does _exactly_ what `Optional.flatMap()` does, but that doesn't require you to wrap every one of your objects in a garbage wrapper object.

gavinking:
> The only people who think....

(That comment wasn't aimed at you, of course, @DavideD, more at the people who designed this thing.)

DavideD:
> (That comment wasn't aimed at you, of course, @DavideD, more at the people who designed this thing.)

Sure, but even so... was it really helpful as far as comments go?

Anyway, my point is that `Optional` is what Java provides and some user  might find it useful in some circumstances. It's not that I am a big fan of it, I'm mostly trying to understand if users might find it useful to have it in the API.

DavideD:
As far as I can tell most programmers don't like it, so I'm not against removing them from the API. I'm not sure this is the right place to discuss it though, as people won't see it.

gavinking:
> I'm not sure this is the right place to discuss it though, as people won't see it.

Right, that's why I opened #52 a couple of months ago ... to try and get some feedback.

(And we have to make a decision on this ASAP.)

vietj:
Optional is evil

NathanQingyangXu:
There is a serious drawback for the `Optional` from JDK. It doesn't implement `Serializable` for JDK team wanted to avoid the burden to maintain the `Serializable` backward compatibility.
That being said, `Optional` is definitively useful in general scenario. Sometimes we wanna differentiate between the following two possibilities:
* something exists but with `null` value;
* something doesn't exist at all;

No idea about the `hibernate-rx` relevance, though.

aguibert:
+1 on removing optional

gavinking:
> Sometimes we wanna differentiate between the following two possibilities:
>
> * something exists but with `null` value;
> * something doesn't exist at all;

This is indeed something that arises *occasionally* (really the only case I know of is caching/memoizing), it's not a distinction we need to make in the API of hibernate-rx.

DavideD:
NIce, thanks

PR #97, opened 2020-04-27
Add gradle script for publishing as Maven publication

aguibert:
Adds the `hibernate-rx-api` and `hibernate-rx-core` modules as publishable modules. This makes it so when you do `./gradlew publish` or `./gradlew publishToMavenLocal` the build will publish the maven artifacts to your local Maven repo.

There is now a common publish script at publish.gradle and any sub-module that wants to publish a maven artifact can do `apply from: publishScript` in their build.gradle file.

### DIFF
--- a/hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
+++ b/hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
@@ -84,7 +84,7 @@ public interface RxSession {
 	/**
 	 * Asynchronously persist the given transient instance, first assigning
 	 * a generated identifier. (Or using the current value of the identifier
-	 * property if the <tt>assigned</tt> generator is used.) This operation
+	 * property if the <code>assigned</code> generator is used.) This operation
 	 * cascades to associated instances if the association is mapped with
 	 * {@code cascade="save-update"}
 	 *
@@ -140,7 +140,7 @@ public interface RxSession {
 	 * <ul>
 	 * <li>where a database trigger alters the object state upon insert or update
 	 * <li>after executing direct SQL (eg. a mass update) in the same session
-	 * <li>after inserting a <tt>Blob</tt> or <tt>Clob</tt>
+	 * <li>after inserting a <code>Blob</code> or <code>Clob</code>
 	 * </ul>
 	 *
 	 * @param entity a persistent or detached instance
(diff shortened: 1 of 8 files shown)


publish.gradle
```diff
@@ -0,0 +1,71 @@
...
+    if (performRelease) {
+        repositories {
+            maven {
+                url "https://oss.sonatype.org/service/local/staging/deploy/maven2"
+                credentials {
+                    username ossrhUsername
+                    password ossrhPassword
+                }
+            }
+        }
```
DavideD, line 28: I'm not sure this is going to work. We cannot publish our username and password, It should be read from the settings.xml. Also we publish the artifact on our server and then it will take care to sync with Maven central.

In ORM they've just worked on the release scripts, we should check with @dreab8 and use a similar approach.

hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
```diff
@@ -84,7 +84,7 @@
 	/**
 	 * Asynchronously persist the given transient instance, first assigning
```
DavideD, line 85: Unless I'm missing something. Changes to the javadoc should be in a separate commit because they have nothing to do with the build.

publish.gradle
DavideD: @dreab8  could you link us to the part in the build of ORM that releases, please?

DavideD: By the way, we an skeep the configuration of the remote repositories for now and only keep the local one.
We will add them later as a separate issues

dreab8: @DavideD check  https://github.com/hibernate/hibernate-orm/blob/master/release/release.gradle and https://github.com/hibernate/hibernate-orm/blob/master/gradle/publishing-repos.gradle

dreab8: you may be also interested in https://github.com/hibernate/hibernate-orm/blob/master/gradle/publishing-pom.gradle and https://github.com/hibernate/hibernate-orm/blob/master/gradle/published-java-module.gradle

DavideD: THanks

hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
aguibert: I can move them into a separate commit. They were related to publishing the javadoc artifacts because I had to make the javadoc HTML5 compatible

aguibert:
thanks for reviewing @DavideD, PR is updated with your review comments addressed

hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
DavideD: Thanks, it's still nice to have them in a separate commit so that we can explain why that change is required.

DavideD:
Merged https://github.com/hibernate/hibernate-rx/pull/97

Thanks a lot!

PR #98, opened 2020-04-27
Add LGPL 2.1 license

aguibert:
I noticed that there is no license currently associated with this project. Since Hibernate ORM uses LGPL 2.1 license, I assume we would want to use the same license here.

This PR adds the necessary LGPL 2.1 metadata and also adds a build status badge to the readme

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


DavideD:
Thanks. @Sanne?

gavinking:
We've been discussing this issue in email, but AFAICT that discussion has not yet converged and no decision has yet been made.

Sanne:
Right let's wait a little on this one. It's indeed most likely to be LGPL 2.1 but still waiting for this to be confirmed.

Sanne:
Unfreezing.. good to go with LGPL!

Sanne:
thanks @aguibert

Issue #92, opened 2020-04-15
DefaultRxDeleteEventListener#cascadeAfterDelete() needs to be chained correctly


DavideD:
See https://github.com/hibernate/hibernate-rx/pull/91

In short, right now the related CompletionStage for beforeDelete and afterDelete might not be executed in the right order

gbadner:
@gavinking, would it help if I came up with some failing test cases, and fixed this?

gavinking:
@gbadner a failing test case would be great because I don't have one right now ... but I sorta wanted to look at it myself 'cos I need to think a little bit carefully about the whole model for how cascade works now that we know for sure that we can't just do it all synchronously (I think we already realized that in 985e6376367ed290c60727df0c555b468a71c174, if you recall).

DavideD:
I've created this PR to fix the fact that we don't rethrow the exceptions in a couple of places: https://github.com/hibernate/hibernate-reactive/pull/114

PR #100, opened 2020-04-28
make event processing related to cascades more async

gavinking:
This is a proposed fix for #92, which makes the processing that occurs in session event listeners significantly more asynchronous. This wasn't what we were going for originally, but we've found that it's necessary.

I have a couple of doubts about this:

- I can't really guarantee that this code doesn't break anything, or even that it fixes the class of potential problems we'd identified: it passes the existing tests, but we didn't have any failing test, and indeed we really don't have much in the way of tests for recursive cascade.
- Since we've gone this far, we probably should go all the way and make the event listeners fully async.

### DIFF
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.rx.engine.impl;
 
+import org.hibernate.CacheMode;
 import org.hibernate.Hibernate;
 import org.hibernate.HibernateException;
 import org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor;
@@ -97,6 +98,24 @@ public final class Cascade<C> {
 	 * which is specific to each CascadingAction type
 	 */
 	public CompletionStage<Void> cascade() throws HibernateException {
+		return RxUtil.nullFuture().thenCompose( v -> {
+			CacheMode cacheMode = eventSource.getCacheMode();
+			if (action==CascadingActions.DELETE) {
+				eventSource.setCacheMode( CacheMode.GET );
+			}
+			final PersistenceContext persistenceContext = eventSource.getPersistenceContextInternal();
+			persistenceContext.incrementCascadeLevel();
+			try {
+				return cascadeInternal();
+			}
+			finally {
+				persistenceContext.decrementCascadeLevel();
+				eventSource.setCacheMode( cacheMode );
+			}
+		} );
+	}
+
+	private CompletionStage<Void> cascadeInternal() throws HibernateException {
 
 		if ( persister.hasCascades() || action.requiresNoCascadeChecking() ) { // performance opt
 			final boolean traceEnabled = LOG.isTraceEnabled();
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
@@ -103,15 +103,11 @@ public final class Cascade<C> {
 			if (action==CascadingActions.DELETE) {
 				eventSource.setCacheMode( CacheMode.GET );
 			}
-			final PersistenceContext persistenceContext = eventSource.getPersistenceContextInternal();
-			persistenceContext.incrementCascadeLevel();
-			try {
-				return cascadeInternal();
-			}
-			finally {
-				persistenceContext.decrementCascadeLevel();
+			eventSource.getPersistenceContextInternal().incrementCascadeLevel();
+			return cascadeInternal().whenComplete( (vv, e) -> {
+				eventSource.getPersistenceContextInternal().decrementCascadeLevel();
 				eventSource.setCacheMode( cacheMode );
-			}
+			} );
 		} );
 	}
 
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
@@ -464,13 +464,9 @@ public final class Cascade<C> {
 			//not really necessary, but good for consistency...
 			final PersistenceContext persistenceContext = eventSource.getPersistenceContextInternal();
 			persistenceContext.addChildParent( child, parent );
-			try {
-				CompletionStage<?> cascade = action.cascade(eventSource, child, entityName, context, isCascadeDeleteEnabled);
-				stage = stage.thenCompose(v -> cascade).thenAccept(vv->{});
-			}
-			finally {
-				persistenceContext.removeChildParent( child );
-			}
+			stage = stage.thenCompose( v -> action.cascade( eventSource, child, entityName, context, isCascadeDeleteEnabled ) )
+					.whenComplete( (vv, e) -> persistenceContext.removeChildParent( child ) )
+					.thenAccept( vv -> {} );
 		}
 	}
 
--- a/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxRefreshEventListener.java
+++ b/hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxRefreshEventListener.java
@@ -136,119 +136,126 @@ public class DefaultRxRefreshEventListener implements RefreshEventListener, RxRe
 
 		// cascade the refresh prior to refreshing this entity
 		refreshedAlready.add( object );
-		CompletionStage<Void> cascade =
-				new Cascade<>(CascadingActions.REFRESH,
-						CascadePoint.BEFORE_REFRESH,
-						persister,
-						object,
-						refreshedAlready,
-						source)
-						.cascade();
 
-		if ( e != null ) {
-			final EntityKey key = source.generateEntityKey( id, persister );
-			persistenceContext.removeEntity( key );
-			if ( persister.hasCollections() ) {
-				new EvictVisitor( source, object ).process( object, persister );
-			}
-		}
+		return cascadeMerge( source, persister, object, refreshedAlready )
+				.thenCompose(v -> {
 
-		if ( persister.canWriteToCache() ) {
-			Object previousVersion = null;
-			if ( persister.isVersionPropertyGenerated() ) {
-				// we need to grab the version value from the entity, otherwise
-				// we have issues with generated-version entities that may have
-				// multiple actions queued during the same flush
-				previousVersion = persister.getVersion( object );
-			}
-			final EntityDataAccess cache = persister.getCacheAccessStrategy();
-			final Object ck = cache.generateCacheKey(
-					id,
-					persister,
-					source.getFactory(),
-					source.getTenantIdentifier()
-			);
-			final SoftLock lock = cache.lockItem( source, ck, previousVersion );
-			cache.remove( source, ck );
-			source.getActionQueue().registerProcess( (success, session) -> cache.unlockItem( session, ck, lock ) );
-		}
+					if ( e != null ) {
+						final EntityKey key = source.generateEntityKey( id, persister );
+						persistenceContext.removeEntity( key );
+						if ( persister.hasCollections() ) {
+							new EvictVisitor( source, object ).process( object, persister );
+						}
+					}
 
-		evictCachedCollections( persister, id, source );
+					if ( persister.canWriteToCache() ) {
+						Object previousVersion = null;
+						if ( persister.isVersionPropertyGenerated() ) {
+							// we need to grab the version value from the entity, otherwise
+							// we have issues with generated-version entities that may have
+							// multiple actions queued during the same flush
+							previousVersion = persister.getVersion( object );
+						}
+						final EntityDataAccess cache = persister.getCacheAccessStrategy();
+						final Object ck = cache.generateCacheKey(
+								id,
+								persister,
+								source.getFactory(),
+								source.getTenantIdentifier()
+						);
+						final SoftLock lock = cache.lockItem( source, ck, previousVersion );
+						cache.remove( source, ck );
+						source.getActionQueue().registerProcess( (success, session) -> cache.unlockItem( session, ck, lock ) );
+					}
 
-		String previousFetchProfile = source.getLoadQueryInfluencers().getInternalFetchProfile();
-		source.getLoadQueryInfluencers().setInternalFetchProfile( "refresh" );
+					evictCachedCollections( persister, id, source );
 
+					String previousFetchProfile = source.getLoadQueryInfluencers().getInternalFetchProfile();
+					source.getLoadQueryInfluencers().setInternalFetchProfile( "refresh" );
 
-		// Handle the requested lock-mode (if one) in relation to the entry's (if one) current lock-mode
+					// Handle the requested lock-mode (if one) in relation to the entry's (if one) current lock-mode
 
-		LockOptions lockOptionsToUse = event.getLockOptions();
+					LockOptions lockOptionsToUse = event.getLockOptions();
 
-		final LockMode requestedLockMode = lockOptionsToUse.getLockMode();
-		final LockMode postRefreshLockMode;
+					final LockMode requestedLockMode = lockOptionsToUse.getLockMode();
+					final LockMode postRefreshLockMode;
 
-		if ( e != null ) {
-			final LockMode currentLockMode = e.getLockMode();
-			if ( currentLockMode.greaterThan( requestedLockMode ) ) {
-				// the requested lock-mode is less restrictive than the current one
-				//		- pass along the current lock-mode (after accounting for WRITE)
-				lockOptionsToUse = LockOptions.copy( event.getLockOptions(), new LockOptions() );
-				if ( currentLockMode == LockMode.WRITE ||
-						currentLockMode == LockMode.PESSIMISTIC_WRITE ||
-						currentLockMode == LockMode.PESSIMISTIC_READ ) {
-					// our transaction should already hold the exclusive lock on
-					// the underlying row - so READ should be sufficient.
-					//
-					// in fact, this really holds true for any current lock-mode that indicates we
-					// hold an exclusive lock on the underlying row - but we *need* to handle
-					// WRITE specially because the Loader/Locker mechanism does not allow for WRITE
-					// locks
-					lockOptionsToUse.setLockMode( LockMode.READ );
+					if ( e != null ) {
+						final LockMode currentLockMode = e.getLockMode();
+						if ( currentLockMode.greaterThan( requestedLockMode ) ) {
+							// the requested lock-mode is less restrictive than the current one
+							//		- pass along the current lock-mode (after accounting for WRITE)
+							lockOptionsToUse = LockOptions.copy( event.getLockOptions(), new LockOptions() );
+							if ( currentLockMode == LockMode.WRITE ||
+									currentLockMode == LockMode.PESSIMISTIC_WRITE ||
+									currentLockMode == LockMode.PESSIMISTIC_READ ) {
+								// our transaction should already hold the exclusive lock on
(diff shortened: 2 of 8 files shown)


hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
```diff
@@ -97,6 +98,24 @@ public Cascade(final CascadingAction<C> action,
...
+			if (action==CascadingActions.DELETE) {
+				eventSource.setCacheMode( CacheMode.GET );
+			}
+			final PersistenceContext persistenceContext = eventSource.getPersistenceContextInternal();
+			persistenceContext.incrementCascadeLevel();
+			try {
+				return cascadeInternal();
+			}
+			finally {
+				persistenceContext.decrementCascadeLevel();
```
DavideD, line 112: Is it OK ot decrement the cascade level even if cascadeInternal is not completed?

gavinking: Good question. I'm not sure. I *think* it's OK, and this was what we were doing previously, but I can't guarantee it was correct. Let me take a closer look.

gbadner:
@gavinking, sorry, I missed this somehow.

I've mostly ported a really hideous test case from ORM that exercises lots of cascades. Here is the ORM test: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/test/java/org/hibernate/test/annotations/cascade/multicircle/jpa/sequence/MultiCircleJpaCascadeSequenceTest.java#L28

Before this fix, persist worked properly but merge had some bug.

I just need to get the cleanup code working properly so I can do multiple tests within the same class, then I'll give that test a try with this fix.

gbadner:
@gavinking, my test fails using this branch as well. The failure happens after an entity (and all its assocations) is merged, when the session is flushed. I need to look more closely to see if something goes wrong during merge, or if the bug is in fush.

Do I have privileges to push a commit with my test to your branch?

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+ * associated entities (e.g., B cascades persist to D, but D does not cascade
+ * persist to B);
+ *
+ * b, c, d, e, f, and g are all transient unsaved that are associated with each other.
+ *
+ * When persisting b with ORM, the entities are added to the ActionQueue in the following order:
+ * c, d (depends on e), f (depends on d, g), e, b, g.
+ *
+ * Entities are inserted in the following order:
+ * c, e, d, b, g, f.
```
gbadner, line 60: `session#persist(b)` followed by flush inserts the entities in the same order as documented above.

When `b` and all its associated entities are transient and`session.merge(b)` is used to persist the entities, the following exception is thrown on flush:
`org.hibernate.TransientPropertyValueException: object references an unsaved transient instance - save the transient instance before flushing : org.hibernate.rx.CascadeComplicatedTest$D.e -> org.hibernate.rx.CascadeComplicatedTest$E

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+ *
+ * In the diagram, all associations are bidirectional;
+ * assocations marked with '*' cascade persist, save, merge operations to the
+ * associated entities (e.g., B cascades persist to D, but D does not cascade
+ * persist to B);
+ *
+ * b, c, d, e, f, and g are all transient unsaved that are associated with each other.
+ *
+ * When persisting b with ORM, the entities are added to the ActionQueue in the following order:
+ * c, d (depends on e), f (depends on d, g), e, b, g.
```
gbadner, line 57: `session.persist(b)` adds insert actions to `RxActionQueue` in the same order as documented above.

When `b` and all its associated entities are transient, `session.merge(b)` adds insert actions to `RxActionQueue` in a different order, and 2 `D` entities get added(?!?): `D (id ==2)`,  `c`, `f`, `D (id ==4)`, `e`, `b`, `g`.

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+		configuration.addAnnotatedClass( C.class );
+		configuration.addAnnotatedClass( D.class );
+		configuration.addAnnotatedClass( E.class );
+		configuration.addAnnotatedClass( F.class );
+		configuration.addAnnotatedClass( G.class );
+		return configuration;
+	}
+
+	@Test
+	public void testPersist(TestContext context) {
```
gbadner, line 84: This test works as expected.

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+
+	@Test
+	public void testMergeTransient(TestContext context) {
+		test(
+				context,
+				openSession()
+						.thenCompose(s -> s.merge(b)
+								.thenApply(bMerged -> s)
+						)
+						.thenCompose(s -> s.flush())
```
gbadner, line 102: `flush` results in:
`org.hibernate.TransientPropertyValueException: object references an unsaved transient instance - save the transient instance before flushing : org.hibernate.rx.CascadeComplicatedTest$D.e -> org.hibernate.rx.CascadeComplicatedTest$E`

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+
+	@Test
+	public void testMergeDetached(TestContext context) {
+		test(
+				context,
+				openSession()
+						.thenCompose(s -> s.persist(b))
+						.thenCompose(s -> s.flush())
+						.thenCompose(ignore -> openSession()
+								.thenCompose(s2 -> s2.merge(b))
```
gbadner, line 115: Merging a detached copy with all associations initialized results in:
`AssertionFailure( "Unexpected access to the database" )`

This happens when ORM tries to initialize `d.e` via `SessionImpl#internalLoad`.

hibernate-rx-core/src/test/java/org/hibernate/rx/CascadeComplicatedTest.java
```diff
@@ -0,0 +1,442 @@
...
+						.thenCompose(s -> s.persist(b))
+						.thenCompose(s -> s.flush())
+						.thenCompose(ignore -> check( openSession(), context ))
+						.thenAccept(ignore -> {
+							// Cascade-remove is not configured, so remove all associations.
+							// Everything will need to be merged, then deleted in the proper order
+							prepareEntitiesForDelete();
+						})
+						.thenCompose(v -> openSession())
+						.thenCompose(s2 -> s2.merge(b).thenApply(merged -> {
```
gbadner, line 137: Merging a detached entity with associations removed results in:
`AssertionFailure( "Unexpected access to the database" )`

This happens when ORM tries to initialize d.e via `SessionImpl#internalLoad`.

gbadner:
@gavinking, @DavideD, I've added `CascadeComplicatedTest` and added comments about what goes wrong.

I have not had much of a chance to try to troubleshoot. I can do that tomorrow.

gavinking:
@gbadner 

> I've mostly ported a really hideous test case from ORM that exercises lots of cascades.

That's awesome, thanks!

> Do I have privileges to push a commit with my test to your branch?

Yes of course!

gavinking:
@gbadner so the tests which fail, they also failed *before* this patch, is that right? Or did this patch make something worse?

gbadner:
@gavinking, it looks like the same tests failed before this fix. 

I think I understand why `SessionImpl#internalLoad` happens during `merge`, but not during `find`.

In general, the SQL generated for a loading a `B` entity using `find` joins all non-lazy associations in a single query. There are exceptions to this that don't affect this test case.

The SQL generated for loading an existing `B` entity that will be used for copying state from a detached entity being merged only joins non-lazy associations that have cascade-merge mapped. The other non-lazy associations are loaded by a separate query using `SessionImpl#internalLoad`.

@gavinking, I'm pretty sure you know this already since you probably wrote the original code.

The case I'm describing above probably falls under the scope of https://github.com/hibernate/hibernate-rx/issues/39.

I do think there is an issue with cascade order that gets exposed by `CascadeComplicatedTest#testMergeTransient`, which throws `org.hibernate.PropertyValueException`. 

There is specific logic in ORM that should be able to deal with inserts in the wrong order. Something is going wrong in hibernate-rx that inserts are not in the right order. 

Also, the test case only has one `D` instance. For some reason, hibernate-rx is trying to persist 2 `D` instances.

Do you think it is worthwhile for me to look into what is going wrong with `CascadeComplicatedTest#testMergeTransient`?

gavinking:
> The SQL generated for loading an existing `B` entity that will be used for copying state from a detached entity being merged only joins non-lazy associations that have cascade-merge mapped. The other non-lazy associations are loaded by a separate query using `SessionImpl#internalLoad`.

Right, OK, so if I'm understanding correctly, it sounds like this results from what I guess I would consider a "bad" mapping in the context of Hibernate Reactive.

That is to say, we should typically avoid mapping non-lazy associations, and make them all lazy except in very particular circumstances. (Note that this is also my recommendation with regular plain-old Hibernate, so I think it's fine to say that here too.)

Now, OK, I suppose you could argue that an alternative would be to say that we should adjust the behavior of the merge loader so that it loads non-lazies as well as things matched cascade-merge, but I'm not sure I think that's a good path to go down.

> I'm pretty sure you know this already since you probably wrote the original code.

Heh. I'm quite sure I knew it *once*, many years ago.

> There is specific logic in ORM that should be able to deal with inserts in the wrong order. Something is going wrong in hibernate-rx that inserts are not in the right order. 

Hrm. OK. Do you mean cascade inserts specifically?

> Also, the test case only has one `D` instance. For some reason, hibernate-rx is trying to persist 2 `D` instances.

That sounds .... bad.

> Do you think it is worthwhile for me to look into what is going wrong with `CascadeComplicatedTest#testMergeTransient`?

If you have time that would be awesome.

But how about changing the test code to remove non-lazy associations first, since that's something we know is not meant to work?

gbadner:
> But how about changing the test code to remove non-lazy associations first, since that's something we know is not meant to work?

I'll do this next...

gbadner:
@gavinking, I noticed that calling `RxSession#fetch on a lazy entity proxy initializes the target of the proxy, then returns that target (instead of the proxy).

It that what you expect? I found it a little surprising.

I was thinking that `RxSession#fetch` would be like calling `Hibernate.initialize(proxy)`, which returns `void`.

gavinking:
> It that what you expect? I found it a little surprising.

I don't recall whether I intended this when I first wrote the code, or whether I just didn't realize that `LoadEventListener.IMMEDIATE_LOAD` has `setNakedEntityReturned( true )`. Whatever, I certainly didn't *document* this behavior so let's just assume it was an oversight.

What do we *want* to happen here?

- It would certainly sometimes be nice to be able to get a proxy-free reference that you can safely downcast, and the explicit call to `fetch()` makes that a "safe" behavior.
- However, it would mean you can't safely do stuff like `==` inside the `fetch()` callback, and
- it's probably not what people are going to expect.

So I guess I would say you're right, and this is something we should change.

But probably we should to add a second method `RxSession.unproxy()` which does `Hibernate.unproxy()` does for non-reactive code?

And now that makes me wonder if it's bad that `fetch()` isn't named `initialize()`....

gavinking:
@gbadner see #109

gbadner:
Thanks @gavinking . 

Either `#fetch` or `#initialize` are fine with me. Should the method be changed to return `RxSession`?

gbadner:
>> There is specific logic in ORM that should be able to deal with inserts in the wrong order. Something is going wrong in hibernate-rx that inserts are not in the right order.

> Hrm. OK. Do you mean cascade inserts specifically?

@gavinking, the logic is used whenever an insert action is added, so it could be `persist`, `merge`, `flush`, or other Hibernate-specific events that add insert actions.

If you're interested, take a look at the following:

- [`ActionQueue#addInsertAction`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java#L253)
- [`ActionQueue#addResolvedEntityInsertAction`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/spi/ActionQueue.java#L277)
- [`UnresolvedEntityInsertActions`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/action/internal/UnresolvedEntityInsertActions.java)

The same logic could probably be used for deletes as well. No one ever complained about issues deleting entities, so I never extended it to that case. It's always been in the back of my mind to do that though.

gbadner:
@gavinking , I see the code I mentioned in my previous comment has been copied into `RxActionQueue`.

I changed the eager many-to-one associations to be lazy, and the same tests are still failing. A couple are failing for a different reason though.

There is definitely something going wrong while merging a transient `B`. I can see that the "extra" `D` entity has a `bCollection` containing an extra `B` entity.

I'm looking into this...

gbadner:
I just pushed a commit that fixes the weird merge issue that resulted in multiple `D` and `B` entities. The problem was that the `MergeContext` was being cleared before all the cascades completed. As a result, when those entities were encountered in cascades after `MergeContext` was cleared, it was assumed that they were new entities.

There are still failures showing up with lazy associations. I'm still looking at those.

gbadner:
I pushed a fix to `CascadeComplicatedTest`. Now `#testMergeTransient` passes.

gbadner:
@gavinking, I've changed the many-to-ones in `CascadeComplicatedTest` to be lazy.

I found a bug that caused merging an initialized lazy property to fail. I've also fixed that issue.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
gbadner: I think this could cause problems. I might be able to come up with a test case. Would that be worthwhile?

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxRefreshEventListener.java
```diff
@@ -136,119 +136,126 @@ public void onRefresh(RefreshEvent event, Map refreshedAlready) throws Hibernate
...
-				}
-				else {
-					source.setReadOnly(result, e == null ? source.isDefaultReadOnly() : e.isReadOnly());
-				}
-			}
-
-			UnresolvableObjectException.throwIfNull(result, id, persister.getEntityName());
-		})
-		.whenComplete( (v,t) -> source.getLoadQueryInfluencers().setInternalFetchProfile(previousFetchProfile) );
+		return cascadeMerge( source, persister, object, refreshedAlready )
```
gbadner, line 140: Shouldn't this be called `cascadeRefresh`?

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxRefreshEventListener.java
```diff
@@ -136,119 +136,126 @@ public void onRefresh(RefreshEvent event, Map refreshedAlready) throws Hibernate
...
+									}
+								}
+
+								UnresolvableObjectException.throwIfNull(result, id, persister.getEntityName());
+							})
+							.whenComplete( (vv,t) -> source.getLoadQueryInfluencers().setInternalFetchProfile(previousFetchProfile) );
+				} );
+	}
 
+	private CompletionStage<Void> cascadeMerge(
```
gbadner, line 246: Shouldn't this be called `cascadeRefresh`?

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
gavinking: @gbadner if I'm not mistaken the `cascadeLevel` is only used for checking that the user didn't do Bad things and throwing exceptions, so I don't think we invest time in writing a test for this.

On the other hand, I think I agree with you that&mdash;now that `cascadeInternal()` is doing more of its work asynchronously&mdash;this should be done in `handle()`. I'll make that change.

gavinking:
@gbadner 

> The problem was that the `MergeContext` was being cleared before all the cascades completed.

Nice catch! Thanks!

hibernate-rx-core/src/main/java/org/hibernate/rx/event/impl/DefaultRxRefreshEventListener.java
gavinking: Yes, sorry, thanks.

gavinking: Fixed by 68dddeeb732a81ec7c62d116b4361236b427a61a

gavinking:
@gbadner 

> I found a bug that caused merging an initialized lazy property to fail. I've also fixed that issue.

OK, that looks good to, I guess. 

FTR: d46ae21102ea264e5f8403b060f4ec027f2cb736

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
gavinking: Oh well, now I look more closely at this, the logic was just plain wrong. Sorry @DavideD, I didn't realize at first.

Fixed by 35bb3628692c7c7de93e2981879aa4ff8c623c1b

gavinking:
Alright, @DavideD @gbadner, I think I'm happy with this now.

Should we merge it? Or Davide would you prefer that I wait until after you have pushed your current work?

DavideD:
@gbadner Could you have a last look at this, please? If it's fine for you feel free to merge it. Thanks.

gbadner:
@gavinking, @DavideD, I just added another commit to Ensure cascade order in Cascade#cascadeToOne is correct: [c6dfc0b952705dbc3e58f1358ca1ba4f274e0c67](https://github.com/hibernate/hibernate-reactive/pull/100/commits/c6dfc0b952705dbc3e58f1358ca1ba4f274e0c67).

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
```diff
@@ -464,13 +464,9 @@ private void cascadeToOne(
...
 			persistenceContext.addChildParent( child, parent );
-			try {
-				CompletionStage<?> cascade = action.cascade(eventSource, child, entityName, context, isCascadeDeleteEnabled);
-				stage = stage.thenCompose(v -> cascade).thenAccept(vv->{});
-			}
-			finally {
-				persistenceContext.removeChildParent( child );
-			}
+			stage = stage.thenCompose( v -> action.cascade( eventSource, child, entityName, context, isCascadeDeleteEnabled ) )
+					.whenComplete( (vv, e) -> persistenceContext.removeChildParent( child ) )
```
DavideD, line 468: If there is an exception and you don't rethrow it, it will dissapear

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-reactive/pull/100

Thanks!

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/Cascade.java
gavinking: @DavideD is that really the case with `whenComplete()`? Isn't it only `handle()` that works like that?

DavideD: Maybe you are right, I will check

Issue #95, opened 2020-04-21
RxPersistenceContextAdapter should extend StatefulPersistenceContext


DavideD:
Right now it implements `PersistenceContext` but it forces use to delegate a lot of code. It requires some changes in ORM

gavinking:
Yep!

gbadner:
Fixed.

gavinking:
Great, thanks!

PR #101, opened 2020-04-29
#95 extend stateful persistence context

gbadner:
https://github.com/hibernate/hibernate-rx/issues/95

This PR relies on:
https://github.com/gbadner/hibernate-core/tree/hibernate-rx-requests


gbadner:
The build failed because hibernate-core 5.5.0-SNAPSHOT was not found.

I'll look into this tomorrow, if someone doesn't get to it before me.

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
```diff
@@ -72,9 +73,18 @@ public RxOuterJoinLoader(SessionFactoryImplementor factory, LoadQueryInfluencers
...
+						RxUtil.rethrow(e);
 					}
+					return list;
+				})
+				.thenCompose(list -> {
+					// only initialize non-lazy collections after everything else has been refreshed
+					return ((RxPersistenceContextAdapter) persistenceContext ).rxInitializeNonLazyCollections()
+							.thenApply(v -> list);
+				})
+				.handle( (list, e) -> {
```
gbadner, line 87: I imagine there's a more elegant way to do this than `handle( ... ).thenCompose( ... ).handle( ... )`.

Suggestions?

gavinking: My suggestions:

1. Can you suck the calls to `setDefaultReadOnly()` into `rxInitializeNonLazyCollections()`? They seem be called around every invocation? Then you wouldn't need `handle()` at all, if I'm not mistaken.
2. For single-line anonymous functions I would prefer not using `{ return .... }`. Minor thing.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContextAdapter.java
```diff
@@ -44,624 +22,41 @@
...
-	@Override
-	@Deprecated
-	public Map getEntitiesByKey() {
-		return delegate.getEntitiesByKey();
-	}
-
-	@Override
-	public Map.Entry<Object, EntityEntry>[] reentrantSafeEntityEntries() {
-		return delegate.reentrantSafeEntityEntries();
+		final NonLazyCollectionInitializer initializer = new NonLazyCollectionInitializer();
```
gavinking, line 38: @gbadner where is this class `NonLazyCollectionInitializer` defined?

DavideD: https://github.com/hibernate/hibernate-rx/blob/e2517f41dca045972cf639f82a1940586e25fcb1/hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContextAdapter.java#L43

gavinking: Ah, thanks! Didn't see it there.

hibernate-rx-core/src/main/java/org/hibernate/rx/engine/impl/RxPersistenceContextAdapter.java
```diff
@@ -44,624 +22,41 @@
...
+		initializeNonLazyCollections( initializer );
```
DavideD, line 39: Just checking but I assume this method is in the superclass, isn't it?

DavideD: My bad, I didn't see the link to your ORM branch

hibernate-rx-core/src/main/java/org/hibernate/rx/loader/RxOuterJoinLoader.java
gbadner: @gavinking, 

> * Can you suck the calls to `setDefaultReadOnly()` into `rxInitializeNonLazyCollections()`? They seem be called around every invocation? Then you wouldn't need `handle()` at all, if I'm not mistaken.

Related to your question -- I'm not sure if the temporary setting above (out of view), in line 65, is in the proper place. Could this setting inadvertently affect future operations, since the setting only gets restored after `#doRxQuery` and friends complete?

I think it would be better to move this temporary "set" and "unset" to the code where objects are actually being added to the `PersistenceContext`.

I think this may also affect `persistenceContext.beforeLoad()` and `persistenceContext.afterLoad()`.

Am I understanding this correctly?  If so, I would prefer to deal with this in a separate issue.

> * For single-line anonymous functions I would prefer not using `{ return .... }`. Minor thing.

I pushed a commit with that change.

BTW, when this gets approved, I'll squash the commits, and make sure the message refers to #95.

gbadner:
Depends on ORM issue: https://hibernate.atlassian.net/browse/HHH-13997
PR: https://github.com/hibernate/hibernate-orm/pull/3385

DavideD:
We still have wait for ORM for this one, isn't it?

DavideD:
@gbadner We merged the PR in ORM so it should be possible to merge this if there aren't any other changes you need to include. Thanks!

gavinking:
@gbadner we need to do the big package rename, and this is the last open PR before we do that.

gbadner:
@DavideD, @gavinking, this PR changes `hibernateOrmVersion` from '5.4.14.Final' to '5.5.0-SNAPSHOT'.

I'll assume that the version should be 5.4.16.Final.

DavideD:
Yes, the final version should be 5.4.16

gbadner:
Superseded by https://github.com/hibernate/hibernate-reactive/pull/124.

PR #106, opened 2020-05-07
Switch to use the Gradle wrapper version with sources

Sanne:
Useful when dealing with Gradle builds.


aguibert:
What is the advantage of using a gradle wrapper with sources? Is that just for debugging into gradle source code if needed? I doubt that will be necessary since the build is pretty straightforward. OTOH the "with sources" zip is larger so it will take longer for new contributors to download when they want to build locally.

Sanne:
The main reason I did it is that otherwise IDEA keeps nagging that I should switch to use this version :)

AFAIK it's also rather useful, just forgot what exactly as I normally use this version, but if you prefer I'll close this one.

aguibert:
makes no difference to me, was just curious why, but nobody likes IDE nagging so we better merge!

PR #107, opened 2020-05-07
Change Maven groupId to 'org.hibernate.reactive'

Sanne:


README.md
```diff
@@ -114,7 +114,7 @@ Hibernate or JPA.
 
 Add the following dependency to your project:
 
-	org.hibernate.rx:hibernate-rx-core:1.0.0-SNAPSHOT
+	org.hibernate.reactive:hibernate-rx-core:1.0.0-SNAPSHOT
```
aguibert, line 117: should the artifact IDs be changed to match also? e.g. `hibernate-reactive-core`

aguibert:
@Sanne should java package names also be changed from `rx --> reactive`?

DavideD:
@aguibert We are going to change the packages later because at the moment we are working on several PR and it would cause conflicts

Sanne:
@aguibert that's all valid suggestions but I was keeping this one change at a minimum, so to not interfere with large patches that Gavin and Davide are actively working on.

Sanne:
BTW just to clarify: I wanted to accelerate on the groupId being set already so that we could start setting up the release group on bintray.

README.md
gavinking: My 2c is that if the group ID is exactly identical to the artifact id, then we have a redundancy and what the hell are artifact ids even for?

PR #108, opened 2020-05-07
Set source compatibility to Java 8

aguibert:
I kept running into issues pulling int HibernateRX into other projects because I would accidentally build RX with Java 11 and the other projects were running with Java 8.

The build files should explicitly state what source/target compat we are aiming for, so that the produced artifacts are more consistent and do not differ based on what JDK is used to build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,8 +3,14 @@ plugins {
 }
 
 subprojects {
+    apply plugin: 'java-library'
+
     group = 'org.hibernate.reactive'
     version = '1.0.0-SNAPSHOT'
+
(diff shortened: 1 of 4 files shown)


Sanne:
thanks @aguibert, I'd approve this but it seems to conflict with the PR I just merged .. sorry.

Could you fix it and then push directly?

aguibert:
@Sanne fixed the merge conflict, should be ready for review+merge now

DavideD:
Rebased and merged https://github.com/hibernate/hibernate-rx/pull/108

THanks

PR #110, opened 2020-05-11
Allow configuration via Vert.x Pool

### LINKED ISSUE #105: Allow configuration via Vertx Pool
Currently it is only possible to configure RX via URL. However, it would also be good to configure RX by passing in an `io.vertx.sqlclient.Pool` instance.

This would have 2 benefits:
 - It will allow the same Vertx Pool to be used by other components. For example, in Quarkus you can configure a Pool via a set of properties and inject the `Pool` into your application. This would allow the same underling pool to be used directly by the application and by RX.
 - It is consistent with non-reactive case where you can configure via `AvailableSettings.URL` or `AvailableSettings.DATASOURCE`

I propose we add a new `AvailableRxSettings` interface with the following:

```java
public interface AvailableRxSettings extends org.hibernate.cfg.AvailableSettings {

    String VERTX_POOL = "hibernate.rx.connection.vertx.pool";

    // Can add more reactive-specific settings as needed...

}
```

aguibert:
Closes #105

### DIFF
--- a/hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
+++ b/hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
@@ -6,11 +6,17 @@
  */
 package org.hibernate.rx.configuration;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.internal.util.config.ConfigurationException;
 import org.hibernate.rx.cfg.AvailableRxSettings;
+import org.hibernate.rx.containers.DatabaseConfiguration;
 import org.hibernate.rx.containers.PostgreSQLDatabase;
 import org.hibernate.rx.service.RxConnectionPoolProviderImpl;
 import org.hibernate.rx.service.initiator.RxConnectionPoolProvider;
@@ -18,6 +24,7 @@ import org.junit.Test;
 
 import io.vertx.pgclient.PgPool;
 import io.vertx.sqlclient.Pool;
+import io.vertx.sqlclient.Row;
 
 public class RxConnectionPoolTest {
 
@@ -28,7 +35,18 @@ public class RxConnectionPoolTest {
 		Map<String,Object> config = new HashMap<>();
 		config.put( AvailableRxSettings.VERTX_POOL, pgPool );
 		RxConnectionPoolProvider rxPool = new RxConnectionPoolProviderImpl( config );
-		rxPool.getConnection().close();
+		verifyConnectivity( rxPool );
+	}
+	
+	@Test
+	public void configureWithIncorrectPoolType() {
+		Map<String,Object> config = new HashMap<>();
+		config.put( AvailableRxSettings.VERTX_POOL, "I'm a pool!" );
+		try {
+			new RxConnectionPoolProviderImpl( config );
+			fail("Expected a ConfigurationException but one was not thrown");
+		} catch (ConfigurationException expected) {
+		}
 	}
 	
 	@Test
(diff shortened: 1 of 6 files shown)


hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hibernate.rx.cfg.AvailableRxSettings;
+import org.hibernate.rx.containers.PostgreSQLDatabase;
+import org.hibernate.rx.service.RxConnectionPoolProviderImpl;
+import org.hibernate.rx.service.initiator.RxConnectionPoolProvider;
+import org.junit.Test;
+
+import io.vertx.pgclient.PgPool;
+import io.vertx.sqlclient.Pool;
+
+public class RxConnectionPoolTest {
```
DavideD, line 22: Could you add a test to make sure that the right exception is thrown when the Object is of the wrong type?
Also what happens  if the credentials or something else is wrong

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
```diff
@@ -0,0 +1,57 @@
...
+		rxPool.getConnection().close();
+	}
+	
+	@Test
+	public void configureWithJdbcUrl() {
+		String url = PostgreSQLDatabase.getJdbcUrl();
+		Map<String,Object> config = new HashMap<>();
+		config.put( AvailableSettings.URL, url );
+		RxConnectionPoolProvider rxPool = new RxConnectionPoolProviderImpl( config );
+		rxPool.getConnection().close();
```
DavideD, line 40: I'm not sure what you are testing here.
The url is a static string the we create specifically for tests.

Does it even fail if the credential or something is wrong with in the JDBC? I think nothing happens in getConnection() unless you run a query. And close() is a null method.

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+	
+	@Test
+	public void configureWithCredentials() {
+		// Set up URL with invalid credentials so we can ensure settings are used
```
DavideD, line 45: Maybe I'm tired but I don't understand the logic here. If the credentials are wrong, shouldn't the test fail somewhere?

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
```diff
@@ -0,0 +1,57 @@
...
+		String url = PostgreSQLDatabase.getJdbcUrl();
+		url.replace( "user=" + PostgreSQLDatabase.USERNAME, "user=bogus" );
+		url.replace( "password=" + PostgreSQLDatabase.PASSWORD, "password=bogus" );
```
DavideD, line 48: You need to assign the result of the replace function:
`url = url.replace(...`

hibernate-rx-core/src/test/java/org/hibernate/rx/containers/PostgreSQLDatabase.java
```diff
@@ -5,6 +5,10 @@
 public class PostgreSQLDatabase {
 
 	private static final boolean USE_DOCKER = Boolean.getBoolean( "docker" );
+	
+	public static final String USERNAME = "hibernate-rx";
+	public static final String PASSWORD = "hibernate-rx";
+	public static final String DB_NAME = "hibernate-rx";
```
DavideD, line 11: I would like these credentials to be the same for all the databases, maybe we can move them in an interface. I wasn't planning to do it in this PR but you are already moving them to constants, so ...

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
DavideD: Note that the test is not affected by the change in the URL.

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
aguibert: Good idea, I'll add a test for invalid Vertx pool type.

As for invalid credentials, that would get passed through and when the first connection is made we'd get the error/message directly from the Vertx driver.

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
aguibert: This is indeed a pretty simple test. If this test fails then pretty much every other test would also be failing. But in case we ever switch to a different default way of configuring the tests, I thought it would be good to have this configuration path tested explicitly here.

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
aguibert: The intent of this test is to ensure that explicitly defined credentials (e.g. `config.put( AvailableSettings.USER, "admin" )`) take precedence over any user/password that might be part of the JDBC URL. In this test we purposely put bad values in the JDBC URL, but good values in the config map, and expect the connection to succeed which would prove the config map credentials take precedence.

hibernate-rx-core/src/test/java/org/hibernate/rx/configuration/RxConnectionPoolTest.java
aguibert: Good catch on the URL replace, thanks!

hibernate-rx-core/src/test/java/org/hibernate/rx/containers/PostgreSQLDatabase.java
aguibert: This might be possible for the current DBs, but for DB2 this will not work because DB_NAME cannot contain dash chars, and the username cannot contain dash chars and cannot exceed length 8.
I can move them to an interface now, but we will need to adjust the values as we support more DBs in the future with different requirements (such as DB2)

DavideD: Let's wait then and treat it as a separate issue

DavideD:
I've squashed the commits and merged it https://github.com/hibernate/hibernate-reactive/pull/110

Thanks

DavideD:
This is the commit: https://github.com/hibernate/hibernate-reactive/commit/01c0292f6abdefb55da675cca6292e772fef6560

DavideD:
Sorry it took me longer than anticipated to review it

Issue #104, opened 2020-05-01
schema export via Vert.x driver instead of JDBC


aguibert:
Currently the `RxPersistenceProvider` does not have schema support, which means things like automatic table creation need to be done by JDBC.

It seems a bit awkward to require users to add/configure a JDBC driver if their app is only using RX. In theory we should be able to do all of the same schema generation using RxConnection instead of JDBC Connection.

DavideD:
It is a bit weird. At the same time I would prefer not to have too many ways to configure the same connection string. In the future a user will probably be able to use ORM with Rx and it might want to use the Rx API only in some cases. It is also nice that we use properties that are already familiar to an ORM user.

DavideD:
Ah sorry. I've just realized that you weren't talking about the property names.

aguibert:
By the same token I don't think we should over-reach with making things convenient/familiar. For example, right now we support configuring RX with a `jdbc` protocol URL, which is convenient/familiar but not technically correct since we're not using JDBC.

DavideD:
Mostly, I'm wondering if there is a use case where  one would want to use
different configuration parameters.
I guess it could happen if the native reactive driver has some specific
properties.

Anyway, I guess we will have to have similar properties to the ORM ones but
with the`jdbc` part removed from the name.
And decide what happens if only one type of properties is set.

gavinking:
> Anyway, I guess we will have to have similar properties to the ORM ones but with the`jdbc` part removed from the name.

I think we should stick with the standard JDBC URL format because:

- it's standard
- it's well-documented by the database vendors
- everyone already knows it
- it makes it easier to configure Hibernate RX *and* plain Hibernate or plain JDBC at the same time

aguibert:
@DavideD I may have not properly described the issue in the OP, but how does config parameters come into play for schema gen?

To re-state the intent I had with this issue, I'm suggesting that the automatic drop/create table procedures can be done without needing to involve any JDBC drivers or jdbc-only config (if RX happens to re-use the same properties, that's fine though)

DavideD:
> Ah sorry. I've just realized that you weren't talking about the property names.

Yeah, got it. My bad. I read it too quickly the first time.

> It seems a bit awkward to require users to add/configure a JDBC driver if their app is only using RX. 

This sentence here made me think you were talking about the fact that we use the properties for the JDBC configuration that already existis in ORM. It's still nice that we talked about it and I'm not planning to change them for now.

[...]

PR #112, opened 2020-05-13
schema export via reactive driver

gavinking:
This patch depends on a proposed change to Hibernate core. Please do not merge.

For #104


hibernate-rx-core/src/main/java/org/hibernate/rx/service/RxGenerationTarget.java
```diff
@@ -0,0 +1,41 @@
...
+
+	@Override
+	public void prepare() {
+		connection = registry.getService( RxConnectionPoolProvider.class ).getConnection();
+	}
+
+	@Override
+	public void accept(String command) {
+		try {
+			connection.preparedQuery(command).toCompletableFuture().join();
```
gavinking, line 29: @vietj @FroMage @Sanne @DavideD WDYT of this?

I want to run Schema Export against the Vert.x driver, but I don't need it to actually be reactive. Also, I don't want to rewrite all of Hibernate's schema export stuff to use `CompletionStage` or whatever.

This seems to work fine, *except* with the `@RunWith(VertxUnitRunner.class)` where sometimes I get a harmless stacktrace from Vert.x at the end of my test.

Is this for some reason a terrible thing to be doing? Is there some other way we should be doing it?

gavinking:
@aguibert I think you were searching for this work.

Note that you might be better off with [this branch](https://github.com/gavinking/hibernate-rx/tree/113), since it fixes some bugs exposed by this PR.

aguibert:
@gavinking thanks for the pointer. I see that branch requires a snapshot of hibernate core. Should I be building a certain branch from core? Or better yet, is the required snapshot hosted on bintray or similar?

hibernate-rx-core/src/main/java/org/hibernate/rx/service/RxGenerationTarget.java
aguibert: @gavinking What's the stack trace complaining about? Is it the Vertx blocked thread checker?  If that's the case, I think we can safely ignore that because we are intending to be blocking here. To silence the blocked thread checker, we could use our own `Vertx` instance where we can set the blocked thread checker interval to a really high value like so:
```java
    VertxOptions vertxOptions = new VertxOptions();
    vertxOptions.setBlockedThreadCheckInterval(600);
    vertxOptions.setBlockedThreadCheckIntervalUnit(TimeUnit.MINUTES);
    Vertx vertx = Vertx.vertx(vertxOptions);
```

Sanne:
> @gavinking thanks for the pointer. I see that branch requires a snapshot of hibernate core. Should I be building a certain branch from core? Or better yet, is the required snapshot hosted on bintray or similar?

sorry, we no longer deploy nightly snapshots.

Project:
 - https://github.com/hibernate/hibernate-orm
Branch : `5.4`

To deploy a snapshot to maven local:
    
    ./gradlew publishToMavenLocal

gavinking:
>  I see that branch requires a snapshot of hibernate core. Should I be building a certain branch from core?

`master`

gavinking:
> To deploy a snapshot to maven local:
>    
>    ./gradlew publishToMavenLocal

That's actually super-slow. I prefer:

    ./gradlew hibernate-core:publishToMavenLocal

(Though you probably also need `hibernate-testing`.)

hibernate-rx-core/src/main/java/org/hibernate/rx/service/RxGenerationTarget.java
gavinking: Yeah it looks to me like Vert.x shuts down the thread because it thinks it's misbehaving.

Let's try your suggestion. Thanks.

Sanne:
N.B. "master" will most likely work as well, but master is now version
5.5.x -   Quarkus uses 5.4.x (as we don't have a stable release of 5.5.x
yet) so I'm backporting anything that Reactive needs to 5.4 too.

It's also a strategy to keep possibly invasive patches that other might
merge into master out of your way ;)

gavinking:
Now part of #118.

hibernate-rx-core/src/main/java/org/hibernate/rx/service/RxGenerationTarget.java
gavinking: @aguibert so I looked for a place to set this up in `VertxUnitRunner` and didn't find anything obvious. @vietj any suggestion?

PR #114, opened 2020-05-14
Lost exceptions

DavideD:
Follows: https://github.com/hibernate/hibernate-reactive/issues/100
A couple of minor changes so that we don't lose the exceptions.


gavinking:
@DavideD are you *sure* about this one?

My interpretation of the documentation of `whenComplete()` is that it automatically propogates the exception.

Aren't you thinking of `handle()`?

DavideD:
I just assumed that it  behaves the same way of `.handle()`. I think I will check with a test.

DavideD:
Sorry, my bad. It actually works as  intended,

Issue #15, opened 2020-01-14
HQL queries


gavinking:
The #1 priority right now is query support. This might be easier than it sounds if it turns out that we can tie in at the same level as the `ScrollableResults` stuff, i.e. using `Loader.loadSequentialRowsReverse()`/`Loader.loadSingleRow()`.

gavinking:
Closing. Remaining work is in #120. 

Thanks @DavideD!

PR #116, opened 2020-05-14
#15 Initial support for HQL queries

DavideD:
**It requires some changes in ORM: https://github.com/DavideD/hibernate-orm/tree/changes-for-rx**

I followed the same structure used by ORM but changing the methods so that they would use CompletionStage instead of the usual types.

I still have to convert the parameters used in queries to the one we use in the connectors, so at the moment using parameters won't work. Shouldn't be hard to solve.

I should be able to send a PR for executing SQL queries soon, but this was already becoming quite big and I've decided to split in a couple of  PRs.

It needs more testing but it should be enough to start a discussion

Thanks


hibernate-rx-api/src/main/java/org/hibernate/rx/RxSession.java
```diff
@@ -187,7 +191,7 @@
 	 *
 	 * @see javax.persistence.EntityManager#createQuery(String)
 	 */
-	<R> RxQuery<R> createQuery(Class<R> resultType, String queryString);
+	<R> RxQuery<R> createQuery(String queryString, Class<R> resultType);
```
gavinking, line 194: Is this better? Not sure I think it is.

DavideD: It's consistent with `Session#createQuery`. Do you really think that one is better than the other? I don't mind to remove this change.

gavinking: Well I guess the argument for putting it second is that it's an optional argument, and optional arguments generally go after required args.

OTOH, since the HQL string is almost always going to be much *longer* than the class name, I feel like I would much prefer to put the class name first.

But I guess since both `Session` and `EntityManager` do it this way, we should be consistent.

DavideD:
I've pushed force a small change. In case somebody is working on this.

DavideD:
Note: I meesed up the test for theRx AutoFlushEventListener... I'm fixing it

DavideD:
Fixed https://github.com/hibernate/hibernate-reactive/pull/116/commits/b46a31756c712382d5d90fa023feee3268e23c93

gavinking:
OK, honestly this looks great. Well done, @DavideD.

I just tested out a couple of extra things: joins, fetch joins, projections, grouping, and nothing broke.

The code looks fine to me.

I think we need to try and get this merged ASAP, without waiting for parameter support, so that we can also merge #118, and move onto other issues that this is currently blocking work on, but that depends on https://github.com/hibernate/hibernate-orm/pull/3401.

(Merging this means that for a time Hibernate Reactive is going to depend on a snapshot build of Hibernate ORM, but I think we can live with that. Alternatively we could add a `dev` branch where we integrate all current development until the next release of ORM.)

DavideD:
Merged a more up-to-date pr: https://github.com/hibernate/hibernate-reactive/pull/119

Issue #143, opened 2020-05-20
Check the version of the containers we use for testing


DavideD:
At the moment with podman we are using the latest version every time but the defaults for `TestContainers` are a bit behind that.

We need to update the tests so that they run with the exact version we expect to run and update the README accordingly and  the podman examples accordingly

DavideD:
Done

PR #146, opened 2020-05-20
[#143] Set the tag of the docker images to use

DavideD:
Otherwise we rely on whatever version TestContainer uses as default
  and it's not always the latest or could change between updates.

  I decided to use the family version where available but we might
  want to change this in the future and use the exact version to be
  extra safe.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/DB2Database.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/DB2Database.java
@@ -4,6 +4,8 @@ import org.testcontainers.containers.Db2Container;
 
 public class DB2Database {
 
+	public final static String IMAGE_NAME = "ibmcom/db2:11.5.0.0a";
+
 	/**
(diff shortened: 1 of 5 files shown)


aguibert:
Historically Testcontainers never changes the default label for exactly this reason -- updating could break users. But I still think it's a good idea to explicitly specify the version so that our docker (testcontainers) and podman devs are always using the same DBs

DavideD:
I really don't understand the way they choose defaults in TestContainers. If they never update the tag why not always use the latest and clarify it in the javadoc?

Anyway, @aguibert  did you also build the PR? GitHub seems to have some problem and cannot test it with the actions

gavinking:
> GitHub seems to have some problem and cannot test it with the actions

The CI is busted since the change to the DB2 database name.

DavideD:
Thanks, I will check

DavideD:
Something was wrong with GitHub. It's working now.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/146

PR #147, opened 2020-05-20
Refactor to not depend on any specific Vertx drivers

### LINKED ISSUE #128: Eliminate dependencies on specific Vertx drivers
Currently core depends on all of the individual Vertx drivers:

```
    compile 'io.vertx:vertx-pg-client:3.9.1'
    compile 'io.vertx:vertx-mysql-client:3.9.1'
    compile 'io.vertx:vertx-db2-client:3.9.1'
```

With a few minor exceptions, this is needed to create the DB-specific ConnectOptions object (e.g. `DB2ConnectOptions`).

Recently a new API was added to Vertx that will allow generic creation of the SqlConnectOptions like this:
```java
SqlConnectOptions opts = new SqlConnectOptions()
    .setHost("mydb.com")
    .setPort(1234)
    // etc...
Pool pool = Pool.pool(opts);
```

The only limitation with this generic approach is that there must be exactly 1 Vertx driver found on the classpath. Otherwise, an error is thrown.

So I think we have 3 options here:
1) Use the generic SPI described above (with the limitation that only 1 driver can be on the classpath)
2) Use the current approach, but load/instantiate the DB-specific ConnectOptions using reflection as to not have a compile dependency on all drivers
3) Use a hybrid of options (1) and (2). First try option (2), which will tolerate cases where multiple drivers are on the classpath. If none are found there, try option (1) which will account for any new Vertx drivers that we don't have explcit knowledge of in option (2) (e.g. the SQL Server driver).

I'm leaning towards option (3) since it's the most flexible. 
Any other thoughts on this @gavinking @Sanne @DavideD ?

aguibert:
Fixes #128

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -14,18 +14,16 @@ dependencies {
         exclude group: 'org.hibernate.common', module: 'hibernate-commons-annotations'
         exclude group: 'org.jboss.spec.javax.transaction', module: 'jboss-transaction-api_1.2_spec'
     }
-    testCompile ("org.hibernate:hibernate-testing:${hibernateOrmVersion}")
+    compile 'io.vertx:vertx-sql-client:3.9.1'
     
-    // TODO: Once we restructure our tests we can change these driver dependencies
-    // to the generic vertx-sql-client artifact so that core no longer depends on all drivers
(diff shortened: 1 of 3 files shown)


gavinking:
This LGTM.

Issue #94, opened 2020-04-16
Mutiny Session API


gavinking:
Alright, so I've been playing with Mutiny, very superficially, just to get a feel for how the reactive `Session` API would look with this.

So first, it seems that there are sorta two reasonable modes of usage. The first looks like this, and is the the way I think they want me to do it:

```java
openMutinySession()
        .onItem().produceUni( s -> s.persist( new GuineaPig( 10, "Tulip" ) ) )
        .onItem().produceUni( s -> s.flush() )
        .onItem().produceUni( v -> selectNameFromId( 10 ) )
        .onItem().invoke( selectRes -> context.assertEquals( "Tulip", selectRes ) )
```

The second way looks like this:

```java
openMutinySession()
        .flatMap( s -> s.persist( new GuineaPig( 10, "Tulip" ) ) )
        .flatMap( s -> s.flush() )
        .flatMap( v -> selectNameFromId( 10 ) )
        .map( selectRes -> context.assertEquals( "Tulip", selectRes ) )
```

My initial, superficial, impression is that I prefer the look of the second code fragment, because:

- Well, I guess I get bored of typing `.onItem()`.
- I don't love the name `produceUni()`, for a very specific reason: from my point of view of writing business code, there's little difference between `map()` and `flatMap()`, and thus I definitely like the fact that they have similar names.

Purely due to the name, `produceUni()` and `apply()` look like completely different operations, but that's not how I view them in my code. I would have thought that chaining `Uni`s would be like almost the most basic operation in Mutiny, and I would have expected it to have a really snappy name. But OK, I guess `flatMap()` is not *especially* snappy. I guess my real question is why is it `produceUni()` instead of `applyUni()`, which is definitely what I feel like I'm doing in my code?

OK, but fine, so anyway I'm rather happy with `map()` and `flatMap()` except for one little thing: AFAICS, there's no real equivalent to `invoke()` at the level of `map()` and `flatMap()`, which is a problem because sometimes I have `void` functions. If I had one, I think I would be about 98% happy, which is certainly a very high level of happiness for someone so curmudgeonly.

(Thankfully, Mutiny gets the return type of `invoke()` correct, one of my gripes with `CompletionStage` is that `thenAccept()` throws away my stream elements.)

@FroMage @cescoffier reactions?

gavinking:
P.S. I could push some code onto a branch if anyone else wants to try this out.

DavideD:
[...]

PR #149, opened 2020-05-20
Mutiny API

gavinking:
This work introduced a new parallel API for Mutiny, `Mutiny.Session`, `Mutiny.Query`, etc, and cleans up some mess associated with the `SessionFactory` and `SessionFactoryBuilder`.

See #94.


gavinking:
@FroMage take a look please and let me know if this is what you need for Panache.

gavinking:
Alright, so I'm done with my rebase/squash and some extra cleanups, so I guess this is ready to merge.

DavideD:
This is awesome!

gavinking:
Merged!

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseMutinyTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+	@Rule
+	public Timeout rule = Timeout.seconds( 3600 );
+
+	private Mutiny.Session session;
+	private ReactiveConnection connection;
+	private org.hibernate.SessionFactory sessionFactory;
+	private ReactiveConnectionPool poolProvider;
+
+	protected static void test(TestContext context, CompletionStage<?> cs) {
```
FroMage, line 35: I would expect a mutiny test to use `Uni` here, but :shrug: :)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -0,0 +1,250 @@
...
+	}
+//
+//	private Uni<Integer> cleanDB() {
+//		return connection().flatMap( conn -> Uni.createFrom().completionStage( conn.update( "DELETE FROM Pig" ) ) );
+//	}
+
+	private Uni<String> selectNameFromId(Integer id) {
+		return connection().flatMap(
+				connection -> Uni.createFrom().completionStage(
+						connection.select(
```
FroMage, line 35: Is it normal to have to use `CompletionStage` for `select`?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseMutinyTest.java
gavinking: Yes yes, at this point I had not finished work on the test harness but it is finished now, so if you would be looking at the current code, and not a 4-day-old pull request, you wouldn't need to see things like this :-P

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
gavinking: There is no Mutiny API for `ReactiveConnection`. It's an internal API, not supposed to be visible to the program. (You have to be an expert in Hibernate `Service`s to even get a reference to it.)

gavinking:
@FroMage this is where you should be looking:

<https://github.com/hibernate/hibernate-reactive/blob/master/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java>

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseMutinyTest.java
FroMage: Well, you raise the PR while I'm on PTO and I'm kind enough to review it even when you didn't wait for me to merge it, so don't complain too much ;)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
FroMage: OK, fair.

Issue #151, opened 2020-05-21
package layout


gavinking:
Before the first release, we need to improve the package layout.

- In general, I still find our package structure quite confusing.
- Worse, it's not very clear which packages are public APIs and which aren't.
- I still hate all the useless little `impl` qualifiers.

(Note that `impl` packages are fine when you have a package containing APIs, and then a package containing implementations of those APIs. That's not the function they're performing here.)

We currently only expose three APIs to clients: `Stage`, `Mutiny`, and `ReactiveConnectionPoolProvider`/`ReactiveConnection`. But it's almost impossible to see that by looking at our source code layout.

Instead of using lots of `impl` packages to indicate non-public APIs, I would propose the following top-level layout:

- `org.hibernate.reactive.stage` for the `CompletionStage` API
- `org.hibernate.reactive.mutiny` for the Mutiny API
- `org.hibernate.reactive.connection` for the abstraction over reactive connections
- `org.hibernate.reactive.internal` package for everything else

gavinking:
Alright, so I'm now *much* happier with the package structure, and I can finally start to see that codebase coalescing into a set of logical subsystems.

I don't think we need to introduce the `internal` package after all because we would wind up with `ReactiveIdentifierGenerator` in the (non-internal) `id` package exposing`ReactiveSession` from the internal `session` package. And I think we'll likely add other APIs which expose `ReactiveSession`.

I still want to get rid of some `impl`s, but I don't want to cause further disruption by doing it right now.

So I think I'm going to close this for now.

PR #152, opened 2020-05-21
source code organization

gavinking:
This pull request cleans up some inconsistency (and an error) in package naming. It also removes some single-class packages and moves some classes into more logical locations.

This is better, however, more work is still needed, see #151.


gavinking:
I've also deleted the unused class `FlatClassLoaderService`. @DavideD please yell at me if that class was needed for something.

DavideD:
I don't think so. It's probably a leftover from something else

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/152

Nice, Thanks!

Issue #154, opened 2020-05-21
allow replacement of the ReactiveConnectionPool


gavinking:
If we let users replace the `ReactiveConnectionPool`, they could in principle integrate different database drivers.

Sanne:
This is already possible, if the users aren't too scared to inject a custom implementation into the bootstrap registry: provided services take precedence over the ServiceInitiator.

So while this will be easy for us, for example in Quarkus, did you mean we should make it "easier" for end users?

Sanne:
I just see the PR now.

Are you aware you can inject custom service implementations at boot by doing:

    StandardServiceRegistry registry = new StandardServiceRegistryBuilder()
				.applySettings( properties )
				.addService( ReactiveConnectionPoolProvider.class, someCustomImplementationInstance );
				.build();

gavinking:
No it's not possible because there was no *well-defined*, *well-specified* API for doing it.

> So while this will be easy for us, for example in Quarkus, did you mean we should make it "easier" for end users?

Exactly. They need a well-defined published API for that.

> Are you aware you can inject custom service implementations at boot by doing:

I am indeed aware of that. That's the functionality my PR takes advantage of.

gavinking:
> No it's not possible

Also, up until yesterday it wasn't possible at all because `ReactiveConnection` leaked Vert.x-specific types in its APIs. But I already fixed that as a preliminary to this.

Sanne:
ah, so the gist of your change is that you shift the focus to a new `Service` which abstracts from `ReactiveConnection` ?

gavinking:
My change makes it clear that `ReactiveConnectionPool` and `ReactiveConnection` are user-implementable APIs, and that our Vert.x-based default implementation (`SqlClientPool` and `SqlClientConnection`) is "hidden" from the rest of Hibernate Reactive.

Previously this code was scattered almost at random across three different packages and it wasn't clear what was API and what was implementation.

gavinking:
Also if I'm not mistaken, there used to be a warning from Hibernate because the `StandardServiceInitiator` for the `ReactiveConnectionPool` claimed to be a service implementation for the implementation class instead of for the service API interface. (Even though the client code was correctly requesting the service API interface.)

DavideD:
Merged

gavinking:
Thanks, done.

Sanne:
Ok, sounds great. I'll start my second wave of conflict resolutions :)

[...]

PR #155, opened 2020-05-21
proposed extension point for connection pools

gavinking:
See #154.

I believe that this patch is enough to allow people to replace the `ReactiveConnectionPool` with their own implementation via the Hibernate `Service`s stuff.

@Sanne @DavideD WDYT?


Sanne:
@gavinking I don't understand the problem that this is solving?

gavinking:
@Sanne 

>  I don't understand the problem that this is solving?

It's for someone who wants to use a non-Vert.x SQL client underneath Hibernate Reactive. (We have already had people opening issues requesting that.)

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/155

Thanks!

PR #159, opened 2020-05-21
vertx package and SessionFactory.close()

gavinking:
This work:

- moves @Sanne's new code for managing the `Vertx` instance into a public package, and renames the classes/interfaces
- introduces `SessionFactory.close()`, to clean up vert.x and the `Pool`

Does this look good, @Sanne @DavideD?


example/src/main/java/org/hibernate/example/reactive/Main.java
```diff
@@ -79,7 +79,7 @@ public static void main(String[] args) {
 		)
 				.toCompletableFuture().join();
 
-		System.exit(0);
+		sessionFactory.close();
```
Sanne, line 82: whoa, why did you have a Sytem.exit there? :)

gavinking: 'Cos I had no way to shut down Vert.x before :-)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/VertxInstance.java
```diff
@@ -0,0 +1,34 @@
+package org.hibernate.reactive.vertx;
+
+import io.vertx.core.Vertx;
+import org.hibernate.service.Service;
+
+/**
+ * Used by {@link org.hibernate.reactive.pool.impl.SqlClientPool}
```
Sanne, line 7: :100:

example/src/main/java/org/hibernate/example/reactive/Main.java
Sanne: that's what I suspected.. so you already had a clue that something was leaking :)

gavinking: C'mon, all good Java programs call `System.exit()`, right @emmanuelbernard?

Sanne: now you're both in that club! :-)

emmanuelbernard: > C'mon, all good Java programs call `System.exit()`, right @emmanuelbernard?

It's an underrated API.

Issue #160, opened 2020-05-21
subselect fetching


gavinking:
Using `@Fetch(FetchMode.SUBSELECT)` produces:

```
java.lang.ClassCastException: org.hibernate.loader.collection.SubselectOneToManyLoader cannot be cast to org.hibernate.reactive.loader.collection.impl.ReactiveCollectionLoader
java.util.concurrent.CompletionException: java.lang.ClassCastException: org.hibernate.loader.collection.SubselectOneToManyLoader cannot be cast to org.hibernate.reactive.loader.collection.impl.ReactiveCollectionLoader
```

We must need a `ReactiveSubselectOneToManyLoader`.

PR #164, opened 2020-05-21
batch and subselect fetching

### LINKED ISSUE #51: batch fetching for entities and one to many associations
We have some code for batch fetching but we have no idea if it works (probably not completely) because there are no tests for it. We need to try this out.

gavinking:
Fixes #51.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveOuterJoinLoader.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveOuterJoinLoader.java
@@ -33,6 +33,7 @@ import static org.hibernate.reactive.adaptor.impl.QueryParametersAdaptor.toParam
  * </p>
  */
 public class ReactiveOuterJoinLoader extends OuterJoinLoader {
+
 	public ReactiveOuterJoinLoader(SessionFactoryImplementor factory, LoadQueryInfluencers loadQueryInfluencers) {
 		super(factory, loadQueryInfluencers);
 	}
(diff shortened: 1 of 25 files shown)


gavinking:
I've now added support for subselect fetching, fixing #160.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveOuterJoinLoader.java
```diff
@@ -109,184 +102,160 @@ public ReactiveOuterJoinLoader(SessionFactoryImplementor factory, LoadQueryInflu
...
-			if ( object != null ) {
-				instanceAlreadyLoaded( rs, i, persisters[i], key, object, lockModes[i], hydratedObjects, session);
-			}
-			else {
-				object = instanceNotYetLoaded( rs, i, persisters[i], entityAliases[i].getRowIdAlias(), key, lockModes[i],
-						optionalObjectKey, optionalObject, hydratedObjects, session );
-			}
-		}
-		return CompletionStages.completedFuture( object );
-	}
```
gavinking, line 289: @DavideD none of the above code was ever called.

Do you know the story behind it? Can we just delete it?

gavinking:
OK, this is ready to go!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveOuterJoinLoader.java
DavideD: What do you mean it's not used? It should be called by the class. If it's not needed, it's probably because there's a test missing. It's the equivalent of `Loader#getRow`.

DavideD: It was the result of converting the `OuterJoinLoader`, so it's possible that I've not added a test for that particular method. Not sure when we expect it to be called.

gavinking: No, none of that code is ever executed. I believe it's code that in Hibernate ORM is related to processing scrollable result sets. (But I'm not sure.)

gavinking: (I mean, it's never executed, because there is nothing in the codebase that references it, not because of missing tests.)

DavideD: Ah, sorry. I just noticed that it's not getObject that doesn't get called but collectionPersister.  I think it was part of that big PR regarding CollectionPersister that I've changed and when I've cleaned it up I didn't realized that these methods were never called.

gavinking: Yes, that's what I figured.

DavideD: Yes, they were called by `rxProcessResultSet` that was an adaptation of `processResultSet`. Sorry about that.

PR #165, opened 2020-05-21
Review the project's dependency metadata

### LINKED ISSUE #126: dependencies
So in the Gradle build we have:

```gradle
    compile ("org.hibernate:hibernate-core:${hibernateOrmVersion}") {
        exclude group: 'org.javassist', module: 'javassist'
        exclude group: 'net.bytebuddy', module: 'byte-buddy'
        exclude group: 'org.dom4j', module: 'dom4j'
        exclude group: 'javax.activation', module: 'javax.activation-api'
        exclude group: 'org.glassfish.jaxb', module: 'jaxb-runtime'
        exclude group: 'javax.xml.bind', module: 'jaxb-api'
        exclude group: 'org.hibernate.common', module: 'hibernate-commons-annotations'
        exclude group: 'org.jboss.spec.javax.transaction', module: 'jboss-transaction-api_1.2_spec'
    }
    testCompile ("org.hibernate:hibernate-testing:${hibernateOrmVersion}")
    compile 'io.vertx:vertx-pg-client:3.9.0'
    compile 'io.vertx:vertx-mysql-client:3.9.0'

    runtimeOnly 'org.jboss.spec.javax.transaction:jboss-transaction-api_1.2_spec:1.1.1.Final'
```

I would like to know what is the reason for all the exclusions here. I would naïvely have expected us to just declare a dependency to `hibernate-core`, and then programs that use HR would not need to explicitly declare their own dependency to Hibernate. But with all these exclusions, they need to redeclare the dependency in order to pick up stuff like `net.bytebuddy`.

I suspect that these exclusions might not now be the right thing, since I've eliminated the separate `-api` module.

Also I'm confused by the exclusion of `jboss-transaction-api_1.2_spec` and then the re-inclusion as a runtime dependency.

@DavideD could you throw some light on this, I'm sure there was some reason.

Sanne:
Fixes #126

I've copied the example into a standalone Maven project, so to verify it and see how it would look like:
 - https://github.com/Sanne/reactive-maven-demo/blob/master/pom.xml#L11-L24

To see the pom.xml file generated for this project, run:

    ./gradlew generatePomFileForMavenJavaPublication

You can them open it from `hibernate-reactive-core/build/publications/mavenJava/pom-default.xml`

This also generates Gradle specific metadata, which is more powerful as it can express more details, but it's only useful for user applications built via Gradle.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -1,8 +1,7 @@
 description = 'Hibernate Reactive Example'
 
 dependencies {
-    compile project(':hibernate-reactive-core')
-    compile 'org.hibernate:hibernate-core'
+    implementation project(':hibernate-reactive-core')
     runtimeOnly 'io.vertx:vertx-pg-client:3.9.1'
     runtimeOnly 'io.vertx:vertx-mysql-client:3.9.1'
 }
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -4,27 +4,31 @@ description = 'Hibernate Reactive Core'
 
 dependencies {
 
-    compile ("org.hibernate:hibernate-core:${hibernateOrmVersion}") {
+    api ("org.hibernate:hibernate-core:${hibernateOrmVersion}") {
+        //We don't need any of these:
         exclude group: 'org.javassist', module: 'javassist'
-        exclude group: 'net.bytebuddy', module: 'byte-buddy'
-        exclude group: 'org.dom4j', module: 'dom4j'
         exclude group: 'javax.activation', module: 'javax.activation-api'
-        exclude group: 'org.glassfish.jaxb', module: 'jaxb-runtime'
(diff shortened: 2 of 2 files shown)


example/build.gradle
```diff
@@ -1,8 +1,8 @@
 description = 'Hibernate Reactive Example'
 
 dependencies {
-    compile project(':hibernate-reactive-core')
-    compile 'org.hibernate:hibernate-core'
+    implementation project(':hibernate-reactive-core')
+    implementation 'org.hibernate:hibernate-core'
```
gavinking, line 5: @Sanne can't we just remove this dependency altogether?

gavinking:
Thanks, looks much better.

example/build.gradle
Sanne: ah yes, let me fix that..   I didn't focus on the example other than removing the deprecation warnings I was getting :)

gavinking:
Great, thanks!

Issue #167, opened 2020-05-22
custom id generators


gavinking:
We need to turn `ReactiveIdentifierGenerator` into an extension point so that users can write their own id generators.

PR #168, opened 2020-05-22
first step towards custom id generators

gavinking:
With a view to eventually solving #167, I've pulled `ReactiveIdentifierGenerator` out into its own package, and massaged the signature of `generate()` a bit to make it look more like the equivalent interface in Hibernate ORM.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/impl/ReactiveSessionInternal.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/impl/ReactiveSessionInternal.java
@@ -5,6 +5,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletionStage;
 
+import org.hibernate.Incubating;
 import org.hibernate.LockMode;
 import org.hibernate.engine.spi.QueryParameters;
 import org.hibernate.event.internal.MergeContext;
@@ -23,6 +24,7 @@ import org.hibernate.reactive.stage.Stage;
  *
  *  @see Stage.Session the actual user visible API
  */
+@Incubating
 public interface ReactiveSessionInternal extends org.hibernate.Session {
 
 	ReactiveActionQueue getReactiveActionQueue();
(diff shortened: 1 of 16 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/impl/ReactiveSessionInternal.java
```diff
@@ -23,6 +24,7 @@
  *
  *  @see Stage.Session the actual user visible API
  */
+@Incubating
 public interface ReactiveSessionInternal extends org.hibernate.Session {
```
DavideD, line 28: Just a question, do we need to add the annotation here considering that this is under the package `impl`?

gavinking: Well the thing is it's exposed as a parameter in `ReactiveIdentifierGenerator`. And since it's kinda the central access point that ties everything else together, I think it eventually needs to move to a public package. But I'm not yes sure what to call that package.

gavinking: `org.hibernate.reactive.session` and `org.hibernate.reactive.session.imp`, perhaps?

gavinking: Actually I think that's a good idea.

DavideD: `org.hibernate.reactive.session` sounds good to me. What about `org.hibernate.reactive`?

gavinking: Well the problem with putting it in the root package is that it would then be the most visible API to users: apparently more important than `Stage` or `Mutiny`, which isn't correct.

PR #169, opened 2020-05-22
Fix podman scripts

DavideD:
* DB2 server start without sudo
 * Remove --memory-swappiness=0 from Postgres (requires root)

@Sanne I noticed that DB2  actually works without sudo on my machine but it ends on the console without a clear message and looks like it's not started. Could you try these changes and let me know?
Postgres wasn't working with the --memory-swappiness=0 option.
@gbadner Could you test these as well when you have a minute? Thanks

### DIFF
--- a/podman.md
+++ b/podman.md
@@ -16,7 +16,7 @@ These are the credentials used to connect to the dbs:
 
 * Server
 ```
-podman run -it --rm=true --memory-swappiness=0 --name HibernateTestingPGSQL -e POSTGRES_USER=hreact -e POSTGRES_PASSWORD=hreact -e POSTGRES_DB=hreact -p 5432:5432 postgres:12
+podman run -it --rm=true --name HibernateTestingPGSQL -e POSTGRES_USER=hreact -e POSTGRES_PASSWORD=hreact -e POSTGRES_DB=hreact -p 5432:5432 postgres:12
 ```
 
 * CLI
@@ -44,10 +44,10 @@ podman exec -it HibernateTestingMariaDB mysql -U hreact -phreact
 
 * Server
 ```
-sudo podman run --rm -it -e LICENSE=accept --privileged=true --name HibernateTestingDB2 -e DBNAME=hreact -e DB2INSTANCE=hreact -e DB2INST1_PASSWORD=hreact -e PERSISTENT_HOME=false -e ARCHIVE_LOGS=false -e AUTOCONFIG=false -p 50000:50000 ibmcom/db2:11.5.0.0a
+podman run --rm -it -e LICENSE=accept --privileged=true --name HibernateTestingDB2 -e DBNAME=hreact -e DB2INSTANCE=hreact -e DB2INST1_PASSWORD=hreact -e PERSISTENT_HOME=false -e ARCHIVE_LOGS=false -e AUTOCONFIG=false -p 50000:50000 ibmcom/db2:11.5.0.0a
 ```
 
 * CLI
 ```
-sudo podman exec -ti HibernateTestingDB2 bash -c "su - hreact -c db2 connect"
+podman exec -ti HibernateTestingDB2 bash -c "su - hreact -c db2 connect"
 ```


podman.md
```diff
@@ -16,7 +16,7 @@ These are the credentials used to connect to the dbs:
 
 * Server
 ```
-podman run -it --rm=true --memory-swappiness=0 --name HibernateTestingPGSQL -e POSTGRES_USER=hreact -e POSTGRES_PASSWORD=hreact -e POSTGRES_DB=hreact -p 5432:5432 postgres:12
```
Sanne, line 19: why deleting `memory-swappiness` ?

podman.md
```diff
@@ -44,10 +44,12 @@ podman exec -it HibernateTestingMariaDB mysql -U hreact -phreact
...
-sudo podman run --rm -it -e LICENSE=accept --privileged=true --name HibernateTestingDB2 -e DBNAME=hreact -e DB2INSTANCE=hreact -e DB2INST1_PASSWORD=hreact -e PERSISTENT_HOME=false -e ARCHIVE_LOGS=false -e AUTOCONFIG=false -p 50000:50000 ibmcom/db2:11.5.0.0a
+podman run --rm -it -e LICENSE=accept --privileged=true --name HibernateTestingDB2 -e DBNAME=hreact -e DB2INSTANCE=hreact -e DB2INST1_PASSWORD=hreact -e PERSISTENT_HOME=false -e ARCHIVE_LOGS=false -e AUTOCONFIG=false -p 50000:50000 ibmcom/db2:11.5.0.0a
 ```
 
 * CLI
 ```
-sudo podman exec -ti HibernateTestingDB2 bash -c "su - hreact -c db2 connect"
+podman exec -ti HibernateTestingDB2 bash -c "su - hreact -c"
+db2
+connect to hreact
```
Sanne, line 54: why 3 steps?

I had already patched the command to connect directly - in a single step

podman.md
DavideD: rootless Postgres wasn't working with the --memory-swappiness=0 option

podman.md
DavideD: Sorry, I will revert it

podman.md
Sanne: it works fine for me?

podman.md
```diff
@@ -44,10 +44,10 @@ podman exec -it HibernateTestingMariaDB mysql -U hreact -phreact
 
 * Server
 ```
-sudo podman run --rm -it -e LICENSE=accept --privileged=true --name HibernateTestingDB2 -e DBNAME=hreact -e DB2INSTANCE=hreact -e DB2INST1_PASSWORD=hreact -e PERSISTENT_HOME=false -e ARCHIVE_LOGS=false -e AUTOCONFIG=false -p 50000:50000 ibmcom/db2:11.5.0.0a
```
Sanne, line 47: I don't know why, but when I remove `sudo` the DB2 container crashes on my machine. I'd leave it be and move on

podman.md
DavideD: OK, but I tested it with Andrea and it doesn't work for him either. Let's hear from @gbadner

podman.md
DavideD: It's a bit annoying to have some machine starting with sudo and some without because they won't appear if you run `podman ps` or `sudo podman ps`. I would prefer if our instructions were consistent.

aguibert: I think `sudo` is required depending on what level of permissions you normally have without `sudo`. The DB2 container requires `--privileged` so that's probably why `sudo` is also needed.

aguibert: Seems to always work fine (without `sudo`) using `docker` fwiw

Sanne: right, it works fine for @DavideD as well. I don't know why but on my specific machine this container seems very problematic - it crashes often with either option, but I've only been able to boot it successfully with `sudo`

podman.md
Sanne: Let's merge it - but I'd still be curious to know of you all happen to test it further. No need to change back though: if it's problematic it's problematic.

PR #173, opened 2020-05-22
move 'impl' stuff to a 'session' package

gavinking:
This nails down `ReactiveSession` and `ReactiveQuery` as a well-defined internal APIs.


hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveSession.java
```diff
@@ -12,20 +12,25 @@
...
+ * allowing it to be passed around through code in Hibernate core, its
+ * non-reactive operations are expected to throw
+ * {@link UnsupportedOperationException}.
+ *
+ *  @see org.hibernate.reactive.stage.Stage.Session
+ *  @see org.hibernate.reactive.mutiny.Mutiny.Session
  */
 @Incubating
-public interface ReactiveSessionInternal extends org.hibernate.Session {
+public interface ReactiveSession extends org.hibernate.Session {
```
DavideD, line 33: Nice

PR #174, opened 2020-05-22
introduce static version of fetch()

### LINKED ISSUE #171: static Reactive.fetch()
I think we should add a static version of `session.fetch()`.

gavinking:
Fixes #171.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -5,6 +5,11 @@ import org.hibernate.CacheMode;
 import org.hibernate.Filter;
 import org.hibernate.FlushMode;
 import org.hibernate.LockMode;
+import org.hibernate.collection.internal.AbstractPersistentCollection;
+import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.reactive.session.ReactiveSession;
 
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -626,4 +632,37 @@
...
+	 * {@code Mutiny.fetch(author.getBook()).thenAccept(book -> print(book.getTitle()));}
+	 * </pre>
+	 *
+	 * @param association a lazy-loaded association
+	 *
+	 * @return the fetched association, via a {@code CompletionStage}
+	 *
+	 * @see org.hibernate.Hibernate#initialize(Object)
+	 */
+	static <T> Uni<T> fetch(T association) {
```
Sanne, line 649: What if the proxy was disconnected from the Session ?

Sanne:
Looks great, but I'd suggest add a good error message for the case of detached proxies.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
gavinking: Well it's actually not totally kosher to reassociate a proxy from one session with a different session Sanne. (JPA semantics vs legacy Hibernate semantics.)

gavinking:
Yeah, OK, NPE :-/

gavinking:
Pushed manually after adding error for closed session.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
Sanne: Sure - I'd never suggest anyone to do it :)  Just pointing out that it could happen.

PR #175, opened 2020-05-22
Enable mavenLocal() repository

DavideD:
Maybe I'm not familiar enough with gradle, but why would you disable the local maven repository?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,7 @@ subprojects {
     compileJava.options.encoding = 'UTF-8'
 
     repositories {
-        // Only enable these for local development, never push it:
-        // mavenLocal()
+        mavenLocal()
         // jcenter()
         mavenCentral()
     }


DavideD:
I guess it makes sense on CI to make sure that we only have publish artifacts, but I'm sure we can find a way to disable it when necessary on CI. I would prefer to have it enable for now since we develop a lot with ORM SNAPSHOTS.

aguibert:
I'm not sure why `mavenLocal()` would be bad, but @Sanne was the one who just added it recently, so probably better to get his opinion. If he went out of his way to do it, it was probably for a reason though.

DavideD:
Ah sorry, I thought it was part of your last PR. I should have checked better.

Sanne:
I'd highly recommend to never use it. The metadata that Gradle offers is much more precise than what the maven structure can offer, and it has no notion of snapshots (which is a good thing - as builds need to be reproducible even during development)

We've been burned by this many times in Hibernate ORM and it's really great what you can do with Gradle without having the state "tainted" by the maven repo.

BTW not having `mavenLocal()` is just about _consuming_ other dependencies; you can still deploy to the maven local repository to test things from a different Maven project.

Sanne:
I'll close, trust me ... I can share some war stories about this another time.

DavideD:
It's ok, thanks. I will add a more explicit comment, though

Issue #46, opened 2020-01-20
native SQL queries


gavinking:
We need these.

gavinking:
Uff, sorry @DavideD, I did not see that you had assigned this issue to yourself .... I hope you weren't actively working on it!

DavideD:
It's OK, just keep in mind that Gail is working on the support for parameters.

gavinking:
> just keep in mind that Gail is working on the support for parameters.

Yes, I was deliberately not touching that.

PR #180, opened 2020-05-23
support for native SQL queries, named queries, and criteria queries

gavinking:
Initial implementation of #46.

Still need to add some tests.


hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -250,7 +266,123 @@ public String getCursorName() {
...
+
+			@Override
+			public String getCatalogName(int column) {
+				return null;
+			}
+
+			@Override
+			public int getColumnType(int column) {
+				//TODO: we need SQLCLient to provide some column type info
+				return Types.VARCHAR;
```
gavinking, line 348: We need the `RowSet` to provide us with column type info if we want this to work really nicely without having to write a `@SqlResultSetMapping`.

gavinking:
Also added named query support.

gavinking:
.... aaaand now criteria queries!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveCriteriaQueryImpl.java
```diff
@@ -0,0 +1,145 @@
...
+	private static Field queryStructureField;
+	static {
+		try {
+			queryStructureField = CriteriaQueryImpl.class.getDeclaredField("queryStructure");
+			queryStructureField.setAccessible(true);
+		}
+		catch (Exception e) {
+			throw new AssertionFailure("missing field", e);
+		}
+	}
```
gavinking, line 42: We can't use reflection in Quarkus so we definitely need to expose this field in ORM.

gavinking:
Still needs tests!!

gavinking:
OK, tests done, this is ready to go!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveCriteriaQueryImpl.java
gavinking: See https://github.com/hibernate/hibernate-orm/pull/3419

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: See https://github.com/eclipse-vertx/vertx-sql-client/issues/657

gavinking:
Gotta merge this because other work depends on it.

PR #191, opened 2020-05-25
Introduce the spotless build plugin and file headers

### LINKED ISSUE #60: Add licenses
We need to add a license to all classes similar to the one we have for the other hibernate projects

Sanne:
What would you think of this approach @gavinking , @DavideD , @gbadner , @aguibert ?

Rather than getting in the way and fail the build on a missing license, it will automatically fix such issues.

Fixes #60 

Please try it out and see what happens when you fetch this and do a regular `./gradlew clean build` ...


gavinking:
That sounds reasonable. Lemme try it out.

hibernate-reactive-core/build.gradle
```diff
@@ -48,3 +48,18 @@ test {
...
   systemProperty 'db', project.hasProperty('db') ? project.getProperty('db') : 'postgresql'
 }
+
+spotless {
+    //Don't fail during the check: rather than enforcing guidelines, we use this plugin to fix mistakes automatically.
+    enforceCheck false
+    java {
+        licenseHeaderFile 'spotless.license.java'
+        //These seem rather cool but let's introduce them later iff there's consensus - and probably at a better time.
+//        removeUnusedImports()
```
gavinking, line 58: Oh if this works it would be magical.

gavinking:
So I'm a little confused. It seems really very unobtrusive ... except ... it also didn't seem to actually *do* anything. It didn't add any headers for me AFAIKS.

gavinking:
Oh, OK, `gradle spotlessApply`.

And it's fast as hell.

LGTM.

hibernate-reactive-core/build.gradle
```diff
@@ -48,3 +48,18 @@ test {
...
+//        trimTrailingWhitespace()
+//        indentWithTabs()
```
gavinking, line 60: Don't enable this one. It does some sort of weird violence to my Javadoc comments.

hibernate-reactive-core/spotless.license.java
```diff
@@ -0,0 +1,6 @@
+/**
```
gavinking, line 1: @Sanne please don't use a Javadoc comment here.

hibernate-reactive-core/build.gradle
gavinking: Yep, this one is awesome.

gavinking:
@Sanne I proposed to small changes. Other than that this looks good.

hibernate-reactive-core/build.gradle
Sanne: right I noticed the same. I guess we can explore these options later: I also wouldn't want it to reformat the whole project right now, it would cause a load of conflicts with all current work in progress.

Sanne:
> So I'm a little confused. It seems really very unobtrusive ... except ... it also didn't seem to actually _do_ anything. It didn't add any headers for me AFAIKS.

hum it didn't? I had set it up so that the task `build` would imply `spotlessApply` automatically.

gavinking:
> hum it didn't? I had set it up so that the task `build` would imply `spotlessApply` automatically.

I did try editing a file to see if that would trigger it.

Sanne:
@gavinking I've applied your feedback and improved the hook a bit: could you try it again?

I suspect it didn't work for you before as it wasn't triggered if there were any failing tests. Now it's run before compiling, probably preferreable

Sanne:
Added a second commit which actually applies all the change the sources.

DavideD:
The second commit cannot be applied anymore. @Sanne Could you send a new one? Or if it's just a result of the `gradle build` I can do it. Whatever you prefer. Then  I think we should merge it and not wait any longer.

Sanne:
ok since I have two approvals I'll repeat the formatting step and push it all.

Issue #96, opened 2020-04-25
Blobs/Clobs


gavinking:
We need to discuss Blobs and Clobs. Apparently the vert.x driver has some level of support for them (mapping them to `Buffer`), though I don't know any details, and it's not discussed at all in the documentation.

Now, Blob/Clob support has always been a bit uncomfortable in regular Hibernate, but it occurs to me that perhaps it might be a little more natural in this model.

I wonder who knows more details about how this works in the Vert.x driver.

emmanuelbernard:
@vietj is the right person.

gavinking:
> Now, Blob/Clob support has always been a bit uncomfortable in regular Hibernate, but it occurs to me that perhaps it might be a little more natural in this model.

Apparently not, AFAICS.

PR #192, opened 2020-05-25
testing @Lob

gavinking:
So here's a little bit of an experiment with `BLOB`/`CLOB` support.

I had to mess with the Hibernate `Dialect`s, because by default they use `Stream`s or actual `java.sql.Blob` \ `java.sql.Clob` objects to bind LOB parameters.

So I'm passing:

- a Vert.x `Buffer` for the `BLOB`, and
- a `String` for the `CLOB`.

This works on Postgres and MySQL, but not quite on DB2, since the DB2 client doesn't support binding `Buffer`s (cc @aguibert).

But @vietj is this the right thing to do? Is there some better way more specialized way we should be reading and writing LOBs?

See #96.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -62,8 +62,7 @@ public class ResultSetAdaptor implements ResultSet {
 	}
 
 	@Override
-	public void close() {
-	}
+	public void close() {}
 
 	@Override
 	public boolean wasNull() {
@@ -240,18 +239,33 @@ public class ResultSetAdaptor implements ResultSet {
 	}
 
 	@Override
-	public InputStream getAsciiStream(String columnLabel) {
-		throw new NotYetImplementedException( "This type hasn't been implemented yet" );
+	public int getHoldability() {
+		return CLOSE_CURSORS_AT_COMMIT;
 	}
 
 	@Override
-	public InputStream getUnicodeStream(String columnLabel) {
-		throw new NotYetImplementedException( "This type hasn't been implemented yet" );
+	public boolean isClosed() {
+		return false;
 	}
 
 	@Override
-	public InputStream getBinaryStream(String columnLabel) {
-		throw new NotYetImplementedException( "This type hasn't been implemented yet" );
+	public <T> T getObject(int columnIndex, Class<T> type) {
+		return row.get(type, columnIndex);
+	}
+
+	@Override
+	public <T> T getObject(String columnLabel, Class<T> type) {
+		return row.get( type, row.getColumnIndex(columnLabel) );
+	}
+
+	@Override
+	public <T> T unwrap(Class<T> iface) {
+		return null;
+	}
+
+	@Override
+	public boolean isWrapperFor(Class<?> iface) {
+		return false;
 	}
 
 	@Override
@@ -260,9 +274,7 @@ public class ResultSetAdaptor implements ResultSet {
 	}
 
 	@Override
-	public void clearWarnings() {
-
-	}
+	public void clearWarnings() {}
 
 	@Override
 	public String getCursorName() {
@@ -407,16 +419,6 @@ public class ResultSetAdaptor implements ResultSet {
 	}
 
 	@Override
-	public Reader getCharacterStream(int columnIndex) {
-		throw new NotYetImplementedException( "This type hasn't been implemented yet" );
-	}
-
-	@Override
-	public Reader getCharacterStream(String columnLabel) {
-		throw new NotYetImplementedException( "This type hasn't been implemented yet" );
-	}
-
-	@Override
 	public BigDecimal getBigDecimal(int columnIndex) {
 		BigDecimal decimal = row.getBigDecimal(columnIndex);
 		return (wasNull=decimal==null) ? null : decimal;
@@ -429,79 +431,15 @@ public class ResultSetAdaptor implements ResultSet {
 	}
 
 	@Override
-	public boolean isBeforeFirst() {
-		return false;
-	}
-
-	@Override
-	public boolean isAfterLast() {
-		return false;
-	}
-
-	@Override
-	public boolean isFirst() {
-		return false;
-	}
-
-	@Override
-	public boolean isLast() {
-		return false;
-	}
-
-	@Override
-	public void beforeFirst() {
-
-	}
-
-	@Override
-	public void afterLast() {
-
-	}
-
-	@Override
-	public boolean first() {
-		return false;
-	}
-
-	@Override
-	public boolean last() {
-		return false;
-	}
-
-	@Override
-	public int getRow() {
-		return 0;
-	}
-
-	@Override
-	public boolean absolute(int row) {
-		return false;
-	}
-
-	@Override
-	public boolean relative(int rows) {
-		return false;
-	}
-
-	@Override
-	public boolean previous() {
-		return false;
-	}
-
-	@Override
-	public void setFetchDirection(int direction) {
-
-	}
+	public void setFetchDirection(int direction) {}
 
 	@Override
 	public int getFetchDirection() {
-		return 0;
+		return FETCH_FORWARD;
 	}
 
 	@Override
-	public void setFetchSize(int rows) {
-
-	}
+	public void setFetchSize(int rows) {}
 
 	@Override
 	public int getFetchSize() {
@@ -510,7 +448,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public int getType() {
-		return 0;
+		return TYPE_FORWARD_ONLY;
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/boot/dialect/ReactivePostgresDialect.java
```diff
@@ -0,0 +1,31 @@
...
+	public SqlTypeDescriptor getSqlTypeDescriptorOverride(int sqlCode) {
+		switch ( sqlCode ) {
+			case Types.BLOB:
+				return VarbinaryTypeDescriptor.INSTANCE;
+			case Types.CLOB:
+				return VarcharTypeDescriptor.INSTANCE;
+			default:
+				return super.getSqlTypeDescriptorOverride( sqlCode );
+		}
+	}
```
gavinking, line 30: @Sanne can you think of any other way to override the type mappings for `BLOB` and `CLOB` that doesn't involve subclassing the `Dialect`s?

(Bearing in mind that the dialects already override these mappings.)

I could not find, and can't recall, any way to do this globally.

gavinking:
So to get this test to pass I had to subclass all the `Dialect`s. I don't like doing that *at all*, and I was hoping I would find another way to do it. But I have not found anything. Perhaps @Sanne can think of something.

gavinking:
> But @vietj is this the right thing to do? Is there some better way more specialized way we should be reading and writing LOBs?

After reviewing the Vert.x docs, it seems to me that this really is the only way to do it.

gavinking:
> the DB2 client doesn't support binding Buffers (cc @aguibert).

That's this issue: https://github.com/eclipse-vertx/vertx-sql-client/issues/496.

Sanne:
I'm not sure as I'm not familiar with this area - perhaps a `org.hibernate.boot.model.TypeContributor` could help?  cc/ @sebersole 

Also: ifyou expect the vertx-sql drivers to implement this some time soon, how would we deal with that?
Suddently starting to use it would be an annoyingly breaking change to the schema - so perhaps it would be better now to just make it clear that these types aren't supported, and fix it later?

gavinking:
> Also: ifyou expect the vertx-sql drivers to implement this some time soon, how would we deal with that?

WDYM?

The reason I'm overriding the `Dialect`s is that Hibernate by default will pass my a `java.sql.Blob` which I very much doubt that Vert.x is going to ever support...

Sanne:
> The reason I'm overriding the `Dialect`s is that Hibernate by default will pass my a `java.sql.Blob` which I very much doubt that Vert.x is going to ever support...

OK, sad but if that's that the case then I agree.

We'll need to consider that this makes it impossible to have a traditional, blocking instance of Hibernate ORM share the same schema as the Reactive ORM - even if the user is trying to reuse exactly the same model.

Which might be an acceptable limitation, but it implies we can't reuse e.g. the standard ORM for schema updates, etc..

gavinking:
> We'll need to consider that this makes it impossible to have a traditional, blocking instance of Hibernate ORM share the same schema as the Reactive ORM

Well the only *schema* change I made was to override `OID` in Postgres. (Since the Vert.x driver doesn't support that.) But look at the comment I made here:

<https://github.com/hibernate/hibernate-orm/blob/cc8aaab9add0a8a9f9078f908b1d69777f567df4/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLDialect.java#L99>

The *only* reason we use `OID` in Hibernate core is that the Postgres JDBC driver is broken.

The reason I'm overriding the dialects isn't to change the schema export, it's to change how parameters are bound.

gavinking:
> perhaps a org.hibernate.boot.model.TypeContributor could help? cc/ @sebersole

So thanks for the suggestion, I just tried that. And it didn't work for two reasons, it seems:

- I believe that `Dialect.getSqlTypeDescriptorOverride()` type overrides take precedence.
- I couldn't even figure out how to actually override the type mapping for `@Lob`s. Contributing new types didn't seem to help.

Anyway it doesn't look like that's going to work, sadly.

sebersole:
`#getSqlTypeDescriptorOverride` happens as part of the hacked pre-6 approach to allow Dialect to influence the `SqlTypeDescriptor` to use -> `#remapSqlTypeDescriptor`.  

In `#remapSqlTypeDescriptor`, if the `SqlTypeDescriptor` being checked says it cannot be re-mapped, it won't and control will never pass to `#getSqlTypeDescriptorOverride`.  So probably you need new `SqlTypeDescriptor` impls for these cases which specify they cannot be remapped.

This "type combining" process is simpler in 6, although I'm not sure this exact case is handled.  As long as there are specific tests (hopefully ones that specifically assert the types used) over the case we'll see how it manifests in 6

gavinking:
> So probably you need new `SqlTypeDescriptor` impls for these cases which specify they cannot be remapped.

Ah, great, perfect, I missed that. Thanks.

gavinking:
@sebersole Hrm, that's still not enough. It's still using `BlobTypeDescriptor`, instead of my `SqlTypeDescriptor` I contributed for `Types.BLOB`.

I feel like there might be something special that maps `@Lob` to `BlobTypeDescriptor`. Is that right?

gavinking:
Oh, this doesn't help.

It looks like in `JavaTypeDescriptorRegistry` you can't actually override `JavaTypeDescriptor`s.

```java
	public void addDescriptor(JavaTypeDescriptor descriptor) {
		JavaTypeDescriptor old = descriptorsByClass.put( descriptor.getJavaType(), descriptor );
		if ( old != null ) {
```

gavinking:
Excuse me, just deleted a retarded comment.

sebersole:
> I feel like there might be something special that maps `@Lob` to `BlobTypeDescriptor`.

Ah... for `@Lob` yes... there is lots of nastiness in annotation binding that does stuff off-script.  That's unfortunately just the way that annotation binding grew up.

So that particular thorn is in `org.hibernate.cfg.annotations.SimpleValueBinder#setType` which proceeds to use lots of hard-coded things including which specific type resolutions to use

gavinking:
> Ah... for `@Lob` yes... there is lots of nastiness in annotation binding that does stuff off-script. That's unfortunately just the way that annotation binding grew up.

OK, that's what I guessed.

So for now I figure I'll have to override the `Dialect`s.

But hopefully we can figure out some sort of patch to Hibernate ORM that will let me perform `SqlTypeDescriptor` remapping in some other way.

sebersole:
I can see wanting that.  But, personally I'm not going to spend time on it.  I finally got these things working the way I think they should be in 6 and I just do not see an easy way to do this in previous versions.

The problem is fundamentally the tight binding between `JavaTypeDescriptor` and `SqlTypeDescriptor` in the legacy Type system, especially in terms of resolution.  Versions prior to 6 want to perform this resolution in-total to the point of determining a Type name.  Which is obviously a lot different than resolving each individually and resolving a composition for the 2

sebersole:
Actually... thinking about this some more, affecting this at the `JavaTypeDescriptor` / `SqlTypeDescriptor` level is not going to work in v5.  What you are going to have to do here is to adjust the BasicTypeRegistry (for completeness I would still adjust SqlTypeDescriptorRegistry as well).  Override the entries for all standard related LOB mappings.  So you'll need:

- Custom `SqlTypeDescriptor` impls
- Custom `BasicType` impls using the custom `SqlTypeDescriptor` impls
- Register your custom `SqlTypeDescriptor` and `BasicType` impls

The code in `SimpleValueBinder` for LOB handling basically relies on the ability to specify the name of a BasicType to use relatively to the registry.  So your overrides would take precedent.

gavinking:
@sebersole OK, perfect, *that* was what I was looking for. FTR, here is the solution:

```java
public class NoJdbcTypeContributor implements TypeContributor {
	@Override
	public void contribute(TypeContributions typeContributions, ServiceRegistry serviceRegistry) {
		typeContributions.getTypeConfiguration().getBasicTypeRegistry()
				.register(new AbstractSingleColumnStandardBasicType<byte[]>(
						VarbinaryTypeDescriptor.INSTANCE,
						PrimitiveByteArrayTypeDescriptor.INSTANCE
				) {
			@Override
			public String getName() {
				return "materialized_blob";
			}
			@Override
			protected boolean registerUnderJavaType() {
				return true;
			}
		});
		typeContributions.getTypeConfiguration().getBasicTypeRegistry()
				.register(new AbstractSingleColumnStandardBasicType<String>(
						VarcharTypeDescriptor.INSTANCE,
						StringTypeDescriptor.INSTANCE
				) {
					@Override
					public String getName() {
						return "materialized_clob";
					}
					@Override
					protected boolean registerUnderJavaType() {
						return true;
					}
				});
	}
}
```

gavinking:
Alright, done, perfect, thanks everyone!

aguibert:
@gavinking I'm looking at this for DB2 now and there are a couple layers of issues that I'm finding.

First, the DB2 client just doesn't support blob/clob yet -- working on that for the next Vertx release.

Second, the max size that DB2 supports for `varchar` columns is 32k, and your test has length = 100,000. IIUC normally hibernate would use a `blob/clob` column type here, but due to these overrides it's still using `varchar`. To me it seems like the proper thing to do would to still use a column type of `blob/clob` but just map that to the necessary byte buffer types in Vertx.

gavinking:
> but due to these overrides it's still using `varchar`.

Agreed.


Alright, so what I'd really like to do would be this:

```java
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<byte[]>(
		new VarbinaryTypeDescriptor() {
			@Override
			public int getSqlType() {
				return Types.BLOB;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		PrimitiveByteArrayTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_blob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<String>(
		new VarcharTypeDescriptor() {
			@Override
			public int getSqlType() {
				return Types.CLOB;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		StringTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_clob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
```

But that doesn't work for me because on Postgres it results in the type `OID`, which doesn't seem to be properly supported by the Vertx client.

Alternatively, we could try this:

```java
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<byte[]>(
		new VarbinaryTypeDescriptor() {
			@Override
			public int getSqlType() {
				return Types.LONGVARBINARY;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		PrimitiveByteArrayTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_blob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<String>(
		new VarcharTypeDescriptor() {
			@Override
			public int getSqlType() {
				return Types.LONGVARCHAR;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		StringTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_clob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
```
which seems to do pretty much the right thing on Postgres and MySQL, but on DB2 it gives us `long varchar for bit data`, which is not precisely a blob.

So it's a bit of a pain. Of course I can fix all this be subclassing the `Dialect`s, but that's the one thing I *don't* want to customize.

Perhaps we could have the type descriptors we register be sensitive the runtime dialect?

WDYT?

gavinking:
Well, here ya go @aguibert, this seems to work:

```java
BasicTypeRegistry basicTypeRegistry =
		typeContributions.getTypeConfiguration().getBasicTypeRegistry();
final Dialect dialect = serviceRegistry.getService(JdbcEnvironment.class).getDialect();
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<byte[]>(
		new VarbinaryTypeDescriptor() {
			@Override
			public int getSqlType() {
				return dialect instanceof PostgreSQL10Dialect
						? Types.LONGVARBINARY
						: Types.BLOB;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		PrimitiveByteArrayTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_blob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
basicTypeRegistry.register(new AbstractSingleColumnStandardBasicType<String>(
		new VarcharTypeDescriptor() {
			@Override
			public int getSqlType() {
				return dialect instanceof PostgreSQL10Dialect
						? Types.LONGVARCHAR
						: Types.CLOB;
			}
			@Override
			public boolean canBeRemapped() {
				return false;
			}
		},
		StringTypeDescriptor.INSTANCE
) {
	@Override
	public String getName() {
		return "materialized_clob";
	}
	@Override
	protected boolean registerUnderJavaType() {
		return true;
	}
});
```

Ugly as hell, but.

gavinking:
Well, it's here if you want to try it:

<https://github.com/gavinking/hibernate-reactive/commit/3a1a2e32474020eb1931c32cf2b34f7aed532b20>

aguibert:
@gavinking nice, I was going down a rabbit hole of getting this to work without needing to register any `AbstractSingleColumnStandardBasicType` impls.

Instead, I was going down the path of creating a Blob/ClobAdapter, but really all we need to do is implement the currently unsupported methods in PreparedStatementAdaptor to intercept things like `setBlob(int idx, Blob b)` and convert them to a data type that Vertx supports. For now that' pretty much just `Buffer/ByteBuf` for LOBs. Then when we pull these values out in the ResultSetAdaptor we can do the conversion back the other way

gavinking:
> Instead, I was going down the path of creating a Blob/ClobAdapter, but really all we need to do is implement the currently unsupported methods in PreparedStatementAdaptor to intercept things like `setBlob(int idx, Blob b)` and convert them to a data type that Vertx supports. For now that' pretty much just `Buffer/ByteBuf` for LOBs. Then when we pull these values out in the ResultSetAdaptor we can do the conversion back the other way

Well, yeah, but then we have Hibernate wrapping our byte arrays as `Blob`s, then `PreparedStatementAdaptor` unwrapping the `Blob`s as byte arrays and re-wrapping them as `Buffer`s, and then the opposite useless process in reverse for a `ResultSet`.

And what for? There's no *actual* `Blob`s and `Clob`s anywhere, there's only byte arrays and strings.

aguibert:
Right, the adaptors would not be of any use. The only positive effect that seemed to have was getting the correct column type used (e.g. `blob(100000)` instead of the invalid `varchar(100000) for bit data`)

gavinking:
So what is the plan here, does it help if I push that commit?

aguibert:
Yes, I think that commit would help, but I haven't had time to try it out yet. Been focusing on getting the metadata SPI done in Vertx.

IIUC that commit will essentially keep default behavior (blob/clob column types) for all DBs except for PG, which will special-case to using longvarchar/longvarbinary types? If so, sounds good to me

gavinking:
On Postgres it will use `bytea` and `text`, which is actually what we want.

Issue #142, opened 2020-05-19
add lock() method


gavinking:
Following on from #43, we should add the `Session.lock()` method.

This is not a high priority at all.

gavinking:
So it looks like we'll have *something* in the first release. But Hibernate has a plethora of lock modes and I'm not sure I want to implement all of them right now.

gavinking:
Done!

gavinking:
(Followup work in #201.)

PR #195, opened 2020-05-26
lock() method

gavinking:
I have hacked together some sort of support for the `lock()` method.

The only supported lock modes (for now) are:

- ` PESSIMISTIC_WRITE`
- `PESSIMISTIC_READ`

There are no `LockingStrategy`s, this is hardcoded for now.

see #142


gavinking:
I've added support for `PESSIMISTIC_FORCE_INCREMENT`, and cleaned up some code.

This is ready to merge.

gavinking:
OK, after a review of the semantics, I've decided that this now supports all `LockMode`s except `OPTIMISTIC` and `OPTIMISTIC_FORCE_INCREMENT` (which we *definitely* don't need now).

gavinking:
(Note that it's all implemented in one big method: there's no need for `LockingStrategy`s right now, since we're supporting such a small number of databases.)

PR #196, opened 2020-05-26
add SessionFactory.createSession()

gavinking:
This is a request from @aguibert.

Currently `SessionFactory.openSession()` returns a `CompletionStage`. This would add an additional operation that gives you a `Session` with no `CompletionStage`, fetching the database connection lazily.

Would like some feedback on this issue.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -756,9 +756,25 @@ public interface Mutiny {
 	interface SessionFactory extends AutoCloseable {
 
 		/**
-		 * Obtain a new {@link Session reactive session}, the
-		 * main interaction point between the user's program and
-		 * Hibernate Reactive.
+		 * Obtain a new {@link Session reactive session}, the main
+		 * interaction point between the user's program and Hibernate
+		 * Reactive.
+		 *
+		 * The underlying database connection is obtained lazily
+		 * when the returned {@link Session} needs to access the
+		 * database.
+		 *
+		 * The client must close the session using {@link Session#close()}.
+		 */
+		Session createSession();
+
+		/**
+		 * Obtain a new {@link Session reactive session}, the main
+		 * interaction point between the user's program and Hibernate
+		 * Reactive.
+		 *
+		 * The underlying database connection is obtained before the
+		 * {@link Session} is returned via a {@link Uni}.
 		 *
 		 * The client must close the session using {@link Session#close()}.
 		 */
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnectionPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnectionPool.java
@@ -31,8 +31,16 @@ import java.util.concurrent.CompletionStage;
 public interface ReactiveConnectionPool extends Service {
 
 	/**
-	 * Obtain a reactive connection.
+	 * Obtain a reactive connection, returning the connection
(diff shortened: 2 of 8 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnectionPool.java
```diff
@@ -31,8 +31,16 @@
 public interface ReactiveConnectionPool extends Service {
 
 	/**
-	 * Obtain a reactive connection.
+	 * Obtain a reactive connection, retuning the connection
```
DavideD, line 34: Typo: returning

gavinking: Yeah, but c'mon, it would be cool if it did "retune" things.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -756,9 +756,25 @@
...
+		 */
+		Session createSession();
+
+		/**
+		 * Obtain a new {@link Session reactive session}, the main
+		 * interaction point between the user's program and Hibernate
+		 * Reactive.
+		 *
+		 * The underlying database connection is obtained before the
+		 * {@link Session} is returned via a {@link Uni}.
```
aguibert, line 777: Since `openSession()` also retrieves a connection lazily, I wonder if we still need this method at all? We can keep them both around for now, but by first release we may want to pick one or the other.  Obviously I prefer `createSession()` (because I requested it) and users can pretty easily wrap the `Session` in a `Uni` on their own if they want

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -120,6 +120,14 @@ public void after(TestContext context) {
 		return sessionFactory.unwrap( Stage.SessionFactory.class );
 	}
 
+	protected Stage.Session getSession() {
```
DavideD, line 123: I think we should call this `createSession` or similar

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
gavinking: Well `openSession()` is good for people who might want to report a connection failure in a specific way.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
DavideD: Also it's not used anywhere. do we want' to include it anyway for the future?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -109,6 +110,32 @@ public void reactiveFind(TestContext context) {
 		);
 	}
 
+	@Test
+	public void reactivePersistFindDelete(TestContext context) {
+		final GuineaPig guineaPig = new GuineaPig( 5, "Aloi" );
+		Stage.Session session = getSessionFactory().createSession();
```
DavideD, line 116: I assume you were planning to use the getSession() method here

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
gavinking: Yes, it's there for people who want this convenience when writing new tests.

PR #198, opened 2020-05-27
163 Create gradle task to test all dbs

DavideD:


DavideD:
Something is not right

.github/workflows/build.yml
```diff
@@ -19,37 +19,9 @@ jobs:
         java-version: 1.8
         server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
         settings-path: ${{ github.workspace }} # location for the settings.xml file
-    - name: Build and Test with PostgreSQL
-      run: ./gradlew build -Pdocker -Pdb=pg
+    - name: Test and build all databases
```
aguibert, line 22: @DavideD This is simpler in the pipeline file, but the major drawback is that it will now run everything in sequence . Before, everything ran in parallel (jobs run in parallel if they have no dependencies on other jobs).
I'd really like to see a solution that runs everything in parallel so that CI completes faster, but I don't think that will be possible with an all-in-one gradle task.

aguibert: I see a new PR was opened so I will move this comment over there

PR #200, opened 2020-05-27
integrate Hibernate Validator in example app

gavinking:
WDYT, should we do this?

So, on the one hand, it makes the `build.gradle` more complex.

On the other it shows that Hibernate Reactive still works with other stuff in the Hibernate ecosystem.

@emmanuelbernard?

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -3,14 +3,18 @@ description = 'Hibernate Reactive Example'
 dependencies {
     implementation project(':hibernate-reactive-core')
 
-    // JPA metamodel generation for criteria queries
+    // Hibernate Validator (optional)
+    implementation 'org.hibernate.validator:hibernate-validator:6.1.5.Final'
+    runtimeOnly 'org.glassfish:jakarta.el:3.0.3'
+
+    // JPA metamodel generation for criteria queries (optional)
     annotationProcessor "org.hibernate:hibernate-jpamodelgen:${hibernateOrmVersion}"
 
     // database drivers for PostgreSQL and MySQL
     runtimeOnly 'io.vertx:vertx-pg-client:3.9.1'
     runtimeOnly 'io.vertx:vertx-mysql-client:3.9.1'
 
-    // logging
+    // logging (optional)
     runtimeOnly 'org.slf4j:slf4j-log4j12:1.7.30'
 }
 
--- a/example/src/main/java/org/hibernate/example/reactive/Author.java
+++ b/example/src/main/java/org/hibernate/example/reactive/Author.java
@@ -5,6 +5,8 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -13,10 +15,10 @@ import static javax.persistence.CascadeType.PERSIST;
 @Entity
 @Table(name="authors")
 class Author {
-	@Id
-	@GeneratedValue
+	@Id @GeneratedValue
 	Integer id;
 
+	@NotNull @Size(max=100)
 	String name;
 
 	@OneToMany(mappedBy = "author", cascade = PERSIST)
--- a/example/src/main/java/org/hibernate/example/reactive/Book.java
+++ b/example/src/main/java/org/hibernate/example/reactive/Book.java
@@ -5,19 +5,24 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 import static javax.persistence.FetchType.LAZY;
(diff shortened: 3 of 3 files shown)


gavinking:
P.S. @emmanuelbernard there is a pretty bad behavior in Hibernate Validator when a `HV000183` error occurs because you don't have EL on the classpath.

Instead of logging a message it just throws a `IntegrationException` which is then immediately caught and swallowed by `TypeSafeActivator.activate()` with just the following *debug* level message in the log:

    LOG.debug( "Unable to acquire Bean Validation ValidatorFactory, skipping activation" );

So not having EL in the classpath results in Hibernate Validator completely silently not working.

emmanuelbernard:
I can’t vouch for the build complexity but having and testing such example is really good. It avoids bad surprises late in the integration chain.

emmanuelbernard:
@gsmet you aware of this HV behavior of swallowing the EL missing exception ?

gavinking:
> I can’t vouch for the build complexity

Well, it's two lines. Which isn't perfect, since it should be just one line.

But it's certainly manageable ;-)

gavinking:
> I can’t vouch for the build complexity but having and testing such example is really good. 

OK, I pushed it manually

gsmet:
IIRC, that's not an issue in HV, that's how they wanted it to work in JPA.

I haven't checked but my recollection is that setting `javax.persistence.validation.mode` to `CALLBACK` gives you a proper exception as it expects HV to be there.

If set to AUTO which is the default, if you have an error, it's just not enabled. I personally think it was a mistake but I was told it was supposed to work that way.

Also, I must admit I'm more and more leaning towards making EL a non-optional dependency (that you could exclude if you want and set up the ad hoc config) as it's needed with the default configuration. I never understood why it was optional considering you need it if you keep everything default.

gavinking:
> If set to AUTO which is the default, if you have an error, it's just not enabled. I personally think it was a mistake but I was told it was supposed to work that way.

But in this case the problem wasn't thet *bean validator* was missing, it's that it was there, and *one off its dependencies* was missing.

I'm pretty sure that's got to be a visible error to the user.

> Also, I must admit I'm more and more leaning towards making EL a non-optional dependency

Yes, I think it should be.

gsmet:
On Wed, May 27, 2020 at 6:00 PM Gavin King <notifications@github.com> wrote:

> If set to AUTO which is the default, if you have an error, it's just not
> enabled. I personally think it was a mistake but I was told it was supposed
> to work that way.
>
> But in this case the problem wasn't thet *bean validator* was missing,
> it's that it was there, and *one off its dependencies* was missing.
>

Yes exactly. As soon as you have an error bootstrapping HV, whatever it is,
it gets swallowed and the HV integration is disabled in ORM.

Not sure if it's the JPA spec mandating that or if someone thought it would
be a good idea. It's like that since far before I joined the project. I
remember Gunnar not being too happy about it either.

-- 
Guillaume

gavinking:
Well I have the spec in front of me. It says:

> If the auto validation mode is specified by the `validation-mode` element or the `javax.persis- tence.validation.mode` property, or if neither the `validation-mode` element nor the `javax.persistence.validation.mode property` is specified, and a Bean Validation provider is present in the environment, the persistence provider must perform the automatic validation of entities as described in section 3.6.1.2. If no Bean Validation provider is present in the environment, no lifecycle event validation takes place.

And almost exactly the same text is copy/pasted into the Javadoc of `ValidationMode`.

So it's definitely not the case that this behavior is mandated by the spec.

emmanuelbernard:
> Well I have the spec in front of me. It says:
> 
> > If the auto validation mode is specified by the `validation-mode` element or the `javax.persis- tence.validation.mode` property, or if neither the `validation-mode` element nor the `javax.persistence.validation.mode property` is specified, and a Bean Validation provider is present in the environment, the persistence provider must perform the automatic validation of entities as described in section 3.6.1.2. If no Bean Validation provider is present in the environment, no lifecycle event validation takes place.
> 
> And almost exactly the same text is copy/pasted into the Javadoc of `ValidationMode`.
> 
> So it's definitely not the case that this behavior is mandated by the spec.

Right. It was not my intend to swallow "any issue" when I wrote that bit.

gsmet:
I have a vague recollection of a discussion where the issue was that we
didn't have a proper exception to differentiate between no implementation
and any other error.

I think we probably need to add a specific exception type in the spec so
that we can handle this case differently from other errors.

Again, vague recollection. And I'm on PTO so I will check that another day
:).

emmanuelbernard:
At least we could make it work for HV, which is the only reasonable implementation anyways out there these days.

gavinking:
> I have a vague recollection of a discussion where the issue was that we
didn't have a proper exception to differentiate between no implementation
and any other error.

Well note that all the code I'm seeing is completely internal to Hibernate Validator. I'm not sure if Hibernate *ORM* subsequently *reswallows* exceptions, I didn't get that far.

gsmet:
Sure, that could be an option if people are open to it.

gsmet:
IIRC, if you set the JPA validation mode to `callback`, you will have a proper error if you don't have an EL implementation. So it must be done in ORM.

I'll have a look when I'm back from PTO.

Issue #163, opened 2020-05-21
Create gradle task to test all db profiles


DavideD:

DavideD:
Done https://github.com/hibernate/hibernate-reactive/pull/218

gavinking:
Excellent, finally -)

PR #202, opened 2020-05-27
#163 Gradle testall task

DavideD:
This adds a task `testAll`  to test all the selected dbs. Right now the list of available databases is in an array.

I've updated the workflow on github so that it calls the new task, but I suppose there is a trade off   with speed (test are not running in parallel like before). I don't think it makes much difference now  with 3 dbs. It's nice not having to think about the wrorkflow when we add a new db.

I haven't changed the default behaviour so it shouldn't change anything for developers. The code is not great but my knowledge with groovy and gradle is limited.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,8 @@ on:
     branches: master
 
 jobs:
-  test_postgresql:
-    name: Test with PostgreSQL
+  test_dbs:
+    name: Test available databases
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -20,36 +20,8 @@ jobs:
         server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
         settings-path: ${{ github.workspace }} # location for the settings.xml file
     - name: Build and Test with PostgreSQL
-      run: ./gradlew build -Pdocker -Pdb=pg
+      run: ./gradlew testAll -Pdocker
```
aguibert, line 23: @DavideD This is simpler in the pipeline file, but the major drawback is that it will now run everything in sequence . Before, everything ran in parallel (jobs run in parallel if they have no dependencies on other jobs).
I'd really like to see a solution that runs everything in parallel so that CI completes faster, but I don't think that will be possible with an all-in-one gradle task.

aguibert: This issue will become more pronounced as we add more DBs too

DavideD: My main concerned her is to keep everything aligned and not having to touch CI, if possible, everytime we add a new database. But I will have a look to see if I can find a different solution

DavideD: We can also skip the commit that update the pipeline for now. I still like to test everything locally from time.

DavideD:
I'm going to close this for now. I don't think there is a way to make it work in gradle without  having multiple projects.

.github/workflows/build.yml
aguibert: Since the CI jobs are defined as a plain file along with the rest of the source files in the repo "touching CI" is not as bad as if all these jobs had to be manually set up by someone logging into a Jenkins UI or something.
Perhaps one option is to use the matrix functions in github to generate the matrix of jobs (quarkus does this afaik) but currently dynamic matrixes are not supported by GH actions

DavideD: Let's keep it as it is for now and we might evolve it later if we need to. Soon I will have to create a JOB on our CI and we can revisit the issue there. I don't think there is much we can do about this specific issue with the current structure of the project.

aguibert: "our CI" meaning some internal Hibernate CI? Is there a requirement to be using that? The current github actions setup works pretty well IMO (and it's free!)

PR #206, opened 2020-05-27
Fix #102 - publish artifacts to GitHub packages / Bintray

### LINKED ISSUE #102: Release process
We need to be able to release the artifacts via gradle.
Hibernate ORM already has a solution for this and we should aim to use the same approach.
Here's some of the files in ORM responsible for the release:

- https://github.com/hibernate/hibernate-orm/blob/master/release/release.gradle
- https://github.com/hibernate/hibernate-orm/blob/master/gradle/published-java-module.gradle
- https://github.com/hibernate/hibernate-orm/blob/master/gradle/publishing-repos.gradle
- https://github.com/hibernate/hibernate-orm/blob/master/gradle/publishing-pom.gradle

yrodiere:
This adds support for:

* Publishing actual releases to Bintray
* Publishing snapshots to GitHub Packages
* Publishing snapshots after every build on master

Bintray doesn't accept snapshot artifacts, so even if it's our platform of choice for artifacts that need to be synced to Maven Central, we need another one for snapshots.

You can find an example of what Bintray releases look like here: https://bintray.com/yrodiere/hibernate-release-testing/hibernate-reactive/0.0.1
I included instructions for releasing in the file `publish.gradle`.

I decided not to publish snpashots to JBoss Nexus, because it silently rejects some artifacts (accepts the upload, but does not display them in the repository afterwards). I didn't find a way to debug this: there's absolutely no feedback from the Nexus server and no way to know what the rules for rejecting artifacts are.

GitHub Packages accepts snapshot artifacts, but:

1. They are not kept separate from actual releases in the web interface, and there's apparently no way to do so.
2. They need to have a unique version number and they cannot be deleted, ever. So reusing "1.0.0-SNAPSHOT" for every snapshot cannot work, we need to assign a unique version number (based on the commit SHA) to every snapshot artifact.

You can find an example of what GitHub Packages releases look like here: https://github.com/yrodiere/hibernate-reactive/packages/244270?version=1.0.0-SNAPSHOT-e822255088ce1382cb103123f114404dbe26547e
I included instructions for publishing snapshots in the file `publish.gradle`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,23 +61,5 @@ jobs:
         java-version: 1.8
         server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
         settings-path: ${{ github.workspace }} # location for the settings.xml file
-    - name: Detect the version of Hibernate Reactive
-      id: detect-version
-      run: |
-        ./gradlew hibernate-reactive-core:properties | grep '^version:' | sed -E 's/version: (.*)/::set-output name=version::\1/'
-    - name: Generate a unique version
-      id: generate-unique-version
-      env:
-        UNIQUE_VERSION: ${{ format( '{0}-{1}', steps.detect-version.outputs.version, github.sha ) }}
-      run: echo "::set-output name=version::$UNIQUE_VERSION"
     - name: Create maven package
-      env:
-        ORG_GRADLE_PROJECT_projectVersion: ${{ steps.generate-unique-version.outputs.version }}
       run: ./gradlew assemble publishToMavenLocal
-    - name: Publish to GitHub Packages (SNAPSHOT only, push only, PR excluded)
-      env:
-        ORG_GRADLE_PROJECT_projectVersion: ${{ steps.generate-unique-version.outputs.version }}
-        ORG_GRADLE_PROJECT_githubUser: ${{ github.actor }}
-        ORG_GRADLE_PROJECT_githubToken: ${{ github.token }}
-      if: endsWith( steps.detect-version.outputs.version, '-SNAPSHOT' ) && github.event_name == 'push'
-      run: ./gradlew publish
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,15 +61,23 @@ jobs:
         java-version: 1.8
         server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
         settings-path: ${{ github.workspace }} # location for the settings.xml file
-    - name: Create maven package
-      run: ./gradlew assemble publishToMavenLocal
     - name: Detect the version of Hibernate Reactive
       id: detect-version
       run: |
         ./gradlew hibernate-reactive-core:properties | grep '^version:' | sed -E 's/version: (.*)/::set-output name=version::\1/'
-    - name: Publish to JBoss Nexus (SNAPSHOT only, push only, PR excluded)
+    - name: Generate a unique version
+      id: generate-unique-version
       env:
-        ORG_GRADLE_PROJECT_jbossNexusUser: ${{ secrets.jbossNexusUser }}
-        ORG_GRADLE_PROJECT_jbossNexusPassword: ${{ secrets.jbossNexusPassword }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -61,13 +61,23 @@ jobs:
...
+      env:
+        ORG_GRADLE_PROJECT_projectVersion: ${{ steps.generate-unique-version.outputs.version }}
+      run: ./gradlew assemble publishToMavenLocal
+    - name: Publish to GitHub Packages (SNAPSHOT only, push only, PR excluded)
+      env:
+        ORG_GRADLE_PROJECT_projectVersion: ${{ steps.generate-unique-version.outputs.version }}
+        ORG_GRADLE_PROJECT_githubUser: ${{ github.actor }}
+        ORG_GRADLE_PROJECT_githubToken: ${{ github.token }}
+      if: endsWith( steps.detect-version.outputs.version, '-SNAPSHOT' ) && github.event_name == 'push'
+      run: ./gradlew publish -x test
```
DavideD, line 83: If tests are missing, doesn't this mean that there are going to be test failures?

yrodiere: It's not a release, just a snapshot.

This runs after we run `./gadlew publishToMavenLocal` (in one of the previous steps), so unit tests (if any) have already run. I'll just remove this `-x test`, it's not useful.

Regarding integration tests... I suppose I could move this to a separate job and only publish after all other jobs have run.
That being said, I'm not sure it matters much: do you expect code merged to master to fail tests very often? Worst case, we just publish a snapshot that doesn't work, which means we'll publish another one soon after.

DavideD: I was just trying to understand the flow. I don't think that's intuitive when looking at the file.

DavideD: To be fair, enabling the tests won't test on all the dbs anyway.

Sanne:
Many thanks yoann, looks great. Do we need to do anything else (besides merging) to enable those new github actions?

yrodiere:
@Sanne For the GitHub actions, there's nothing to do beyond merging, it should work out of the box. No secrets needed in particular, GitHub creates authentication tokens automatically and provides them to the build.

Just make sure you're okay with publishing snapshots to GitHub packages, since you can't delete them afterwards. I suppose it could get quite crowded after a few years :)

Regarding releases, I already created [the Bintray package](https://bintray.com/hibernate/artifacts/hibernate-reactive). Maybe you'll want to review the info there, but it should be usable as-is. You just need a username and API token.

Going forward... There are two limitations:

1. Releases still have to be performed manually (create the tag, call Gradle, push, ...). I didn't include a release workflow in the GitHub actions; I just added something to publish snapshots at the end of every build.
2. I only looked at releasing to Maven repositories. I didn't do anything to publish documentation, a dist ZIP, etc.

I can do that as a second step if you want?

Sanne:
### snapshots published on github
right that could get crowded & I'm not sure of the benefits. Interesting to know that it's possible but maybe let's think carefully about that. Best to to comment that out for now?

### release automation
Could that be done via a github action? Clearly we don't want to release on each push, but maybe it can be triggered on tags being pushed?  If you could explore this, that would be awesome. I expect we'll be releasing quite often soon :)
Alternatively, a release job on ci.hibernate.org would also be welcome.

### ZIP distributions
Let's not do that, until someone persuades us otherwise. We did it on other projects but that's a decision made ~20y ago.. I wonder if that's still being useful and I suspect we're just doing it still as the infrastructure is there already.

### Documentation publishing
We don't have any yet :/

### Javadoc publishing
If you could look at this, that would be great. Last priority though - we don't have a website to point at them yet :)

yrodiere:
Alright, I added a commit to remove the automatic pushing to GitHub packages. Let's revert this later if necessary.

I'll work on release automation then javadoc tomorrow, first thing in the morning.

DavideD:
Thanks @yrodiere 
By the way, @dreab8 already did some work around this for ORM: https://github.com/hibernate/hibernate-orm/pull/3349

DavideD:
Rebased and merged https://github.com/hibernate/hibernate-reactive/pull/206

PR #214, opened 2020-05-28
Bootstrap and initialization rework

### LINKED ISSUE #40: Hibernate RX alongside regular Hibernate

Sanne:
A draft as it requires an ORM snapshot - I'd rather release an ORM version than use a snapshot.

But feel free to have a look already? The commits should be self descriptive.

Commit f72627c includes an hack I'm not particularly proud of, but will do for the time being as I believe better strategies would take too much time.

### JPA boot
People will need to explicitly opt-in for the Reactive provider:`<provider>org.hibernate.reactive.provider.ReactivePersistenceProvider</provider>`

### Hibernate boot methods
Users will need to replace using `StandardServiceRegistryBuilder` with `ReactiveServiceRegistryBuilder`. Everything else should be the same.

## Benefits:
 - no longer "taint" a classical ORM instance just by having Hibernate Reactive on classpath
 - more efficient, as we don't overlay custom services on top of the ones we mean to ignore but avoid to initialize them
 - simpler registry setup
 - easier future customizations of service and configuration peculiarities

Fixes #40


hibernate-reactive-core/src/main/java/org/hibernate/reactive/boot/impl/ReactiveServiceInitiators.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.List;
+
+public final class ReactiveServiceInitiators {
+
+    private ReactiveServiceInitiators() {
+    }
+
+    public static final List<StandardServiceInitiator> LIST = buildInitialServiceInitiatorList();
+
+    private static List<StandardServiceInitiator> buildInitialServiceInitiatorList() {
```
Sanne, line 51: @gavinking , @DavideD this list is rather important; it's similar and yet has crucial differences compared to `org.hibernate.service.StandardServiceInitiators`.

@aguibert this list will need to be adapted within Quarkus..  you should be able to paste it in the Quarkus extension and then use it to override the default (this one) list? We certainly don't want to use the same list as Hibernate ORM "vanilla" or the quarkus-hibernate-orm extension.

I predict some pain to maintain this as we know have 4 different variations of this list - defined in different projects. Ideas welcome to improve on this?

The good news is that a mistake in these lists should fail catastrophically, so testing might be good enough to not make it too bad.

gavinking:
@Sanne this seems perfectly reasonable to me.

Does this mean that we will boot with *fewer* services than Hibernate core usually has? (i.e. things we don't need get trimmed out?)

Sanne:
> Does this mean that we will boot with fewer services than Hibernate core usually has? (i.e. things we don't need get trimmed out?)

That will be true for Quarkus; for example the `JaccService` can (and must) be removed - but we're not removing such things from Hibernate Reactive when running standalone.

When in standalone (as in our tests) it's going to have the same exact number of initialized services as Hibernate ORM standalone; but we have other benefits: we will have fewer classes initialized and instantiated, as the services we won't need are being trimmed out.

For example, before this patch we'd initialize the default `JdbcEnvironmentInitiator`, and then after having loaded it we'd replace it with the `NoJdbcEnvironmentInitiator`. Now we just load the `JdbcEnvironmentInitiator` right away.

This is only a minor, subtle difference when running in a standard JVM, but it could make a significant difference in native as the dead code elimination can be more effective.

But also having a custom list of "Service initiators" just gives us a lot more flexibility, e.g. for Quarkus to inject a custom connection pool, etc..

Sanne:
We have the ORM release: merged!

PR #216, opened 2020-05-28
DDL generation for @Lob

gavinking:
This patch ensures that we generate the right column types for properties mapped `@Lob`.


gavinking:
Ups, so new tests added by @gbadner since I wrote this have exposed a bug in it.

gavinking:
There you go, @aguibert, HTH.

aguibert:
awesome, thanks!

PR #218, opened 2020-05-29
[#163] Create task to test all databases

DavideD:
I've cleaned up my previous PR and I think  this one adds a coupe of nice things:

- Add a rule to recognize task named `taskDb<dbName>`
- `testAll` will test a list of dbs
- A nicer summary at the end of the test telling how many of them have been skipped, passed or failed

The new tasks are only available for people wanting to use them, it's not going to affect CI or other workflow


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/218

PR #219, opened 2020-05-29
support for parameters in native and criteria queries

gavinking:
This patch extends @gbadner's work on query parameters by adding:

- positional + named parameters in native queries
- named + anonymous parameters in criteria queries

But for now I could not figure out how to add support for parameters in update/delete queries or criteria.


DavideD:
I've created a new issue for update queries parameters: https://github.com/hibernate/hibernate-reactive/issues/220

Issue #217, opened 2020-05-29
Features for Panache Reactive


FroMage:
Hi,

I've started to port Panache Reactive and found the following methods I used to call on `EntityManager` or the JPA `Query` to be missing:

- `Query.setLockMode` (the one taking no alias and a JPA lock enum). It appears not trivial to call the present version with aliases because the code paths differ, so not sure they're equivalent
- `Query.setHint`
- `Query.getQueryOptions` (required to get current paging info)
- `Multi<T> Query.getResuls()` for streaming results, but that one's probably planned for later and I can probably emulate it by wrapping the `Uni<List>` result for now

If those first three are easy to support, can we get them please?

gavinking:
Damn, so the very first actual user of this stuff is the guy who's using all the features where I said to myself "oh, nobody every uses that".

gavinking:
What is it that you're passing to `setHint()`?

gavinking:
I have given you two of your wishlist items, completely without any sort of testing at all.

FroMage:
Thanks :)

gavinking:
So to summarize where we're at on this one:

1. `Query.setLockMode` is done.
2. `Query.setHint()` is implemented but not exposed, because I can't find any hints where are meaningful for use:
    - the vertx client doesn't have timeouts yet
    - there is no notion of a fetch size
    - we have a `setComment()` method, no need for a hint for that
    - likewise, we have methods for setting the flush mode and lock mode
    - I have not tested fetch plans with queries, and if/when we do, I would add a method for it
    - we don't support the query cache yet (at least, we have not tested it), and when we do I would add methods for this
3. I have not done anything about `getQueryOptions()`. The only relevant options are first/max result, so perhaps I should just add getters for those two settings.
4. Streaming results is #153.

gavinking:
> I have not done anything about getQueryOptions(). The only relevant options are first/max result, so perhaps I should just add getters for those two settings.

Now I've added getters in 6ad870b3512ebc35bcbe5be2588c493f1b678c2a. 

@FroMage may I close this issue please?

FroMage:
Actually I just noticed you didn't give me `setLockMode(LockModeType)` which takes the JPA lock mode enum (not the Hibernate enum). Is that on purpose?

gavinking:
[...]

PR #222, opened 2020-05-29
add setHint() and overloaded setLockMode() to Query

gavinking:
For @FroMage and Panache.

See #217.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -110,10 +110,10 @@ public interface Mutiny {
 		 */
 		Query<R> setComment(String comment);
 
-		/**
-		 * Set a query hint.
-		 */
-		Query<R> setHint(String hintName, Object value);
+//		/**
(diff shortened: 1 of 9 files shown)


gavinking:
I have literally *no clue* if these method will actually do anything, but I'm relying on @FroMage to let me know.

gavinking:
Well, I added tests for `setLockMode()` so now I know one thing worx.

FroMage:
Thanks!

gavinking:
> Thanks!

If you actually tell me what you want to pass to `setHint()`, then I can add a test for that....

FroMage:
Huh, I'm not sure. Our test suite has this: `setHint(QueryHints.HINT_CACHEABLE, "true")`. Does this help?

gavinking:
> Does this help?

Well, I don't think we support query caching at all right now....

gavinking:
So looking over the options in `QueryHints`, I'm really not sure we want to support hints for now.

Most of that stuff that we *do* support can already be set via typesafe methods.

gavinking:
I've applied this pull, but I didn't expose `setHint()` for now, waiting for @FroMage to tell me what he wants to do with it and convince me we really want it.

Issue #211, opened 2020-05-28
Additional test cases for queries with parameters


DavideD:
This follows https://github.com/hibernate/hibernate-reactive/issues/120

If this task becomes too big we can split it.
We have all the logic for queries and parameters in place but some tests are missing:

- [x] native SQL queries
- [x] filters (see #26)
- [x] criteria queries

We can create new issues if we find problems along the way.

gavinking:
Wrong milestone :-)

DavideD:
By the way, What exactly FIRST CUT means ? is it the first release or the first final?

gavinking:
First release.

gavinking:
I've tested filters, and added tests, but there is a remaining problem #223 
Closing this issue.

PR #224, opened 2020-05-29
fixes and tests for filter parameters

gavinking:
This is for #211.

It's *mostly* working, except that filter parameters + collection fetching is currently broken. (I have fixed the SQL generation, but the parameter values don't get bound.)


gavinking:
> filter parameters + collection fetching is currently broken

I've opened #225.

DavideD:
If I rebase it to the latest changes some tests fail with MySQL

DavideD:
It seems that problem is on master, not this PR. sorry

PR #226, opened 2020-05-29
tests for eager fetching of collections

gavinking:
So apparently some good person (either @gbadner or @DavideD or perhaps a combination of both) already implemented support for eager select fetching for collections.

This is half of #39 done!

This was something we had agreed we didn't need to support yet, and it's a carefully-documented limitation in the README, but I'm certainly not complaining!

Anyway I've added a test, just in case we didn't have a very explicit one already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,14 +312,19 @@ features:
 - `@ManyToMany` associations, and
 - one-sided `@OneToMany` associations without `mappedBy`.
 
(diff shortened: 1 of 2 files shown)


DavideD:
THis fails for MySQL for me, if I rebase to the latest master

DavideD:
> THis fails for MySQL for me, if I rebase to the latest master
Sorry, wrong PR

Issue #223, opened 2020-05-29
Upgrade to Hibernate ORM 5.4.17.Final


Sanne:
it's available now, and we need it for various improvements.

The upgrade breaks something in `ReactiveQueryLoader` and `ReactiveCustomLoader`, I assume @DavideD knows what needs to be done?

gavinking:
Is it related to #207?

DavideD:
Working on it

DavideD:
The problem is that CachingReactiveLoader#bindPreparedStatement scope is in conflict with Loader#bindPreparedStatement

Adding the merhod in Loader was part of this PR: hibernate/hibernate-orm#3420
that was sent during the work to support parameters #197

Renaming the method in CachingReactiveLoader seems the simplest trick for now.

gavinking:
> The problem is that `CachingReactiveLoader#bindPreparedStatement` scope is in conflict with `Loader#bindPreparedStatement`

I understand, that's because we're going to delete `CachingReactiveLoader#bindPreparedStatement` :-)

gavinking:
(The relevant [discussion](https://github.com/hibernate/hibernate-reactive/pull/197#issuecomment-634522808))

PR #197, opened 2020-05-27
120 query parameter support

gbadner:
hibernate/hibernate-reactive#120

After force-pushing, this branch no longer requires https://github.com/hibernate/hibernate-orm/pull/3420.

This works for HQL queries. Some similar changes still need to be made for native queries.

More tests are needed for native queries, filters, and criteria queries.

PR #227, opened 2020-05-29
[#223] Upgrade to ORM 5.4.17.Final

DavideD:
@gavinking can you have a look at this?

The problem is that CachingReactiveLoader#bindPreparedStatement scope is in conflict with Loader#bindPreparedStatement

Adding the merhod in Loader was part of this PR: https://github.com/hibernate/hibernate-orm/pull/3420
that was sent during the work to support parameters https://github.com/hibernate/hibernate-reactive/pull/197

Renaming the method in CachingReactiveLoader seems the simplest trick for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ subprojects {
 // ./gradlew clean build -PhibernateOrmVersion=5.4.9-SNAPSHOT
 
 ext {
-    hibernateOrmVersion = '5.4.16.Final'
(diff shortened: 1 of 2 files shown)


gavinking:
@DavideD just push it.

(By the way, where I changed superclass methods to protected, it's almost certainly so we can actually *delete* the method of the subclass. But I'll do it.)

PR #122, opened 2020-05-18
#121 Upgrade to ORM 5.4.16.Final

DavideD:
We can merge it as soon as the release has finished

PR #228, opened 2020-05-29
Automated release

yrodiere:
Follows up on #122 

To release, just run this from your clone of the repo:

```
./gradlew ciRelease -PreleaseVersion=0.0.1.Alpha1 -PdevelopmentVersion=1.0.0-SNAPSHOT -PgitRemote=upstream -PgitBranch=master
```

This will switch to your local copy of the branch `master`, create a commit that changes the version to `0.0.1.Alpha1` and tag it as `0.0.1.Alpha1`, then create another commit that changes the version to `1.0.0-SNAPSHOT`, then push those commits to branch `master` on remote `upstream` along with the tag.

Then GitHub actions will run tests, and finally create the packages and publish them to bintray, here: https://bintray.com/beta/#/hibernate/artifacts/hibernate-reactive?tab=overview

I'll submit another PR later to also push the javadoc to jboss.org.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,8 +18,6 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-        server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
-        settings-path: ${{ github.workspace }} # location for the settings.xml file
     - name: Build and Test with PostgreSQL
       run: ./gradlew build -Pdocker -Pdb=pg
 
@@ -32,8 +30,6 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-        server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
-        settings-path: ${{ github.workspace }} # location for the settings.xml file
     - name: Build and Test with MySQL
       run: ./gradlew build -Pdocker -Pdb=mysql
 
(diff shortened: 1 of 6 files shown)


yrodiere:
Note that I didn't make it so that you would only have to tag a commit, because then the CI would have to make changes to the source (change the version to remove the `-SNAPSHOT` suffix) and the published binary would not be created with the _exact_ same source as the one you tagged.

I suppose we can find ways to improve on this, but this is a good start?

Sanne:
> Note that I didn't make it so that you would only have to tag a commit, because then the CI would have to make changes to the source (change the version to remove the `-SNAPSHOT` suffix) and the published binary would not be created with the _exact_ same source as the one you tagged.
> 
> That's why you need to call the `ciRelease` task locally to trigger the release.
> 
> I suppose we can find ways to improve on this, but this is a good start?

ah, that's indeed a bit surprising. But thinking about it, it's not too bad: it implies we won't see those annoying commits "setting version", "setting to snapshot" in the main timeline.

FWIW, in Infinispan the release process first branches off, then makes the change to the version, then it tags. I don't like it much becasue when looking at the git history it's _too clean_ : you don't see the previous tags on the timeline so it's hard to figure out if some commit was in/out a specific tag.

I agree it's a good start. Not sure if we should (can) keep this around for long.

Sanne:
many thanks @yrodiere ! Let's try it out..

PR #231, opened 2020-05-29
Add missing name and description to POM

yrodiere:


Sanne:
thanks, I keep forgetting about that ordering rule. Merged and followed up with 5a107d77663c72b56fe01b4c212732e16f270fba

Issue #204, opened 2020-05-27
lazy property fetching


gavinking:
I *believe*, but have not tested that lazy property fetching via bytecode enhancement is non-working.

The basis for this belief is that I seem to recall taking a couple of shortcuts when getting much more important features working.

So we should:

1. write some tests to be certain one way or another, and
2. investigate possibly fixing it, or else
3. document the limitation, if necessary.

gavinking:
So, I'm playing about with the bytecode enhancer with Hibernate Reactive, just to see if I can break it, and I noticed the following thing that I find quite confusing:

In `AbstractEntityPersister` we have:

```java
	public boolean hasProxy() {
		// skip proxy instantiation if entity is bytecode enhanced
		return entityMetamodel.isLazy() && !entityMetamodel.getBytecodeEnhancementMetadata().isEnhancedForLazyLoading();
	}
```
But this naïvely seems wrong to me, and inconsistent with what `DefaultLoadEventListener` is expecting.

It looks correct when `hibernate.bytecode.allow_enhancement_as_proxy` is explicitly set to `true`, but not for the default value.

It seems to me that the implementation should be:

```java
boolean hasUnenhancedProxy() {
		// skip proxy instantiation if entity is bytecode enhanced
		return getEntityMetamodel().isLazy()
				&& !( getEntityMetamodel().getBytecodeEnhancementMetadata().isEnhancedForLazyLoading()
					&& getFactory().getSessionFactoryOptions().isEnhancementAsProxyEnabled() );
	}
```

Since we only want to disable proxies when we're using the "enhanced proxies".

There's surely something I'm missing here, since I've never used any of this stuff before.

@Sanne do you understand this?

gavinking:
Well, I've played with this a bit not, and the situation doesn't seem terrible at all: it would be rather straightforward to make it work if we just expose some things in `AbstractEntityPersister` to subclasses.

In particular:

- the `lazyPropertyXxxxxx` attributes need to be exposed via getters, and
-we'll need to write a version of `initializeLazyProperty()` that returns a `CompletionStage`, and 
- call it from our `fetch()` method.

This is going to mean a bit of a copy/paste job but nothing too terrible.

Anyway, I'm not motivated to work on this right now

gavinking:
I have opened https://github.com/hibernate/hibernate-orm/pull/3421 with the required changes to `hibernate-core`.

gavinking:
[...]

PR #233, opened 2020-05-29
lazy property fetching via session.fetch()

gavinking:
So this here's my implementation of #204.

WDYT?

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -18,3 +18,25 @@ dependencies {
     runtimeOnly 'org.slf4j:slf4j-log4j12:1.7.30'
 }
 
+// All of the remaining configuration is only necessary to enable
+// the Hibernate bytecode enhancer and field-level lazy fetching.
+// (This is very optional!)
+
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath "org.hibernate:hibernate-gradle-plugin:${hibernateOrmVersion}"
+    }
+}
+
+apply plugin: 'org.hibernate.orm'
+
+hibernate {
+    enhance {
+        enableLazyInitialization = true
+        enableDirtyTracking = true
+        enableAssociationManagement = false
+    }
+}
--- a/example/src/main/java/org/hibernate/example/reactive/Main.java
+++ b/example/src/main/java/org/hibernate/example/reactive/Main.java
@@ -46,7 +46,11 @@ public class Main {
 				//retrieve a Book
 				session -> session.find(Book.class, book1.id)
 						//print its title
-						.thenAccept( book -> out.println(book.title + " is a great book!") )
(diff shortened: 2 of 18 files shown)


example/src/main/java/org/hibernate/example/reactive/Main.java
```diff
@@ -46,7 +46,11 @@ public static void main(String[] args) {
 				//retrieve a Book
 				session -> session.find(Book.class, book1.id)
 						//print its title
-						.thenAccept( book -> out.println(book.title + " is a great book!") )
+						.thenCompose( book -> {
+							out.println(book.title + " is a great book!");
+							return session.fetch( book, Book_.isbn )
+									.thenAccept( isbn -> out.println(isbn + " is the ISBN") );
+						} )
```
gavinking, line 53: Folks, this is what it looks like to use this feature.

Anybody have any thoughts? Is this a good feature for us to have?

example/build.gradle
```diff
@@ -1,5 +1,24 @@
...
+    }
+}
+
+apply plugin: 'org.hibernate.orm'
+
+hibernate {
+    enhance {
+        enableLazyInitialization= true
+        enableDirtyTracking = true
+        enableAssociationManagement = true
```
DavideD, line 18: Do you need all these 3 properties enabled?

gavinking: *That* I am not sure of. I *presume* we don't need `enableDirtyTracking`, but if we're going to enable the bytecode processor we might as well? It help, I presume?

And now that I realize what `enableAssociationManagement` is, I think I *don't* want it.

gavinking:
Alright, I've no added a test for this stuff. I'm inclined to merge it.

example/src/main/java/org/hibernate/example/reactive/Main.java
DavideD: LGTM

example/build.gradle
```diff
@@ -18,3 +18,25 @@ dependencies {
...
+        classpath "org.hibernate:hibernate-gradle-plugin:${hibernateOrmVersion}"
+    }
+}
+
+apply plugin: 'org.hibernate.orm'
+
+hibernate {
+    enhance {
+        enableLazyInitialization = true
+        enableDirtyTracking = true
```
DavideD, line 39: If I disable dirtyTracking (`enableDirtyTracking=false`) the example fails when I run it. is that expected?

gavinking: It's working for me with:

    enableDirtyTracking = false

So, I dunno...

gavinking: I'm going to merge it @DavideD, I doubt that weird behavior of the bytecode enhancer is related to any bug in my code.

PR #234, opened 2020-05-29
Parameter processing

gavinking:
So initially we were trying to deal with Postgres-style `$n` parameters by getting all up in Hibernate's SQL generation and doing it at the source, by overriding the methods which actually generate the SQL. The reasoning was supposed to be that it would alleviate the need to re-tokenize the SQL when the SQL is executed, something that Hibernate has always preferred.

I think it's fair to say that this approach has pretty much failed, and made the code much uglier.

The pull request gives up on that goal, and just does the parameter substitution *iff* the database is Postgres, right before calling `preparedQuery()`.

Now, this could (or could not), in principle, have an impact on performance in Postgres.

But if it does then I say the fix would be to go back to the source and change the original SQL generation code in Hibernate core, instead of trying to hack it into HR.

Thoughts?


gavinking:
I need to hear if anyone objects to doing this, because if not I'm going to merge it. @Sanne @DavideD ?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
```diff
@@ -0,0 +1,76 @@
...
+	private boolean inQuoted;
+	private boolean inSqlComment;
+	private boolean inCComment;
+	private boolean escaped;
+	private int count = 0;
+	private StringBuilder result;
+	private int previous;
+
+	private Parameters(String sql) {
+		result = new StringBuilder( sql.length() + 10 );
```
DavideD, line 20: What's the 10 about?

gavinking: It's an estimate of how many parameters we're likely to have to expand. Obviously it's not a strict upper bound, especially in the case of `insert`s and `update`s. Perhaps I should increase it for inserts and updates.

(The purpose is to avoid having to grow the `StringBuilder`.)

gavinking: I did consider scanning the string for `?`s but it wasn't clear to me if doing this *every time*, this was faster or slower than *occasionally* growing the `StringBuilder`.

Sanne:
Could someone explain the high level purpose of this?

I'm surprised we need to apply any change to how SQL is generated and parameters are being handled, as it was a project goal to reuse what Hibernate ORM is able to rather than fork such essential aspects?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
DavideD: I kinda guess that, but this is the kind of number that people in the future will wonder about if you don't add a comment to it

gavinking:
@Sanne well this was an issue I reported to the SQL client months ago. It's currently being discussed [here](https://github.com/eclipse-vertx/vertx-sql-client/issues/509).

Suffice to say I seem to be losing the argument.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
gavinking: Actually this is stupid. I actually know how many parameters I have, don't I?

gavinking: I've fixed it to use the exact number of parameters.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/testing/TestingRegistryRule.java
```diff
@@ -71,10 +82,64 @@ public ServiceRegistry getParentServiceRegistry() {
...
         public <R extends Service> R getService(Class<R> serviceRole) {
-            if ( serviceRole == VertxInstance.class )
+            if ( serviceRole == VertxInstance.class ) {
                 return (R) vertxService;
+            }
+            else if ( serviceRole == JdbcEnvironment.class ) {
+                return (R) new JdbcEnvironment() {
+                    @Override
+                    public Dialect getDialect() {
+                        return new MySQL8Dialect();
```
DavideD, line 94: Why do you need this change to TestingRegistryRule?

gavinking: To avoid an NPE, IIRC.

Comment it out; see what happens.🤷‍♂️

gavinking:
> This actually simplify the code

Yes, that was the whole point.

gavinking:
Let's see if I can squash this. Last time I tried I failed.

gavinking:
Yup, it worked this time.

gavinking:
Merged as a single commit.

Issue #220, opened 2020-05-29
Support parameters for update/delete queries and criteria


DavideD:
We already have an ignored test in the code base: [HQLUpdateQueryTest#testUpdateQueryWithParameters ](https://github.com/hibernate/hibernate-reactive/commit/4cc82e9bad11e15495bd5a396cc62db1b5fd36cd#diff-2c741543d9b4ba75d4aca7d803b2344cR72)

gavinking:
This is done.

Thanks, @gbadner for #236.

I tested criteria queries and her fix also works there.

And I made it work for native SQL update/deletes in #238.

PR #235, opened 2020-05-30
Support parameters for update/delete queries

gbadner:
See #220

Currently, multi-table updates and deletes are not supported.

This fix makes `HQLUpdateQueryTest#testUpdateQuery` pass. I have not tried other tests yet.

I suspect we would be better off using something like `BasicExecutor` instead. I'll look into this further next week,


gbadner:
Ignore this. I'm going to recreate...

Issue #239, opened 2020-05-30
properties for configuring the Vert.x connection pool


gavinking:
We need to add config properties to allow you to config the Vert.x connection pool, something like:

- `hibernate.vertx.prepared_statement_cache.max_size`
- `hibernate.vertx.prepared_statement_cache.sql_limit`
- `hibernate.vertx.pool.max_wait_queue_size`

Or perhaps:

- `hibernate.reactive.prepared_statement_cache.max_size`
- `hibernate.reactive.prepared_statement_cache.sql_limit`
- `hibernate.reactive.pool.max_wait_queue_size`

(I'm afraid I cannot explain why hibernate config properties have underscores in them, given that I've always been a zealoted lover of camelcase. A single moment of weakness...)

PR #240, opened 2020-05-30
add properties for configging the Vert.x pool

gavinking:
See #239.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
@@ -188,8 +188,8 @@ public class SqlClientPool implements ReactiveConnectionPool, ServiceRegistryAwa
 			connectOptions.setPassword( password );
 		}
 
-		//enable the prepared statement cache by default
-		connectOptions.setCachePreparedStatements(true);
+		//enable the prepared statement cache by default (except for DB2)
+		connectOptions.setCachePreparedStatements( !scheme.equals("db2") );
 
 		final Integer cacheMaxSize = ConfigurationHelper.getInteger( PREPARED_STATEMENT_CACHE_MAX_SIZE, configurationValues );
 		if (cacheMaxSize!=null) {
@@ -199,6 +199,7 @@ public class SqlClientPool implements ReactiveConnectionPool, ServiceRegistryAwa
 			}
 			else {
 				CoreLogging.messageLogger(SqlClientPool.class).infof( "HRX000015: Prepared statement cache max size: %d", cacheMaxSize );
+				connectOptions.setCachePreparedStatements(true);
 				connectOptions.setPreparedStatementCacheMaxSize(cacheMaxSize);
 			}
 		}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
@@ -188,17 +188,24 @@ public class SqlClientPool implements ReactiveConnectionPool, ServiceRegistryAwa
 			connectOptions.setPassword( password );
 		}
 
+		//enable the prepared statement cache by default
+		connectOptions.setCachePreparedStatements(true);
+
 		final Integer cacheMaxSize = ConfigurationHelper.getInteger( PREPARED_STATEMENT_CACHE_MAX_SIZE, configurationValues );
 		if (cacheMaxSize!=null) {
-			CoreLogging.messageLogger(SqlClientPool.class).infof( "HRX000014: Prepared statement cache max size: %d", cacheMaxSize );
-			connectOptions.setCachePreparedStatements(true);
-			connectOptions.setPreparedStatementCacheMaxSize(cacheMaxSize);
+			if (cacheMaxSize <= 0) {
+				CoreLogging.messageLogger(SqlClientPool.class).infof( "HRX000014: Prepared statement cache disabled", cacheMaxSize );
+				connectOptions.setCachePreparedStatements(false);
+			}
+			else {
+				CoreLogging.messageLogger(SqlClientPool.class).infof( "HRX000015: Prepared statement cache max size: %d", cacheMaxSize );
+				connectOptions.setPreparedStatementCacheMaxSize(cacheMaxSize);
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


gavinking:
Does anyone have any opinion on this one? Property names, etc?

Or should I just merge it?

Sanne:
Looks good; just wondering if these three proprties shouldn't rather share the same prefix?

- `hibernate.vertx.pool.max_wait_queue_size`
- `hibernate.vertx.prepared_statement_cache.max_size`
- `hibernate.vertx.prepared_statement_cache.sql_limit`

I could see them as

- `hibernate.vertx-pool.max_wait_queue_size`
- `hibernate.vertx-pool.prepared_statement_cache.max_size`
- `hibernate.vertx-pool.prepared_statement_cache.sql_limit`

If not, feel free to merge it.

Bear in mind that Quarkus won't be using this pool, but inject the existing managed vertx pool;
might want to mimick the same configuration options from here?

 - https://quarkus.io/guides/reactive-sql-clients#reactive-datasource

One interesting thing we do in Quarkus, is that different properties are available depending on the driver capabilities; e.g. MariaDB/MySQL expose charset and collation, others don't.

gavinking:
>  just wondering if these three proprties shouldn't rather share the same prefix?

The reason is that in vert.x itself, they're methods of different `Options` objects.

gavinking:
> Bear in mind that Quarkus won't be using this pool, but inject the existing managed vertx pool;
might want to mimick the same configuration options from here?

*Definitely*. @aguibert can you make it so?

gavinking:
> > just wondering if these three proprties shouldn't rather share the same prefix?
> 
> The reason is that in vert.x itself, they're methods of different `Options` objects.

(Surprisingly, the prepared statement cache belongs to the database connection options, not to the pool options. I cannot myself explain this. Perhaps @aguibert or @vietj can.)

Sanne:
> > Bear in mind that Quarkus won't be using this pool, but inject the existing managed vertx pool;
> > might want to mimick the same configuration options from here?
> 
> _Definitely_. @aguibert can you make it so?

I meant naming suggestions could flow from Quarkus into this project, since the same properties are already exposed in Quarkus.

gavinking:
So:

- `quarkus.datasource.reactive.postgresql.cache-prepared-statements`
- `quarkus.datasource.reactive.mysql.cache-prepared-statements`

_Really?_ Those don't see like the right names for us? We have no concept of a "datasource", and we're not in Quarkus.

I also don't see why in Quarkus the exact same config property has different names depending on what the database is. That looks incorrect given the current model in Vert.x. (Which, in fairness, has been somewhat in flux.)

And:

- `quarkus.datasource.reactive.url`
- `quarkus.datasource.reactive.max-size`
- `quarkus.datasource."datasource-name".username`
- `quarkus.datasource."datasource-name".password`

Hibernate has preexisting properties for those, which we're reusing. I don't want to change these, since we have no concept of a "datasource", nor of a datasource name, and we're not in Quarkus.

And Quarkus apparently has no properties corresponding to:

- `hibernate.vertx.prepared_statement_cache.max_size`
- `hibernate.vertx.prepared_statement_cache.sql_limit`

On the other hand, we don't support:

- `quarkus.datasource.reactive.postgresql.pipelining-limit`

Perhaps I should add it.

So, um, anyway, how about no?

Sanne:
> So, um, anyway, how about no?

I wasn't `recommending` it, just meant to make you aware of the existing ones for you to evaluate options.

+1 to stick to Hibernate conventions for the Hibernate projects.

Sanne:
thanks, merged these.

You can add pipelinining later if you wish - it doesn't seem a blocker to me to not have it right now :)

aguibert:
> (Surprisingly, the prepared statement cache belongs to the database connection options, not to the pool options. I cannot myself explain this.

This is because prepared statement caching is done at the common layer (along with connection pooling too)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -41,7 +41,21 @@
...
+	 */
+	public static final String PREPARED_STATEMENT_CACHE_SQL_LIMIT = "hibernate.vertx.prepared_statement_cache.sql_limit";
+	/**
+	 * @see SqlConnectOptions#setPreparedStatementCacheMaxSize(int)
+	 */
+	public static final String PREPARED_STATEMENT_CACHE_MAX_SIZE = "hibernate.vertx.prepared_statement_cache.max_size";
+	/**
+	 * @see PoolOptions#setMaxWaitQueueSize(int)
+	 */
+	public static final String MAX_WAIT_QUEUE_SIZE = "hibernate.vertx.pool.max_wait_queue_size";
```
aguibert, line 55: Do these options belong somewhere else if we expect a user to be setting them? We used to have an `AvailableReactiveSettings` interface for reactive-only settings, seems like that would be a proper place for these

gavinking:
> This is because prepared statement caching is done at the common layer (along with connection pooling too)

@aguibert I don't think that can be the explanation, since `PoolOptions` is also part of the common layer.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
gavinking: *Eventually*, yes. Indeed I *just added* a package called `common` which may be a reasonable place to put this.

But what I definitely don't want is classes with useful user-visible things just popping up randomly in packages that we haven't defined as part of the public API.

aguibert:
sorry, misunderstood initially. 

I think it's on `SqlConnectOpions` for 2 reasons:
1. Prep. stmt caching is done at the per-connection level, so configurations of that cache are really a configuration option of the connection itself
2. PoolOptions (although the name is a bit ambiguous) only refers to connection pooling, not pstmt pooling.

gavinking:
@aguibert OK, I mean that does sorta make sense, except that (at least as a user) the `ConnectOptions` is not something I pass to the pool when I create a connection. Just like the `PoolOptions`, it's a thing I pass to create the *pool*.

And it's really a bit surprising (again, for the user) to see the prepared statement cache configured alongside the properties I need for *connecting* to the database, on a class called `ConnectOptions`, when the connection pooling properties aren't.

aguibert:
Agreed, but to look at it from a different perspective, if you just establish a single connection (i.e. no `Pool` involved) you still are able to take advantage of prepared statement caching, which prevented it from being added to `PoolOptions`. We could have made a `StatementPoolOptions` or something, but that seemed rather tedious for not much added value

gavinking:
OK, fine. I had not realized that there was even an API for creating a single connection.

PR #241, opened 2020-05-31
clean up semantics of createNativeQuery()

gavinking:
This work cleans up the API for creating native SQL queries:

- introduces `ResultSetMapping` to reify `@SqlResultSetMapping`
- allows non-entity classes to be passed to `createNativeQuery()`, just as they are allowed as arguments of `createQuery()`

The JPA spec is quite vague in the area of native SQL queries, and it's possible that this improves on JPA and Hibernate ORM.


hibernate-reactive-core/src/main/java/org/hibernate/reactive/ResultSetMapping.java
```diff
@@ -0,0 +1,11 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
```
gavinking, line 6: I probably should make a subpackage to put this interface in.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -440,7 +520,7 @@ protected ReactiveHQLQueryPlan getQueryPlan(String query, boolean shallow) throw
 	}
 
 	private <T> ReactiveNativeQuery<T> createReactiveNativeQuery(NamedSQLQueryDefinition queryDefinition, Class<T> resultType) {
-		if ( resultType != null && !Tuple.class.equals( resultType ) ) {
+		if ( resultType != null && !Tuple.class.equals( resultType ) && !Object[].class.equals( resultType ) ) {
```
gavinking, line 523: I feel like this is a but in Hibernate ORM.

gavinking:
Note that I've designed this API in anticipation of:

<https://hibernate.atlassian.net/browse/HHH-14051>

gavinking:
Any objections to this change? I believe it's the right thing to do.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
```diff
@@ -175,12 +177,18 @@ public boolean contains(Object entity) {
 
 	@Override
 	public <R> Mutiny.Query<R> createNativeQuery(String sql, Class<R> resultType) {
-		return new MutinyQueryImpl<>( delegate.createReactiveNativeQuery( sql, resultType ) );
+		try {
+			delegate.getFactory().getMetamodel().entityPersister(resultType);
+			return new MutinyQueryImpl<>( delegate.createReactiveNativeQuery( sql, resultType ) );
+		}
+		catch (MappingException me) {
```
Sanne, line 184: Could you check if the persister exists rather then relying on an exception?

gavinking: If you can find me a method in Hibernate that does that. I could not.

Sanne: ok I'll amend it after this is merged.

gavinking:
I've moved `ResultSetMapping` to its own package, and added javadoc. Going to merge this now.

PR #242, opened 2020-06-02
Uppercase all GitHub secret names

yrodiere:
GitHub changed some things last weekend, and they are apparently pushing
for secret names to be all-caps.

I already changed the repo configuration accordingly.


gavinking:
So is there anything blocking this? Can't we just push it?

yrodiere:
Nothing blocking from my side.

gavinking:
Thanks.

PR #244, opened 2020-06-02
Showcase the latest released version

Sanne:
This showcases the latest released version, and links to dependency instructions.

Should we add it?


gavinking:
Sure. Thanks.

PR #245, opened 2020-06-02
add package-level javadoc and move services into provider package

gavinking:
The `boot` package is gone and all that stuff now lives under `provider`.

This eliminates confusing cross-package dependencies.


gavinking:
I'm merging this because it's needed for the Javadocs to not look like rubbish.

PR #246, opened 2020-06-02
Sync artifacts with Maven Central upon release using the bintrayPublish task

yrodiere:
I wasn't able to test this all the way through, for obvious reasons, but I checked that the call to the Bintray Rest API responsible for triggering the Maven Central sync actually happens.

I added the necessary secrets to the repository, but I'm not sure I have the correct username/password. What I found was the username and password for the Sonatype OSSRH JIRA account, so I used that, hoping a single set of credentials is used throughout all Sonatype applications.

Normally, the user and password are saved in the configuration of the Hibernate Organization on Bintray, but I don't have administrative rights to that organization.
@Sanne Could you please check the username password under the "Accounts" section of the Hibernate Organization on Bintray match the username and password we have in our Google Doc?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,4 +96,6 @@ jobs:
       env:
         ORG_GRADLE_PROJECT_bintrayUser: ${{ secrets.BINTRAY_USER }}
         ORG_GRADLE_PROJECT_bintrayKey: ${{ secrets.BINTRAY_KEY }}
-      run: ./gradlew bintrayUpload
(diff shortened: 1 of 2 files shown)


Sanne:
@yrodiere I couldn't check those secrets, I guess we'll try. Many thanks!

Sanne:
@yrodiere FYI it worked. 1.0.0.Alpha5 was fully automated, sync included!

yrodiere:
Great, thanks for letting me know!

PR #250, opened 2020-06-03
Cleanup items for Quarkus integration

### LINKED ISSUE #249: Build fails on JDK 11
Trying to build using JDK 11 now fails with the following error:

```
$ ./gradlew assemble
> Configure project :documentation
Forcing Javadoc in Java 8 compatible mode
FAILURE: Build failed with an exception.
* Where:
Build file '/mnt/c/dev/proj/hibernate-reactive/documentation/build.gradle' line: 67
* What went wrong:
A problem occurred evaluating project ':documentation'.
> Could not get unknown property 'baselineJavaVersion' for project ':documentation' of type org.gradle.api.Project.
```

aguibert:
Fixes #249

### DIFF
--- a/documentation/build.gradle
+++ b/documentation/build.gradle
@@ -57,16 +57,7 @@ task aggregateJavadocs(type: Javadoc, group: 'Documentation') {
 				"https://docs.jboss.org/hibernate/orm/${hibernateOrmVersion.family}/javadocs/".toString()
 		]
 		
-		if ( JavaVersion.current().isJava11Compatible() ) {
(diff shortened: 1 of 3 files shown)


aguibert:
Added 2 more commits onto this PR beyond the original commit, both are pretty minor so I didn't want to open separate PRs for each

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/250

Thanks!
I've merged a different commit for fixing the build though: 0d54ec09512b437bdbee0f019364c396636a1e92
because I  prefer to keep the explanation about setting the property.

PR #251, opened 2020-06-04
Add tasks to run main classes in the example

DavideD:
Is this helpful?

I've added while I was playing with the example, maybe it can be helpful for a user or for checking the example during the build.

I don't have strong opinion


gavinking:
Sure, why not. Certainly doesn’t hurt.

But what I *really* want is for our test suite to also smoke test the example.

DavideD:
I will include this in an other pr

PR #252, opened 2020-06-04
Isolate code required for MySQL in test suite

DavideD:
There are a couple of lines that are necessary for our tests
  to work with MySQL but shouldn't impact users.


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/252

PR #254, opened 2020-06-05
Refactoring for DatabaseConfiguration

DavideD:
This code PR contains several minor changes:

* Extract version of Vert.x and Testcontainers in properties
* Allow all the databases to have aliases (not only Postgres)
* Upgrade testcontainers to version 1.14.3 (because, why not?)

None of these changes are essential but I think they will come in handy when adding new databases.


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/254

It is pretty easy to revert if somebody have issues with it.
Cheers.

PR #256, opened 2020-06-09
#253 Smoke test examples

DavideD:
Issue https://github.com/hibernate/hibernate-reactive/issues/253

Run the examples on MySQL and PostgreSQL.

### DIFF
--- a/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
+++ b/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
@@ -15,13 +15,16 @@ import static org.hibernate.reactive.mutiny.Mutiny.fetch;
  */
 public class MutinyMain {
 
+	// The first argument can be used to select a persistenceUnit.
+	// Check resources/META-INF/persistence.xml for available names.
 	public static void main(String[] args) {
+		out.println( "== Mutiny API Example ==" );
 
 		// obtain a factory for reactive sessions based on the
 		// standard JPA configuration properties specified in
 		// resources/META-INF/persistence.xml
 		SessionFactory factory =
-				createEntityManagerFactory("example")
+				createEntityManagerFactory( persitenceUnitName( args ) )
 						.unwrap(SessionFactory.class);
 
 		// define some test data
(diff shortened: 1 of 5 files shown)


example/src/main/resources/META-INF/persistence.xml
```diff
@@ -3,7 +3,7 @@
              xsi:schemaLocation="http://java.sun.com/xml/ns/persistence http://java.sun.com/xml/ns/persistence/persistence_2_0.xsd"
              version="2.0">
 
-    <persistence-unit name="example">
+    <persistence-unit name="postgresqlExamplePU">
```
gavinking, line 6: Very minor but I would prefer `"postgresql-example"`.

example/src/main/resources/META-INF/persistence.xml
```diff
@@ -15,9 +15,39 @@
...
+
+            <!-- SQL statement logging -->
+            <property name="hibernate.format_sql" value="true"/>
+            <!--property name="hibernate.show_sql" value="true"/-->
+
+        </properties>
+
+    </persistence-unit>
+
+    <persistence-unit name="mysqlExamplePU">
```
gavinking, line 40: Very minor but I would prefer `"mysql-example"`.

example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
```diff
@@ -15,13 +15,22 @@
...
 public class MutinyMain {
 
+	private static final String DEFAULT_PU = "postgresqlExamplePU";
+
+	// The first argument can be used to select a persistenceUnit.
+	// Check resources/META-INF/persistence.xml for available names.
+	public static String persitenceUnitName(String[] args) {
+		return args.length > 0 ? args[0] : DEFAULT_PU;
+	}
+
```
gavinking, line 25: I would prefer to move all this stuff down to the bottom of the file so it's not the first thing people see.

Also there's no need to make a `private static final String DEFAULT_PU` for a string literal that is used in exactly one place.

gavinking:
@DavideD I left a couple of small comments.

DavideD:
Fixes #253 

Applied the changes and merged

Thanks

Issue #281, opened 2020-07-10
Duplicate ID exception on delete statement


FroMage:
This might be a user issue, but I don't get it.

I do a bunch of operations in my tests, then in the end I call "delete from Person" to delete all my entities, and I get this exception:

```
io.vertx.pgclient.PgException: duplicate key value violates unique constraint "person2_pkey"
	at io.vertx.pgclient.impl.codec.ErrorResponse.toException(ErrorResponse.java:29)
	at io.vertx.pgclient.impl.codec.QueryCommandBaseCodec.handleErrorResponse(QueryCommandBaseCodec.java:57)
	at io.vertx.pgclient.impl.codec.ExtendedQueryCommandBaseCodec.handleErrorResponse(ExtendedQueryCommandBaseCodec.java:70)
	at io.vertx.pgclient.impl.codec.PgDecoder.decodeError(PgDecoder.java:233)
	at io.vertx.pgclient.impl.codec.PgDecoder.decodeMessage(PgDecoder.java:122)
	at io.vertx.pgclient.impl.codec.PgDecoder.channelRead(PgDecoder.java:102)
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:163)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:714)
[...]

Issue #280, opened 2020-07-10
PgException leaks to the user


FroMage:
I attempted to persist an entity with a unicity constraint which should fail, and I got this:

```
io.vertx.pgclient.PgException: duplicate key value violates unique constraint "uk_1l3x2swifgj81xgvqivp7teto"
	at io.vertx.pgclient.impl.codec.ErrorResponse.toException(ErrorResponse.java:29)
	at io.vertx.pgclient.impl.codec.QueryCommandBaseCodec.handleErrorResponse(QueryCommandBaseCodec.java:57)
	at io.vertx.pgclient.impl.codec.ExtendedQueryCommandBaseCodec.handleErrorResponse(ExtendedQueryCommandBaseCodec.java:70)
	at io.vertx.pgclient.impl.codec.PgDecoder.decodeError(PgDecoder.java:233)
	at io.vertx.pgclient.impl.codec.PgDecoder.decodeMessage(PgDecoder.java:122)
	at io.vertx.pgclient.impl.codec.PgDecoder.channelRead(PgDecoder.java:102)
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:163)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:714)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:650)
[...]

PR #283, opened 2020-07-14
[#281] Return PersistenceException when constraint violation occurs

DavideD:
See #280


gavinking:
@DavideD it looks fine to me.

DavideD:
Done #283

Thanks!

Issue #282, opened 2020-07-12
Session/EntityManager is closed and Unexpected access to the database


mgonzaga1990:
Hi,

Would like to seek your advice regarding on my issue.

So I have an angular front end which is calling my some backend api. One of which is my api that fetches list of person.
at first i dont get any issues.. but when i refresh my angular page in the browser by hitting F5. I'm getting this error.

```
Caused by: java.lang.IllegalStateException: Session/EntityManager is closed

2020-07-13 03:33:54,324 ERROR [org.hib.AssertionFailure] (executor-thread-1) HHH000099: an assertion failure occurred (this may indicate a bug in Hibernate, but is more likely due to unsafe use of the session): org.hibernate.AssertionFailure: Unexpected access to the database
2020-07-13 03:33:54,331 ERROR [org.jbo.res.res.i18n] (executor-thread-1) RESTEASY002020: Unhandled asynchronous exception, sending back 500: org.hibernate.AssertionFailure: Unexpected access to the database
```

This is my first time seeing this kind of error don't kind of lost on which part I missed or if its a bug.

here is my DAO class.

```
@Inject
 Uni<Mutiny.Session> mutinySession;
```

```
public Uni<List<Person>> getAll(){
        return this.mutinySession
                .flatMap(session -> session.createNamedQuery("Person.findAll",Person.class)
                        .getResultList())
                .onFailure().invoke(ex-> new PersonException(ex.getMessage(), ALL_FETCH));
 }`

`public Uni<Person[]> getAll(String sort, String order, Long pageNumber,Integer size){
        return this.mutinySession
                .flatMap(session -> {
                    var sql = "SELECT * " +
                            "FROM PERSON ORDER by :sort " + order + " " +
                            "OFFSET ( :size * ( (:pageNumber) - 1)) ROWS " +
                            "FETCH NEXT :size ROWS ONLY";
                    return session.createNativeQuery(sql,Person.class)
                            .setParameter("sort",sort)
                            .setParameter("pageNumber",pageNumber + 1)
                            .setParameter("size",size)
                            .getResultList();
                })
                .map(persons -> persons.toArray(new Person[persons.size()]))
                .onFailure().invoke(ex-> new PersonException(ex.getMessage(),ALL_FETCH));
 }
```

My code for Save and update is working so far, so i didn't put it here.

Thanks for your help.

mgonzaga1990:
[...]

PR #286, opened 2020-07-16
[#282] Throw LazyInitializationException

DavideD:
https://github.com/hibernate/hibernate-reactive/issues/282


gavinking:
I suppose that's OK. It doesn't feel like the ideal place to do this, but if there's nowhere else we can reasonably do it, it's fine I guess.

DavideD:
Thinking about it more,  I'm not even sure this is a LazyInitializationException. This is not related to the session being closed prematurely.

DavideD:
Well, I suppose it is good enough for now

Merged https://github.com/hibernate/hibernate-reactive/pull/286

gavinking:
Well I do think it's legitimately a `LazyInitializationException`, which has the semantics of "we can't fetch this unfetched association for you right now", even if the reason we can't fetch it is a little different.

DavideD:
I agree, it's just that the javadoc for LazyInitializationException is quite specific.

Issue #287, opened 2020-07-19
Upgrade to vert.x-sql-client 3.9.2


DavideD:
https://github.com/eclipse-vertx/vertx-sql-client/releases/tag/3.9.2

Can we upgrade or is there any reason to wait?

DavideD:
Upgrading might not be as straightforward as I thought. There is a test that fails for MySQL after the upgrade. I'm investigating.

DavideD:
Done for now, we might have to update something when they've fixed the bug on the vert.x sql client: https://github.com/eclipse-vertx/vertx-sql-client/issues/719

PR #289, opened 2020-07-20
[#287] Upgrade to vert.x-sql-client 3.9.2

DavideD:
Any reason to wait before including this update?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ ext {
     }
     // For ORM, we need a parsed version (to get the family, ...)
     hibernateOrmVersion = Version.parseProjectVersion( project.hibernateOrmVersion )
-    vertxVersion = '3.9.1'
+    vertxVersion = '3.9.2'
     testcontainersVersion = '1.14.3'
     baselineJavaVersion = 1.8
 }


DavideD:
There is a test failure, I will check

gavinking:
> Any reason to wait before including this update?

Not as far as I know.

Sanne:
I'd love to upgrade this ASAP, but I think we'd need to make sure Quarkus won't delay the matching upgrade.

Sanne:
@gsmet / @FroMage / @aguibert  WDYT, can we be optimistic about this upgrade to happen soon in Quarkus?

gsmet:
Funny thing, it was supposed to already be done here: https://github.com/quarkusio/quarkus/pull/10779 but... the BOM wasn't updated.

I pinged Clément. It's definitely something that should land soon in master.

Sanne:
great, thanks @gsmet for the pointer. I see Clement's PR actually had the upgrade, maybe we lost some changes during the BOM moves?

Sanne:
ah, no my bad. The `application/pom.xml` is missing it, Clement had it patched in `build-parent/pom.xml ` .. not sure why this property is duplicated.

gsmet:
That's what I was saying in my message :). Clément is working on it, I pinged him a few minutes ago.

The property is duplicated because properties are not inherited from the bom and the one in build-parent is probably used in the doc. I had plans to create a `quarkus-properties` or `quarkus-versions` parent that could served as a common parent but never got to it. I would need to check with Alexey if it's a good idea anyway.

gsmet:
@Sanne the Vert.x 3.9.2 upgrade has been merged to Quarkus master.

Sanne:
awesome, thanks @gsmet !

Issue #288, opened 2020-07-19
Clean up tests for supported types


DavideD:
While working on #285 I noticed our tests for supported types are all over the place making it hard to test for additional types and figure out which mapping is supported by which db. This was fine when we were still working on the prototype but we now need something more manageable.

I'm also going to fix some bugs I've found along the way.

I'm going to send a PR on Monday.

DavideD:
I've updated a bug in `IndentityGenerationTest` and fixed the error message in `DatabaseSelectionRule`.

DavideD:
Done https://github.com/hibernate/hibernate-reactive/issues/288

PR #290, opened 2020-07-20
#288 Cleanup tests

DavideD:
See #288 

This PR became bigger than I've anticipated and could be split in a couple of separate issues. That said, if reviewed following the commit history, it shouldn't be too bad. Let me know if you prefer separate PRs.

It includes:
* The creation of a rule to select which db use for the tests: [DatabaseSelectionRule](https://github.com/hibernate/hibernate-reactive/blob/f4722575e687ffab8023c0a1cbd06d87fd002711/hibernate-reactive-core/src/test/java/org/hibernate/reactive/testing/DatabaseSelectionRule.java)
* Refactoring of `BasicTypesAndCallbacksTest` into `BasicTypesAndCallbacksForAllDBsTest` and `BasicTypesForSelectedDBTest`
* Test URL field
* Test Calendar field using the different TemporalTypes
* Support for `@Lob lob field` for MySQL (it doesn't work for the other dbs)
* Support UUID field: before it was always null when returning the value
* Deletion of [AutoIncrementTest](https://github.com/hibernate/hibernate-reactive/blob/f4722575e687ffab8023c0a1cbd06d87fd002711/hibernate-reactive-core/src/test/java/org/hibernate/reactive/AutoincrementTest.java)  and replaced it with [IdentityGeneratorTest]( https://github.com/hibernate/hibernate-reactive/blob/bacdff77562eadba60867e371e203e0ba15577a7/hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTest.java): I'm not sure what it was supposed to test but I think it was the id generation
* [AutoIncrementTest](https://github.com/hibernate/hibernate-reactive/blob/f4722575e687ffab8023c0a1cbd06d87fd002711/hibernate-reactive-core/src/test/java/org/hibernate/reactive/AutoincrementTest.java) fails when we upgrade to vert.x sql client 3.9.2. I replaced it with this simpler use case: [SerializableAndManyToOneTest](https://github.com/hibernate/hibernate-reactive/blob/4ff3820c9dd9f55c889901473d015ec2674b025f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/SerializableAndManyToOneTest.java). I don't know why it fails yet but I will fix it when sending the PR for the upgrade to vert.x.

Some of these tests will work for all the dbs once we upgrade to Vert.x 3.9.2.

Cheers


hibernate-reactive-core/src/test/java/org/hibernate/reactive/BinInteger.java
```diff
@@ -1,23 +0,0 @@
...
- * Copyright: Red Hat Inc. and Hibernate Authors
- */
-package org.hibernate.reactive;
-
-import javax.persistence.AttributeConverter;
-import javax.persistence.Converter;
-import java.math.BigInteger;
-
-@Converter
-class BinInteger implements AttributeConverter<BigInteger,String> {
```
gavinking, line 13: What became of the test for JPA converters?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
```diff
@@ -1,359 +0,0 @@
...
-		@Lob @Column(length = 100_000) protected byte[] pic;
-		@Lob @Column(length = 100_000) protected String book;
-
-		@Transient boolean prePersisted;
-		@Transient boolean postPersisted;
-		@Transient boolean preUpdated;
-		@Transient boolean postUpdated;
-		@Transient boolean postRemoved;
-		@Transient boolean preRemoved;
-		@Transient boolean loaded;
```
gavinking, line 267: I don't see what happened to the tests for JPA callbacks?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
```diff
@@ -1,359 +0,0 @@
...
-				.thenCompose(s4 ->
-						s4.find( Basic.class, basik.getId() )
-							.thenAccept(context::assertNull))
-		);
-	}
-
-	enum Cover { hard, soft }
-
-	@Embeddable
-	static class Embed {
```
gavinking, line 170: I don't see the tests for enum and embedded types.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BinInteger.java
DavideD: THey are all in: [BasicTypesAndCallbacksForAllDBsTest](https://github.com/hibernate/hibernate-reactive/blob/abb3ecd984df9847a7b8c52ac897248f24c25039/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesAndCallbacksForAllDBsTest.java)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
DavideD: They are all in: BasicTypesAndCallbacksForAllDBsTest

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
DavideD: They are all in: BasicTypesAndCallbacksForAllDBsTest: https://github.com/hibernate/hibernate-reactive/blob/abb3ecd984df9847a7b8c52ac897248f24c25039/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesAndCallbacksForAllDBsTest.java#L245

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
DavideD: https://github.com/hibernate/hibernate-reactive/blob/abb3ecd984df9847a7b8c52ac897248f24c25039/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesAndCallbacksForAllDBsTest.java#L279

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BinInteger.java
DavideD: https://github.com/hibernate/hibernate-reactive/blob/abb3ecd984df9847a7b8c52ac897248f24c25039/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesAndCallbacksForAllDBsTest.java#L256

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
DavideD: https://github.com/hibernate/hibernate-reactive/blob/abb3ecd984df9847a7b8c52ac897248f24c25039/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesAndCallbacksForAllDBsTest.java#L231

DavideD:
I've updated a bug in IndentityGenerationTest and fixed the error message in DatabaseSelectionRule.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
gavinking: OK sorry, I'm not sure how I missed that file. Probably because was trying to use the github iPad client and not used to it.

gavinking:
A question and a comment:

- what is the motivation for mapping `Long` using `@Lob`? Does that even make sense?
- I guess it's fine to let people map `Calendar` for backward compatibility but let's be very clear we don't encourage this! It's already documented that we don't let you pass `TemporalType` with a query parameter, and by extension one simply shouldn't be using all this nasty old deprecated stuff.

DavideD:
> what is the motivation for mapping Long using @Lob? Does that even make sense?

Because I can :)
Joking aside, I copied the test from OGM thinking that a user requested it. Now that I went to look at the issue it was actually something that we added because per specification you can apply a Lob to any field:

> A Lob annotation specifies that a persistent property or field should be persisted as a large object to a
database-supported large object type.
...
The Lob annotation may be used in conjunction with the Basic annotation or with the ElementCollection

I wonder if the ElementCollection case would work.
In the end I've implemented a couple of methods in the ResultSetAdaptor that maybe will help with the mapping of other fields,
so I don't mind to keep it.

> I guess it's fine to let people map Calendar for backward compatibility but let's be very clear we don't encourage this! 

Hey, I didn't write the JPA spec :)
I don't see tests as encouragement to anything, when we have the documentation we can add a blinking big red box.
Or I can add a comment in the test.

I think now we cover a good set of types and I don't plan to add any additional weird mapping.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BasicTypesAndCallbacksTest.java
DavideD: No problem, normally I would keep the history about renaming the file but I think this time I messed up somewhere and didn't want to spend more time to rebase everything.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -633,7 +641,10 @@ public Ref getRef(String columnLabel) {
 
 	@Override
 	public Blob getBlob(String columnLabel) {
-		throw new UnsupportedOperationException();
+		Buffer buffer = (Buffer) row.getValue( columnLabel );
+		return ( wasNull = buffer == null )
+				? null
+				: BlobTypeDescriptor.INSTANCE.wrap( buffer.getBytes(), options );
```
gavinking, line 647: @DavideD so this looks like a layer-breaker to me.

I think this should just be a call directly to `BlobProxy.generateProxy()`, and then we can eliminate the passing of `WrapperOptions`.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTest.java
```diff
@@ -0,0 +1,92 @@
...
+		configuration.addAnnotatedClass( EntityWithIdentity.class );
+		return configuration;
+	}
+
+	private CompletionStage<?> populateDb() {
+		final Object[] identities = new EntityWithIdentity[ENTITY_NUMBER];
+		for ( int i = 0; i < ENTITY_NUMBER; i++ ) {
+			identities[i] = new EntityWithIdentity( i );
+		}
+		return getSessionFactory().withTransaction( (session, tx) -> session.persist( identities ) );
```
gavinking, line 38: @DavideD This test seems inadequate. We don't check that the generated identities are actually populated back into the entities. Remember that with autoincrement/identity columns, the whole lifecycle of the `persist()` operation is pretty different, so we need to properly test it.

DavideD: I see, ok

DavideD: Do you think this is enough? https://github.com/hibernate/hibernate-reactive/pull/290/commits/47863a684fae83686672710204fc138a3ef53c03

gavinking: Shouldn’t we be checking that the generated value is greater than 0?

DavideD: Done: https://github.com/hibernate/hibernate-reactive/pull/290/commits/edac32102af36b75bca3af61b281b08348558e3f#diff-bf65d08103f4e4a90f49edd42293fb94R43

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: I've updated the commit: https://github.com/hibernate/hibernate-reactive/pull/290/commits/08fe85f26d05cc639e865ed09fab9953fded0acc

gavinking: Thanks.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTest.java
gavinking: cool, thanks

DavideD:
I'm removing the test for vertx sql client 3.9.2 and send a separate PR for it: https://github.com/hibernate/hibernate-reactive/pull/290/commits/b8c1cbdba72c013065717c87f93fa6518ecda7f3

DavideD:
@gavinking Anything else? I would like to merge this

gavinking:
sure, go ahead!

DavideD:
Done https://github.com/hibernate/hibernate-reactive/pull/290

Thanks for the feedback!

Issue #291, opened 2020-07-22
Upgrade Mutiny to 0.7.0


DavideD:
The same version used in quarkus: https://github.com/quarkusio/quarkus/blob/master/bom/application/pom.xml#L131

DavideD:
Done https://github.com/hibernate/hibernate-reactive/issues/291

PR #292, opened 2020-07-22
291 mutiny upgrade

DavideD:
See #291


DavideD:
Done https://github.com/hibernate/hibernate-reactive/issues/291

PR #293, opened 2020-07-23
287 Upgrade to vert.x sql client 3.9.2

DavideD:
I think I've found a bug in the vertx client for MySQL.
Disabling the query cache solves the issue.
We have a query that has a Buffer as parameter. If the first time we run the query the parameter is null, the client will infer the type in `DataTypeCodec#inferDataTypeByEncodingValue` as `DataType.NULL` instead of `DataType.BLOB`. It seems that at that point the value is cached and therefore the encoding changes depending on when you run the query.

For now I've disabled the cache for MySQL so that we can upgrade.

@aguibert Would you be able to have a look at this, please?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,9 @@ Hibernate Reactive has been tested with:
 - MySQL 8
 - DB2 11.5
 - [Hibernate ORM](https://hibernate.org/orm/) 5.4.17.Final
-- [Vert.x Reactive PostgreSQL Client](https://vertx.io/docs/vertx-pg-client/java/) 3.9.1
-- [Vert.x Reactive MySQL Client](https://vertx.io/docs/vertx-mysql-client/java/) 3.9.1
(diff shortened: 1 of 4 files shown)


DavideD:
When https://github.com/hibernate/hibernate-reactive/pull/290 is merged, I'm going to enable the tests that will work with this version of the client.

gavinking:
I guess it's not surprising that there's a regression, since I know a bunch of work was done in that area in this release (partly as a result of me reporting a couple of bugs).

DavideD:
I've created this issue on the vert.x sql client: https://github.com/eclipse-vertx/vertx-sql-client/issues/719

DavideD:
I'm merging this one. It's pretty simple and we can re-enable the cache for MySQL when they've fixed the bug

DavideD:
Done https://github.com/hibernate/hibernate-reactive/pull/293

Issue #270, opened 2020-06-30
user manual


gavinking:
We need to fill out the stub asciidoc. At first this is going to be pretty much just a copy/paste of the README, which is why i’ve been holding off on starting work on it (didn’t want to have to maintain the same info in two places).

But I’ve now started work on it.

PR #300, opened 2020-07-29
first stab at reference documentation

gavinking:
For #270.

### DIFF
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -3,13 +3,17 @@
 
 :example: https://github.com/hibernate/hibernate-reactive/tree/master/example
 
-Before you start, we recommend taking a quick look at the example
-program in the {example}[`example`] directory, which shows off all
-the "bits" you'll need to get your own program up and running.
+Creating a new project with Hibernate Reactive isn't hard at all. In this
+short guide, we'll cover all the basic work involved in:
+
+- setting up and configuring a project, and then
+- writing Java code to define a data model and access the database.
+
+But, before you start, we recommend taking a quick look at the example
+program in the {example}[`example`] directory, which shows off all the
+"bits" you'll need to get your own program up and running.
 
 == Setting up a reactive Hibernate project
-
-Setting up a project that uses Hibernate Reactive isn't hard at all.
 
 If you're using Hibernate Reactive outside of the Quarkus environment,
 you'll need to:
@@ -105,7 +109,7 @@ To see the generated SQL as it's sent to the database, either:
 
 - set the property `hibernate.show_sql` to `true`, or
 - enable debug-level logging for the category `org.hibernate.SQL`
-  in for your preferred SLF4J logging implementation.
+  using your preferred SLF4J logging implementation.
 
 For example, if you're using Log4J via `slf4j-log4j12` (as above in
 <<_optional_dependencies>>), add this line to your `log4j.properties`
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -1,5 +1,5 @@
 [[getting-started]]
-= Getting started
+= Introduction to Hibernate Reactive
 
 :example: https://github.com/hibernate/hibernate-reactive/tree/master/example
 
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -260,6 +260,18 @@ them is a matter of deciding which reactive API you want to use for working
 with reactive streams. Your decision won't affect what you can do with
 Hibernate Reactive.
 
+In this introduction, our code examples usually use `CompletionStage`, since
+that's what more people are already familiar with. But please don't take that
+as a recommendation.
+
+When we use the term _reactive stream_ in this document, we mean:
+
+- a chain of ``CompletionStage``s, or
+- a chain of Mutiny ``Uni``s and ``Multi``s
+
+that is built by the program in order to service a particular request,
+transaction, or unit of work.
+
 === Obtaining a reactive session factory
 
 Whatever you decide, the first step to getting a reactive session is to obtain
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -24,7 +24,6 @@ Or, if you're using Hibernate Reactive in Quarkus.... TODO
 === Including Hibernate Reactive in your project build
 
 :build: https://github.com/hibernate/hibernate-reactive/blob/master/example/build.gradle
-:slf4j: http://www.slf4j.org/
 
 Add the following dependency to your project:
 
@@ -46,6 +45,9 @@ can if you like).
 
 === Optional dependencies
 
+:slf4j: http://www.slf4j.org/
+:enhancer: https://docs.jboss.org/hibernate/orm/5.4/topical/html_single/bytecode/BytecodeEnhancement.html
+
 Optionally, you might also add:
 
 - an {slf4j}[SLF4J] logging implementation, for example,
@@ -56,8 +58,8 @@ Optionally, you might also add:
   `org.glassfish:jakarta.el` if you're using Hibernate Validator,
 - `org.hibernate:query-validator` to enable compile-time checking of your
   HQL queries, and/or
-- the Hibernate bytecode enhancer if you want to use field-level lazy
-  fetching.
+- the Hibernate {enhancer}[bytecode enhancer] if you want to use field-level
+  lazy fetching.
 
 TIP: Field-level lazy fetching is an advanced feature that most programs
 don't need. Stick to the basics for now.
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -15,6 +15,8 @@ program in the {example}[`example`] directory, which shows off all the
 
 == Setting up a reactive Hibernate project
 
+:skeleton: https://code.quarkus.io/?g=org.acme&a=code-with-quarkus&v=1.0.0-SNAPSHOT&b=MAVEN&c=org.acme.ExampleResource&s=r1s.8XW.fmW.ih0&cn=code.quarkus.io
+
 If you're using Hibernate Reactive outside of the Quarkus environment,
 you'll need to:
 
@@ -23,7 +25,8 @@ you'll need to:
 - configure Hibernate Reactive with information about your database,
   using Hibernate configuration properties.
 
-Or, if you're using Hibernate Reactive in Quarkus.... TODO
+Or, if you want to use Hibernate Reactive in Quarkus, you can generate
+a preconfigured skeleton project {skeleton}[right here].
 
 === Including Hibernate Reactive in your project build
 
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -28,7 +28,9 @@ Or, if you're using Hibernate Reactive in Quarkus.... TODO
 
 Add the following dependency to your project:
 
-    org.hibernate.reactive:hibernate-reactive-core:{version}
+----
+org.hibernate.reactive:hibernate-reactive-core:{version}
+----
 
 Where `{version}` is the version of Hibernate Reactive you're using.
 
@@ -72,7 +74,10 @@ document which must be placed, as usual, in the `/META-INF` directory.
 The only configuration specific to Hibernate Reactive is the persistence
 `<provider>` element, which must be explicit:
 
-    <provider>org.hibernate.reactive.provider.ReactivePersistenceProvider</provider>
+[source,xml]
+----
+<provider>org.hibernate.reactive.provider.ReactivePersistenceProvider</provider>
+----
 
 Otherwise, configuration is almost completely transparent. Configure
 Hibernate exactly as you usually would, noting that most configuration
@@ -92,17 +97,25 @@ program.
 
 === Logging the generated SQL
 
+:log4j: https://github.com/hibernate/hibernate-reactive/blob/master/example/src/main/resources/log4j.properties
+
 To see the generated SQL as it's sent to the database, either:
 
 - set the property `hibernate.show_sql` to `true`, or
 - enable debug-level logging for the category `org.hibernate.SQL`
   in for your preferred SLF4J logging implementation.
 
-For example, if you're using Log4J via `slf4j-log4j12`, add this line:
+For example, if you're using Log4J via `slf4j-log4j12` (as above in
+<<_optional_dependencies>>), add this line to your `log4j.properties`
+file:
 
-    log4j.logger.org.hibernate.SQL=DEBUG
+[source,properties]
+----
+log4j.logger.org.hibernate.SQL=DEBUG
+----
 
-to your `log4j.properties` file.
+An example {log4j}[`log4j.properties`] file is included in the example
+program.
 
 === Tuning the Vert.x pool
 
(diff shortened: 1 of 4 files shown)


gavinking:
Big question I can't seem to find the answer to: how the hell can I have a fragment of Java source code with highlighting in asciidoc?!

gavinking:
Oh and another question: what's the right way to change the `asciidoctor.css` file? That's one of the ugliest monospaced fonts I've ever seen.

DavideD:
> Oh and another question: what's the right way to change the `asciidoctor.css` file? That's one of the ugliest monospaced fonts I've ever seen.

Have  you already checked ORM? https://github.com/hibernate/hibernate-orm/blob/master/documentation/documentation.gradle#L229

gavinking:
@FroMage I'm going to need at least some sort of text in here talking about HR in Panache and Quarkus, even if it's just something short drawing attention to differences, and linking to other documentation. WDYT, could you help me out on that front?

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+If you're using Hibernate Reactive outside of the Quarkus environment,
+you'll need to:
+
+- include Hibernate Reactive itself along with the appropriate Vert.x
+  reactive database client as dependencies of your project, and
+- configure Hibernate Reactive with information about your database,
+  using Hibernate configuration properties.
+
+Or, if you're using Hibernate Reactive in Quarkus.... TODO
```
gavinking, line 22: Folks, I could use some suggestions here!

DavideD: For quarkus one needs to use the proper extension. You can check this quickstart: https://github.com/quarkusio/quarkus-quickstarts/blob/master/hibernate-reactive-quickstart/pom.xml#L35

Maybe we can link to the quikcstart here.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+You'll also need to add a dependency for the Vert.x reactive database
+driver for your database, for example:
+
+- `io.vertx:vertx-pg-client` for Postgres,
+- `io.vertx:vertx-mysql-client` for MySQL, or
+- `io.vertx:vertx-db2-client` for DB2.
+
+You don't need to depend on the JDBC driver for your database (but you
+can if you like).
```
DavideD, line 45: This sentence sounds weird (why somebody would add it if it's not needed?), maybe you can clarify what's the difference with having the JDBC driver on the classpath.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+- an {slf4j}[SLF4J] logging implementation, for example,
+  `org.slf4j:slf4j-log4j12`,
+- `org.hibernate:hibernate-jpamodelgen` if you're using the JPA criteria
+  query API,
+- `org.hibernate.validator:hibernate-validator` and
+  `org.glassfish:jakarta.el` if you're using Hibernate Validator,
+- `org.hibernate:query-validator` to enable compile-time checking of your
+  HQL queries, and/or
+- the Hibernate bytecode enhancer if you want to use field-level lazy
+  fetching.
```
DavideD, line 60: Is the bytecode enhancer a separate library?

DavideD: A link to the ORM documentation might be helpful

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+<<_optional_dependencies>>), add this line to your `log4j.properties`
+file:
+
+[source,properties]
+----
+log4j.logger.org.hibernate.SQL=DEBUG
+----
+
+An example {log4j}[`log4j.properties`] file is included in the example
+program.
```
DavideD, line 118: The configuration of the logger for Quarkus is a bit different  because it stays in the application.properties. Should we mention it? (No strong opinion about it)

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+=== Tuning the Vert.x pool
+
+The Vert.x database client has built-in connection pooling and prepared
+statement caching.
```
DavideD, line 123: Caching is disabled for DB2 and MySQL

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+  connection pool.
+
+When it comes time for performance tuning, you can further customize the
+pool and cache via the following properties:
+
+- `hibernate.vertx.pool.max_wait_queue_size`
+- `hibernate.vertx.prepared_statement_cache.max_size`
+- `hibernate.vertx.prepared_statement_cache.sql_limit`
+
+TIP: But for now, just leave these three settings alone.
```
DavideD, line 135: I don't understand this TIP. Why mention them and then say that they shouldn't be set without any explanation?

DavideD: Is it because it is an introduction? If that's the case can you mention this property in a separate section and add a link here to it?

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+and execution of SQL against the database is never performed synchronously.
+Persistence operations that belong to a single unit of work must be chained by
+composition within a single reactive stream.
+
+Also remember that a Hibernate session is a lightweight object that should be
+created, used, and then discarded within a single logical unit of work.
+
+TIP: That is to say, you should reuse the same session across multiple
+persistence operations within a single reactive stream representing a certain
+unit of work, but don't share a session between different reactive streams!
```
DavideD, line 308: This is a nice tip but I'm not sure what you mean when you talk about reactive streams

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+[java]
+----
+sessionFactory.withSession(
+        session -> ... //do some work
+);
+----
+
+Alternatively, you may use `openSession()`, but you must remember to `close()`
+the session when you're done.
```
DavideD, line 320: I think it's worth mentioning that `flush()` operations aren't done by default. Should we mention `withTransaction()`?

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Well the information there doesn't really help someone who doesn't already know a lot about Quarkus know what they have to do to use HR in Quarkus. I want a brief outline of the basic steps.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+Now, _here's where Hibernate Reactive is different:_ in the reactive API, each
+of these methods returns its result in a non-blocking fashion via a Java
+`CompletionStage` (or Mutiny `Uni`). For example:
+
+[java]
+----
+session1.find(Book.class, book.id)
+        .thenAccept( book -> System.out.println(book.title + " is a great book!") )
+----
```
DavideD, line 363: Maybe we should mention that from now on the examples will show the `Stage.Session` API.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: I've not yet added any crosslinks to the Hibernate docs because I'm not quite sure of the right way to do it. But sure, I guess I can add an absolute link in here for now, and we can fix it later.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Depends how much detail we're going to get into regarding Quarkus in this guide. So far I don't have a clear view. Depends if we're going to have a separate document (or separate chapter of this document) for Quarkus.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Yes but that's a bug we're in the process of fixing. We' don't need to mention bugs in the user guide.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Just because a brand new user doesn't need to actually *make use* of a feature right now, doesn't mean that they don't need to know that *it exists*, so that they will know what to reach for when the time comes.

Also, we don't currently have a reference doc, and I have absolutely zero intention of writing and maintaining one at this stage, so this introduction needs to serve a double purpose for now.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+Before you start, we recommend taking a quick look at the example
+program in the {example}[`example`] directory, which shows off all
+the "bits" you'll need to get your own program up and running.
+
+== Setting up a reactive Hibernate project
+
+Setting up a project that uses Hibernate Reactive isn't hard at all.
+
+If you're using Hibernate Reactive outside of the Quarkus environment,
+you'll need to:
```
DavideD, line 15: Do we need to mention Quarkus here? Can't we just say something like:
``
You need these dependencies ... or, if you are using Quarkus, these extensions...
```

File: documentation/src/main/asciidoc/reference/introduction.adoc (hunk 7 above)
Comment (gavinking) [reply documentation/src/main/asciidoc/reference/introduction.adoc:310]: Yes well I agree with that I'm having trouble coming up with quite the right language for talking about streams.

Here, a reactive stream is a chain of `CompletionStage`s or `Uni`s. Perhaps I just need to define that language somewhere.

File: documentation/src/main/asciidoc/reference/introduction.adoc [hunk 11]
```diff
@@ -0,0 +1,517 @@
...
+
+:example: https://github.com/hibernate/hibernate-reactive/tree/master/example
+
+Before you start, we recommend taking a quick look at the example
+program in the {example}[`example`] directory, which shows off all
+the "bits" you'll need to get your own program up and running.
+
+== Setting up a reactive Hibernate project
+
+Setting up a project that uses Hibernate Reactive isn't hard at all.
```
DavideD, line 12: `s/isn't hard at all/is [easy|straightforward]/`

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: I do so in the very next section.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -0,0 +1,517 @@
...
+
+If you're using Hibernate Reactive outside of the Quarkus environment,
+you'll need to:
+
+- include Hibernate Reactive itself along with the appropriate Vert.x
+  reactive database client as dependencies of your project, and
+- configure Hibernate Reactive with information about your database,
+  using Hibernate configuration properties.
```
DavideD, line 20: Maybe we can add the GAV coordinates of the dependencies for reactive and the drivers?

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Meh. I guess.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Because this is  an introduction, I'm fine with removing this section from here

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Yes, but this is an introduction, I would just add the extension needed and mention that they exist plus the properties needed to set in the application.properties. I don't think there is much more to do in term of configuration.
I think I would leave a complete guide to configure a project with Quarkus to a separate section.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I would leave it in a separate chapter

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Can we just remove the TIP then? I think this sentence explain it clearly:
```
When it comes time for performance tuning, you can further customize the
pool and cache via the following properties:
```

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Or we can just remove this sentence

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I've just realized that they are mentioned in the later section

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Pah. Boring and pedestrian.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Sure, if you have the list of which extensions are needed...

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Yes, and note I don't want to add version numbers to this document, because that would be a maintenance nightmare.

Remember, @DavideD, well-written technical docs should always have handwavey preambles at the beginning of sections, before getting down to nitty gritty technical details in titled subsections.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Well, that's almost the exact opposite of the information I would like to have in this preamble. I don't want specific details of this kind, I want a broad-brush picture of the *process* I need to follow.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Done.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: > Perhaps I just need to define that language somewhere.

I tried to, though I wonder if reactive purists will think I'm using words wrong.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: For Hibernate Reactive is this one:
```
<dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-hibernate-reactive</artifactId>
        </dependency>
```
plus, the connector for the db:
```
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-reactive-pg-client</artifactId>
        </dependency>
```

There is probably one for mysql and DB2 but I haven't checked

DavideD: ALso the deployment:
```
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-hibernate-reactive-deployment</artifactId>
            <scope>provided</scope>
            <version>${quarkus-plugin.version}</version>
        </dependency>
```

DavideD: I just remembered that we could link to the `code.quarkus.io` for the generation of a project with reactive: https://code.quarkus.io/?g=org.acme&a=code-with-quarkus&v=1.0.0-SNAPSHOT&b=MAVEN&c=org.acme.ExampleResource&s=r1s.8XW.fmW.ih0&cn=code.quarkus.io

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Yes, I think that would be better. Talking about Quarkus here would disrupt the flow.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Ahah, that looks good. Thanks.

gavinking: That's good enough for now, though we might still want to add a Quarkus/Panache chapter later.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I mean, for quarkus the steps are the same: you need to add the dependencies (Hibernate reactive and driver) and configure the application. The only difference is that you use different coordinates. I'm not sure how this is helpful without the exacts coordinates but I'm fine with it.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Looks good to me for now. Stream is a tricky word to use in this context but maybe we will find something better later.

gavinking:
@FroMage is on PTO so I might just merge the work I've already done, and come back to the question of Panche later.

DavideD:
Sounds good to me

Issue #279, opened 2020-07-09
Support for custom UserTypes


akoufa:
Hello to all. Does Hibernate Reactive support custom UserTypes? I want to use Postgres JSONB Type and for that I was using the following library https://github.com/vladmihalcea/hibernate-types which builds it's custom types based on Hibernate's[AbstractSingleColumnStandardBasicType](https://docs.jboss.org/hibernate/orm/5.4/javadocs/org/hibernate/type/AbstractSingleColumnStandardBasicType.html). Unfortunately just moving from Hibernate to Hibernate Reactive the columns that use these types are always null in the entities. As an alternative I found this [article](https://thorben-janssen.com/persist-postgresqls-jsonb-data-type-hibernate/) which builds it's JSONB Postgres support on the Hibernate's [UserType](https://docs.jboss.org/hibernate/orm/5.4/javadocs/org/hibernate/usertype/UserType.html). Is there any known limitation around defining custom types and using Hibernate Reactive at the moment ?

gavinking:
Hi, I didn’t test `UserType` since that interface is rather tied to JDBC. (Though in principle, perhaps it could be made to mostly work.)

Instead, my idea was that people could use JPA converters. I did test that and AFAIK it works. Have you tried that?

DavideD:
I tried with a quick test using UserType and it doesn't seem to work because the PgClient throw an exception when running the query.

gavinking:
> I tried with a quick test using UserType and it doesn't seem to work because the PgClient throw an exception when running the query.  

But that might just be because our fake JDBC layer is incomplete. (Some unimplemented methods.)

DavideD:
To clarify, I tries the example using vlad library:
```
	@TypeDefs({
			@TypeDef(name = "json", typeClass = JsonStringType.class),
			@TypeDef(name = "jsonb", typeClass = JsonBinaryType.class)
	})
	@Entity(name="GuineaPig")
	@Table(name="Pig")
	public static class GuineaPig {
		@Id
		private Integer id;
		private String name;

		@Type(type = "jsonb")
		@Column(columnDefinition = "jsonb")
		private  Location location;
```
This is the error:
```
Caused by: io.vertx.core.impl.NoStackTraceThrowable: Parameter at position[0] with class = [com.fasterxml.jackson.databind.node.ObjectNode] and value = [{"country":"UK","city":"Gotham"}] can not be coerced to the expected class = [java.lang.Object] for encoding.
```

[...]

PR #301, opened 2020-07-30
add test for UserType mappings

gavinking:
see #279


DavideD:
Would it be possible to have something like:
```
@Type(type="passthrough")
Circle circle
```
That doesn't do any conversion but just pass the value as is?

Issue #284, opened 2020-07-16
[Question] Mutiny.Session#createNativeQuery(String) returns Query<Integer>


DavideD:
https://github.com/hibernate/hibernate-reactive/blob/master/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java#L625

It seems on purpose (reading javadoc) but this surprised me:
* `Mutiny.Session#createQuery(String)` doesn't
* we return the integer when running `.executeUpdate()`

Shouldn't this return `<R> Query<R>` like all the other similar functions?
@gavinking (\cc @Sanne, @gbadner)

gavinking:
Definitely intentional. If you don't pass anything describing the type of the result set for a native query, there's no way for Hibernate to know what to do with it.

HQL is different because we can figure out the return type from the query string itself. We can't do that with SQL.

DavideD:
Doesn't ORM returns an Object[]? I was expecting to receive whatever object the driver returns.

DavideD:
> I was expecting to receive whatever object the driver returns.

I mean... for each column. Also, isn't `createNativeQuery` quite misleading as a name when everywhere else we return something different?

gavinking:
> Doesn't ORM return an Object[]? I was expecting to receive whatever object the driver returns.

I don't recall. But if it does, it would need to use JDBC metadata for that, which we don't have yet.

Hrm, I seem to recall I might have opened a feature request against the Vert.x client related to this, but I forget the details.

gavinking:
See my comments [here](https://github.com/hibernate/hibernate-reactive/pull/180#discussion_r429576118).

DavideD:
I'm missing how the column type is needed in this case.

I can actually write:
```
List<?> results = mutinySession.createNativeQuery( "select * from artist a" )
                .getResultList().await().indefinitely();
```
and it returns what I expected (List<Object[]>).

This also compile
```
List<Integer> result = mutinySession.createNativeQuery(" select * from artist a" )
                .getResultList().await().indefinitely();
```

This doesn't compile, though:
```
List<Object[]> results = mutinySession.createNativeQuery(  "select * from artist a" )
                .getResultList().await().indefinitely();
```

gavinking:
TBH I really don't recall all the context. But anyway if that code works then I guess you're right and we should change the return type.

DavideD:
Done https://github.com/hibernate/hibernate-reactive/issues/284

PR #303, opened 2020-07-31
allow createNativeQuery(String) to be used with SQL selects

gavinking:
I made a mistake and thought that this only made sense for DML.

@DavideD set me straight, see #284


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/303

Nice!

PR #308, opened 2020-08-03
Use lambda for remove.

alesj:


gavinking:
@alesj I was actually consistently not using method refs for session operations in the example app.

alesj:
Ah, OK. This was the only thing IDEA suggested, so I thought it was an oversight.

gavinking:
It's not that I have anything against function refs, just that in this demo code I thought it was a bit clearer to make the argument explicit to show what type it is.

Issue #258, opened 2020-06-15
StatelessSession


gavinking:
We could add `Stage.StatelessSession` and `Mutiny.StatelessSession`.

gavinking:
I've begun work on this.

gavinking:
If anyone's interested, the proof of concept is taking form in the ss branch of my repo, though it's still pretty rough for now.

gavinking:
This is now mostly-done, it seems to me, but I need to write tests and Javadoc.

gavinking:
> I need to write tests and Javadoc

Done, and I'm just waiting for https://github.com/hibernate/hibernate-orm/pull/3494 to be merged and released.

gavinking:
Note that #312 introduces support for batching, and there is a little fix needed to `StatelessSession` after that is merged.

gavinking:
Done!

PR #310, opened 2020-08-04
StatelessSession

gavinking:
Add a `StatelessSession` API as proposed in #258.


gavinking:
Note that this should not be merged yet, since it depends on a tiny change to hibernate-core.

gavinking:
@Sanne @DavideD can you take a look over this and let me know what you think?

Sanne:
I'll try to have a look, Davide is away.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -537,6 +537,37 @@ Note that these are "resource local" transactions, delegated to the underlying
...
+
+An arguably-underappreciated feature of Hibernate is the `StatelessSession`
+interface, which provides a command-oriented, more bare-metal approach to
+interacting with the database.
+
+You may obtain a reactive stateless session from the `SessionFactory`:
+
+[java]
+----
+Stage.StatelessSession ss = getSessionFactory().createStatelessSession();
```
Sanne, line 550: I think I'd prefer the `openStatelessSession()` name, as in the original project, as it reminds it's a resource which will need closing to e.g. release connections or other resources (it's not _really_ stateless, is it?)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -871,6 +872,177 @@
...
+	 * <p/>
+	 * A stateless session does not implement a first-level cache nor interact
+	 * with any second-level cache, nor does it implement transactional
+	 * write-behind or automatic dirty checking, nor do operations cascade to
+	 * associated instances. Collections are ignored by a stateless session.
+	 * Operations performed via a stateless session bypass Hibernate's event
+	 * model and interceptors. Stateless sessions are vulnerable to data
+	 * aliasing effects, due to the lack of a first-level cache.
+	 * <p/>
+	 * For certain kinds of transactions, a stateless session may perform slightly
```
Sanne, line 888: I understand what you mean, but perhaps I'd say "operations" rather than "transactions" ?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -57,6 +58,32 @@ public MutinySessionFactoryImpl(SessionFactoryImpl delegate) {
 				.map( MutinySessionImpl::new );
 	}
 
+	@Override
+	public Mutiny.StatelessSession createStatelessSession() {
```
Sanne, line 62: I see this isn't newly introduced on the `StatelessSession` as you have it on `Session` too, but I don't fully understand why we need two different APIs to open/create a XSession.

Shouldn't all connection acquisition be lazy, or left as a configurable policy as in Hibernate ORM? The current policy contract evolved from integration needs with various containers and datasource pools. In Quarkus specifically, we're working to have it eventually be able to use a strategy of "acquire deferred + release just-before-commit" which is the most efficient strategy.

I'm not suggesting we implement the same policy now in Reactive, but I'd prefer the API we choose to expose will leave us some margin of flexibility to improve on this aspect as necessary.

I'll open a separate issue to discuss this.

Sanne: link to #316

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+import io.smallrye.mutiny.Uni;
+import org.hibernate.LockMode;
+import org.hibernate.reactive.common.ResultSetMapping;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.session.ReactiveStatelessSession;
+
+/**
+ * Implements the {@link Mutiny.StatelessSession} API. This delegating
+ * class is needed to avoid name clashes when implementing both
```
Sanne, line 16: Is that the only reason? Couldn't use fully qualified names?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -0,0 +1,418 @@
...
+import javax.persistence.Tuple;
+import java.io.Serializable;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * An {@link ReactiveStatelessSession} implemented by extension of
+ * the {@link StatelessSessionImpl} in Hibernate core. Extension was
+ * preferred to delegation because there are places where
+ * Hibernate core compares the identity of session instances.
```
Sanne, line 53: I don't understand the concern regarding identity of session instances. Surely you don't expect people to use both reactive and non-reactive sessions within the same SessionFactory instance?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -0,0 +1,418 @@
...
+                    getPersistenceContext().clear();
+                    afterOperation(x == null);
+                })
+                //TODO: this typecast is rubbish
+                .thenApply( list -> (List<T>) list );
+    }
+
+    @Override
+    public CompletionStage<Integer> executeReactiveUpdate(String expandedQuery, QueryParameters parameters) {
+        throw new UnsupportedOperationException();
```
Sanne, line 389: Are these going to be implemented later, or an implicit limitation of this session kind?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
gavinking: For *method names*.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
gavinking: No, it's nothing to do with that.

And actually TBH I'm not sure if this comment really applies to `StatelessSession`s (though it might). It's a copy/paste from the comment in `ReactiveSessionImpl`. The issue is that Hibernate core sometimes does `if (session==otherSession)` which caused problems when we had  `ReactiveSessionImpl` delegating to `SessionImpl`.

Here I've just followed the design that worked for `ReactiveSessionImpl`.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
gavinking: Pretty sure it's I just `cos I didn't implement them yet.

(This is for native and HQL `update` and `delete` queries, FTR.)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
gavinking: I've responded in the issue.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
gavinking: I'm really talking about a unit of work, not a single operation.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
Sanne: ah, ok!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
Sanne: ok, feel free to keep it. Maybe adjust the comment if you're not sure?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
Sanne: yes I understand. I suspect most users will read "transactions" as literal transactions, and might think a stateless session won't help for non-transactionl operations (while we actually mean to suggest that it could be helpful either way)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
gavinking: I've implemented these methods.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
gavinking: I've improved the Javadoc.

gavinking:
After some finishing-up work, I'm happy with this now, and so I've merged it.

PR #312, opened 2020-08-08
Batching

### LINKED ISSUE #257: batch inserts (and updates)
The Vert.x client does have an API for batching inserts (and updates?) but currently we have commented out the code which handles batching in Hibernate (which was very JDBC-specific).

We should add support for at least batch inserts, since that can make a big difference to performance.

gavinking:
Fixes #257.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,9 +93,11 @@ Configuration properties of particular interest include:
 
 - `javax.persistence.jdbc.url`, the JDBC URL of your database,
 - `javax.persistence.jdbc.user` and `javax.persistence.jdbc.password`,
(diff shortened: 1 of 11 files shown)


gavinking:
Done. I want to get this merged quickly.

gavinking:
> Done. I want to get this merged quickly.

If nobody has any specific feedback, I think it makes sense for me to just go ahead and merge this.

(Note that batching is not enabled by default, and in that case the `BatchingConnection` won't even be in the code path.)

Sanne:
Only noticing this now... Awesome, thanks @gavinking !

PR #317, opened 2020-08-11
much better implementation of ProxyConnection

gavinking:
WDYT, @Sanne ?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
@@ -266,11 +266,21 @@ public class SqlClientPool implements ReactiveConnectionPool, ServiceRegistryAwa
 
 	private class ProxyConnection implements ReactiveConnection {
 		private ReactiveConnection connection;
+		private boolean connected;
 
 		private <T> CompletionStage<T> withConnection(Function<ReactiveConnection,CompletionStage<T>> operation) {
-			return connection == null ?
-					getConnection().thenApply(newConnection -> connection = newConnection).thenCompose(operation) :
-					operation.apply(connection);
+			if (!connected) {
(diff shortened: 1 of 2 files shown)


Sanne:
nice, very readable.
Is there a real possibility for that `IllegalStateException` ?

gavinking:
> Is there a real possibility for that `IllegalStateException` ?

I believe so, if you were to access the session in an illegal way.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -266,81 +266,91 @@ public static URI parse(String url) {
...
-		CompletionStage<ReactiveConnection> connection() {
-			return connection == null ?
-					getConnection().thenApply(conn -> connection = conn) :
-					CompletionStages.completedFuture(connection);
+		private boolean connected;
+
+		private <T> CompletionStage<T> withConnection(Function<ReactiveConnection,CompletionStage<T>> operation) {
+			if (!connected) {
+				connected = true;
+				return getConnection()
```
DavideD, line 274: Shouldn't you set `connected` to true only after you set connection (o in the same CompletionStage)? Could it cause some issues if `getConnection()` takes a long time before assigning the `connection`?

gavinking: The whole point of having a separate flag there is so that I can detect an illegal "concurrent" request for a connection, and throw an exception instead of obtaining a second connection.

DavideD: ok

gavinking:
I've added documentation explaining the issue with concurrent access to a `ProxyConnection`.

I'm going to merge this, because I think it's a clearly better implementation.

Issue #307, opened 2020-08-03
enable prepared statement cache after vert.x bug fix


gavinking:
Currently the prepared statement cache is disabled on DB2 and MySQL, due to bugs in the client.

See here:

<https://github.com/hibernate/hibernate-reactive/blob/d8743ddf21a0cc97c627e9245ed04db89e7a857b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java#L180>

gavinking:
For reference:

- https://github.com/eclipse-vertx/vertx-sql-client/issues/671
- https://github.com/eclipse-vertx/vertx-sql-client/issues/719

gavinking:
This is already fixed in #319, and we're just waiting for a release of Vert.x to merge.

PR #319, opened 2020-08-13
test out fixes to PreparedStatementCache 

gavinking:
See #307.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ ext {
     }
     // For ORM, we need a parsed version (to get the family, ...)
     hibernateOrmVersion = Version.parseProjectVersion( project.hibernateOrmVersion )
-    vertxVersion = '3.9.2'
+    vertxVersion = '3.9.3-SNAPSHOT'
     testcontainersVersion = '1.14.3'
     baselineJavaVersion = 1.8
 }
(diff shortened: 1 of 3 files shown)


gavinking:
Looks like we have a new issue on DB2 with Vert.x 3.9.3:

```
  Caused by:
        java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;
            at io.vertx.db2client.impl.drda.DRDAConnectRequest.getProductData(DRDAConnectRequest.java:121)
            at io.vertx.db2client.impl.drda.DRDAConnectRequest.buildPRDDTA(DRDAConnectRequest.java:363)
            at io.vertx.db2client.impl.drda.DRDAConnectRequest.buildACCRDB(DRDAConnectRequest.java:68)
```

gavinking:
Problem appears to be specific to JDK 1.8. It works for me on JDK 11.

gavinking:
> Problem appears to be specific to JDK 1.8. It works for me on JDK 11.

@aguibert pointed out that the problem occurs when the Vert.x SQL Client is built in JDK 11. I've verified that the problem is fixed when I use a local JDK 1.8 build of 3.9.3-SNAPSHOT.

The issue is that the Vert.x snapshot built on the sonatype repo was built on JDK 11.

gavinking:
Great, thanks @aguibert, this is now fixed with the latest snapshot. Now all we need is a release.

gavinking:
Closing, was already fixed.

Issue #237, opened 2020-05-30
Multi-table update and delete queries are not supported


gbadner:
Currently, updates and deletes that affect multiple entity tables are not supported 

For example, suppose `Folder`, `Document`, and `SymbolicLink` all extend  ` abstract File`, which has a `name` property, and uses `InheritanceType.TABLE_PER_CLASS`.

That means that `Folder`, `Document`, and `SymbolicLink` each has it's own DB table, each having a `name` column.

In order to delete all `File` entities with name like 'Enron%', rows have to be deleted from all 3 tables.

Fixing this would require something like `MultiTableUpdateExecutor` and `MultiTableDeleteExecutor`.

gavinking:
I have an initial, fairly crappy, fix for this in #325.

gavinking:
This problem is fixed, except on DB2, and I'm just waiting for a release of core to merge #325.

PR #325, opened 2020-08-20
multi-table HQL update+delete

gavinking:
~~Limitations of initial implementation:~~

- ~~does not yet support parameters~~
- ~~does not properly support DB2~~

For #237


gavinking:
@aguibert I ran into some ugly-lookin' errors when I tried to `create` or `declare` a temporary table on DB2. I suspect, but do not know for sure, that this is because the DB2 client doesn't support them yet. Is that right?

aguibert:
@gavinking I haven't tested them specifically, but I would have assumed they work since regular table create statements work. Can you post an example of the SQL you're using to create/declare the temp table and I'll investigate?

gavinking:
> Can you post an example of the SQL you're using to create/declare the temp table and I'll investigate?

Something like `declare global temporary table foo(id int)`

gavinking:
I've added support for parameters, but I need https://github.com/hibernate/hibernate-orm/pull/3507.

aguibert:
Created https://github.com/eclipse-vertx/vertx-sql-client/issues/747 over in vertx-sql-client to track this

gavinking:
Cheers.

gavinking:
Alright, so this work is good to go, once we get a new release of ORM.

gavinking:
@Sanne I would really like to merge this big PR, but I need:

1) This pull request to be merged: https://github.com/hibernate/hibernate-orm/pull/3511
2) A release of core.

PR #330, opened 2020-08-25
upgrade to Mutiny 0.8.0

gavinking:
cc @Sanne


gavinking:
Oh, wait, these API improvements were actually made way back in Mutiny 0.6.0, and when @DavideD upgraded, he didn't change our code to make use of the new APIs. And I guess I was just swapped out.

Anyhow I have pushed them as ba4452454ccaba7705437ed1fcff3b620d08f999.

Issue #342, opened 2020-09-01
JoinColumns trouble


Xset-s:
Hello again, I continue researching your amazing library.
But today i got stuck with problem related to JoinColumns.

There are two entities:

SampleEntity:

```
@Entity
@Table(name = "sample_entities")
public class SampleEntity implements Serializable {
    @Id
    @Column(name = "first_key_id", nullable = false)
    public Long firstKeyId;

    @Id
    @Column(name = "second_key_id", nullable = false)
    public Long secondKeyId;

    @OneToMany(fetch = FetchType.LAZY, mappedBy = "sampleEntity")
    public Set<SampleJoinEntity> sampleJoinEntities;
}
```

And SampleJoinEntity:

```
@Entity
@Table(name = "sample_join_entities")
public class SampleJoinEntity implements Serializable {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    public Long id;

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumns({
            @JoinColumn(name = "first_key_id", referencedColumnName = "first_key_id"),
            @JoinColumn(name = "second_key_id", referencedColumnName = "second_key_id")
    })
    public SampleEntity sampleEntity;
}
```

As you can see there is ManyToOne relation, which uses two keys. 

Let's try to persist them:

```
/* Working with SampleEntity */

final SampleEntity sampleEntity = new SampleEntity();

sampleEntity.firstKeyId = 1L;
sampleEntity.secondKeyId = 2L;

factory.withTransaction((session, transaction) -> session.persist(sampleEntity))
		.await().indefinitely();

/* Working with SampleJoinEntity */

final SampleJoinEntity sampleJoinEntity = new SampleJoinEntity();

sampleJoinEntity.sampleEntity = sampleEntity;

// ISSUE: IT THROWS EXCEPTION HERE!
factory.withTransaction((session, transaction) -> session.persist(sampleJoinEntity))
		.await().indefinitely();

```

When i m trying to persist sampleJoinEntity, hibernate throws this exception:

```
java.lang.UnsupportedOperationException: reactive persistence context
```

Full stack: https://pastebin.com/d68R7VpW

Sample project:
[JoinColumnsProject.zip](https://github.com/hibernate/hibernate-reactive/files/5156575/JoinColumnsProject.zip)

DavideD:
THanks, I'm having a look

gavinking:
[...]

PR #345, opened 2020-09-03
[#342] Support reactive no cascade action

DavideD:
During the noCascade action Hibernate Reactive was calling the non reactive `ForeignKeys.isTransient `.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
@@ -192,13 +192,7 @@ public final class Cascade<C> {
 				}
 				else {
 					if ( action.requiresNoCascadeChecking() ) {
-						action.noCascade(
-								eventSource,
-								parent,
-								persister,
-								types[i],
-								i
-						);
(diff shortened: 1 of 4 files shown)


gavinking (approved):
LGTM @DavideD !

DavideD:
Something fails on CI... I'm checking

DavideD:
@gavinking I cannot figure out why but I have this issue with this PR: https://github.com/hibernate/hibernate-reactive/issues/348

I think I'm going to ignore the failing test for now and get back to it later if I find a way to reproduce it.

Any suggestions?

gavinking:
Look, if you ask me we're not at a point of development here where we need to be anal about the exact type of exceptions that get thrown. And if we want to be anal about it then we're going to need a *hell of a lot* more tests that test the types of much more basic failures.

So if this is just an issue of the concrete exception class, I say don't worry about that for now.

Remember that there is a rather strong policy in Hibernate that exceptions are unrecoverable, and so the concrete exception type isn't very useful most of the time anyway.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/345

I've updated the test and relaxed the check on the exception.

Thanks

PR #350, opened 2020-09-04
break out a new doc section about performance

gavinking:
This change helps keep the section on configuration shorter and sweeter, and lets me address topics in performance more directly rather than as asides.


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/350

Thanks

PR #354, opened 2020-09-05
fix implementation of #342

### LINKED ISSUE #353: Persisting entities with detached objects
Hello again, first of all, thank you so much for your quick response to my issues. Everything works well and i m so glad about that.
But, as far as i'm concerned, for now (after recent commits) its impossible to persist objects with detached entities within.

I mean something like that:

```
final SampleEntity sampleEntity = new SampleEntity();

...

factory.withTransaction((session, transaction) -> session.persist(sampleEntity))
		.await().indefinitely();
```

and then

```
final SampleJoinEntity sampleJoinEntity = new SampleJoinEntity();
sampleJoinEntity.sampleEntity = sampleEntity; // sampleEntity is detached here

...

factory.withTransaction((session, transaction) -> session.persist(sampleJoinEntity))
		.await().indefinitely();
```

After executing ```session.persist(sampleJoinEntity)``` hibernate throws ```object references an unsaved transient instance - save the transient instance before flushing``` (but its not quite transient, its detached 🤔)

Yep, to achive expected functionality, i should use merge:

```
factory.withTransaction((session, transaction) ->
	session.merge(sampleEntity).invokeUni((aSampleEntity) -> {
		final SampleJoinEntity sampleJoinEntity = new SampleJoinEntity();

		sampleJoinEntity.sampleEntity = aSampleEntity;

		return session.persist(sampleJoinEntity);
	})).await().indefinitely();
```

Everything works fine, but in my view, its quite expensive for performance to merge all the necessary entities before persisting. 
Btw some commits ago i was able to persist objects with detached instances within. 

So is that a temporary restriction in your libary? Or maybe i did something wrong and there is a correct way to implement such persisting?

gavinking:
Original change:

4ae4963c5a90db934ac8e4159fd8d40a2e401b67

There were two errors here:

- the return value of `isTransient()` was not used, and so detached instances were treated as transient
- the check was done on every cascade action, instead of just for `PERSIST_ON_FLUSH` as in core

Also the tests assumed that it's wrong for a managed instance to reference a detached instance, but that's a restriction we've never had in Hibernate.

Fixes #353

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CascadingActions.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CascadingActions.java
@@ -98,6 +98,40 @@ public class CascadingActions {
 			LOG.tracev( "Cascading to persist on flush: {0}", entityName );
 			return session.unwrap(ReactiveSession.class).reactivePersistOnFlush( child, context );
 		}
+
+		private boolean isInManagedState(Object child, EventSource session) {
+			EntityEntry entry = session.getPersistenceContextInternal().getEntry( child );
+			return entry != null &&
+					( entry.getStatus() == Status.MANAGED
+							|| entry.getStatus() == Status.READ_ONLY
+							|| entry.getStatus() == Status.SAVING );
+		}
+
+		@Override
+		public CompletionStage<Void> noCascade(EventSource session, Object parent, EntityPersister persister, Type propertyType, int propertyIndex) {
+			if ( propertyType.isEntityType() ) {
+				Object child = persister.getPropertyValue( parent, propertyIndex );
+				if ( child != null
+						&& !isInManagedState( child, session )
+						&& !(child instanceof HibernateProxy ) ) { //a proxy cannot be transient and it breaks ForeignKeys.isTransient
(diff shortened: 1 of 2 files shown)


gavinking:
> assumed that it's wrong for a managed instance to reference a detached instance, but that's a restriction we've never had in Hibernate.

An alternative solution would be to add a `toReference()` operation to `Session`, so that you would be forced do:

```java
Parent detachedParent = ... //detached object
Child newChild = new Child();
newChild.setParent( session.toReference(detachedParent) );
session.save(newChild);
```

That would stop people mixing detached objects and managed objects in the same object graph.

It's an idea worth considering.

DavideD:
@gavinking Thanks!

> Also the tests assumed that it's wrong for a managed instance to reference a detached instance, but that's a restriction we've never had in Hibernate.

I thought you needed cascade enabled for this to work with unmanaged instances. So what's a code example that should throw the exception?

gavinking:
You would get the exception if it's *transient*, not if it's merely detached.

DavideD:
I see, thanks

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-reactive/pull/354

PR #356, opened 2020-09-06
overload getReference() to take detached object/proxy

### LINKED ISSUE #355: Session.asReference()
Hibernate has historically always let you use references to detached objects to set up associations. TBH I'm not even sure if the JPA spec says anything at all about this practice, or if other JPA implementations allow it.

But there are some reasons to think it's a bad thing: it makes it harder, perhaps, for users to keep track of which objects are managed and which are detached in their object graph.

We could *disallow this* (while adding a setting to allow you to recover the legacy behavior), and add a new operation to the `Session` that lets you transform a detached object into an unfetched reference.

```java
Parent detachedParent = ... //detached object
Child newChild = new Child();
newChild.setParent( session.asReference(detachedParent) );
session.save(newChild);
```

TBH it's kinda surprising to me that we don't already *have* an operation like that.

gavinking:
This new method gives you an easy way to transform a detached entity reference to a managed reference without hitting the database.

We should add this to Hibernate ORM core also.

fixes #355

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -369,6 +369,19 @@ public interface Mutiny {
 		<T> T getReference(Class<T> entityClass, Object id);
 
 		/**
+		 * Return the persistent instance of with the same identity as the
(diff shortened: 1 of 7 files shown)


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/356

Thanks

PR #360, opened 2020-09-07
allow proxies to be passed to remove(), lock(), refresh()

### LINKED ISSUE #359: session.remove && session.getReference
Hello, as i understand, following jpa, there is only one way to remove detached entity:

```
factory.withTransaction((session, transaction) -> 
	session.merge(sampleEntity).invokeUni(session::remove)
).await().indefinitely();
```

Everything is fine, but if it's about performance, merge operation takes large amount of time and can cause real performance troubles. For instance, according to my inaccurate time measurements, removing 5000 entities takes ~4s, in hibernate-orm it takes 50ms. 

I've just read about a new method called ```session.getReference``` which returns the persistent instance of given entity. Is it possible to allow its use for ```session.remove```?

gavinking:
automatically fetch them at the start of event processing

fixes #359


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/360

Thanks!

Issue #362, opened 2020-09-10
Test for find after flush


DavideD:
A user had an issue and mentioned this use case. I tried it and it works but I couldn't find any test case for it.

gavinking:
@DavideD So, I mean, I'm not *against* adding tests that just re-test basic functionality of the underlying Hibernate session, but I don't think we need to go out looking for them, and I definitely don't think we need to open an issue for every new test.

DavideD:
OK, I don't plan to make it on habit.

Thanks

gavinking:
No problem.

PR #363, opened 2020-09-10
[#362] Test find after flush

DavideD:
See https://github.com/hibernate/hibernate-reactive/issues/362


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/363

Issue #336, opened 2020-08-28
can't determine native query result types


gavinking:
Some native queries wrongly return null, for example:

    s.createNativeQuery("select current_timestamp", Timestamp.class).getSingleResult()

The cause is https://github.com/eclipse-vertx/vertx-sql-client/issues/657

gavinking:
Once this feature is added to Vert.x, it needs to be wired in to `ReactiveLoader.discoverTypes()`.

DavideD:
We merged a workround with https://github.com/hibernate/hibernate-reactive/pull/364
I will leave the issue open until we can apply the proper solution

PR #364, opened 2020-09-10
temporary workaround solution to #336

gavinking:
An elegant solution requires:

<https://github.com/eclipse-vertx/vertx-sql-client/issues/657>

And we can remove this workaround once it is there.


DavideD:
Done  https://github.com/hibernate/hibernate-reactive/pull/364

thanks!

PR #368, opened 2020-09-17
upgrade to Vert.x 3.9.3

### LINKED ISSUE #336: can't determine native query result types

gavinking:
Fixes #336 and #307.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -9,6 +9,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowIterator;
 import io.vertx.sqlclient.RowSet;
+import io.vertx.sqlclient.desc.ColumnDescriptor;
 import org.hibernate.cfg.NotYetImplementedException;
 import org.hibernate.engine.jdbc.BlobProxy;
 
@@ -34,6 +35,7 @@ import java.sql.Types;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.OffsetDateTime;
 import java.util.Calendar;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


Sanne:
Should we setup CI jobs to check for compatibility with Vert.x snapshots?

It would be quite annoying to not be able to upgrade in case of a regression.

gavinking:
> Should we setup CI jobs to check for compatibility with Vert.x snapshots?

Yes, but I'm even more concerned about Hibernate snapshots.

gavinking:
> @gavinking Something is not right on DB2 and MySQL

Fixed now.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/368

Thanks!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -236,6 +238,17 @@ public Time getTime(String columnLabel) {
...
 		LocalDateTime localDateTime = row.getLocalDateTime(columnLabel);
+		if (localDateTime==null) {
+			try {
+				OffsetDateTime offsetDateTime = row.getOffsetDateTime(columnLabel);
+				if (offsetDateTime!=null) {
+					localDateTime = LocalDateTime.from(offsetDateTime);
+				}
+			}
+			catch (Exception e) {
+				//MySQL client throws a nasty NPE here
```
DavideD, line 249: @gavinking  Can't we do anything better than catch a NullPointer exception?

gavinking: We don't want to. This is what we want to happen.

gavinking:
@DavideD you closed it w/o merging it :-)

DavideD:
Ooops... sorry, I pushed it on my remote instead of upstream

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: I'm pretty sure that this is not how I usually want to deal with NullPointerException, thank you very much.
If you could explain to me why we have to deal with it this way, it would be nice.

gavinking:
Haha

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: Because (1) we're just trying it as a fallback, and (2) the NPE is a bug in the Vert.x MySQL client.

I'm not actually a beginner at this.

gavinking: And FYI it's `Exception` instead of `NullPointerException` because Vert.x is actually trying and failing to throw an `UnsupportedOperationException` but they might fix that.

Sanne:
> > Should we setup CI jobs to check for compatibility with Vert.x snapshots?
> 
> Yes, but I'm even more concerned about Hibernate snapshots.

It's a good point. You don't have to worry as I'm manually checking compatibility with Reactive before tagging an ORM release - but yes it would be nice to automate this.

DavideD:
@Sanne maybe you can have a look at this: https://github.com/hibernate/hibernate-reactive/pull/380

PR #371, opened 2020-09-17
make Mutiny.Session more friendly for method refs

### LINKED ISSUE #370: persist(), merge(), remove(), refresh() on Mutiny.Session
With the Mutiny API changes I requested in 0.8.0, the new `call()` method which replaces `invokeUni()` is now overloaded and this creates an ambiguity when you do:

```java
session.find(Book.class, id).call(session::remove)
```

And so you have to write:

```java
session.find(Book.class, id).call( (b) -> session.remove(b) )
```

So I think we should get rid of the overloading of `persist()`, `merge()`, `remove()`, and `refresh()` on `Mutiny.Session`.

But what should we do instead?

1. Just leave them as unary operations?
2. `persistAll(Object... entities)` 
3. `persist(Object entity, Object... moreEntities)`

I lean toward 2, but eeeeewww, shades of JDO's `makePersistentAll()`!

gavinking:
Fixes #370

### DIFF
--- a/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
+++ b/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
@@ -40,7 +40,7 @@ public class MutinyMain {
 		// obtain a reactive session
 		factory.withTransaction(
 				// persist the Authors with their Books in a transaction
-				(session, tx) -> session.persist(author1, author2)
+				(session, tx) -> session.persistAll(author1, author2)
 		)
 				// wait for it to finish
 				.await().indefinitely();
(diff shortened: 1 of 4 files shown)


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/371
Thanks

I know that it is not required in that case, but should we rename the same methods in `Stage.Session` for consistency?

DavideD:
By the way, these new addition to the Mutiny API are really good

gavinking:
> I know that it is not required in that case, but should we rename the same methods in `Stage.Session` for consistency?

Well, I dunno. I decided not to, obviously, because I actually dislike the `persistAll()` naming. I guess I think it's OK if the two `Session` APIs reflect differences between the stream APIs and we intend them to be used.

> By the way, these new addition to the Mutiny API are really good

Yep, actually they turned out even better than I had thought. (Though I had a bit of a fright when I first sam that I couldn't pass `session::remove()` as a method ref!)

PR #373, opened 2020-09-18
add config properties for connection pool timeouts

### LINKED ISSUE #357: extra connection pool connection-related properties
`SqlConnectionOptions` inherits a bunch of properties from `NetClientOptions` and it seems to me that some of these are relevant, and should be configgable via Hibernate properties. For example, both of the following look useful:

- `hibernate.vertx.pool.idleTimeout`
- `hibernate.vertx.pool.connectTimeout`

@vietj should we add these two properties? Are there any others that would be useful to someone configuring a connection pool for use in HR?

gavinking:
fixes #357


DavideD:
Rebased and merged https://github.com/hibernate/hibernate-reactive/pull/373

Issue #374, opened 2020-09-18
lazy collections and bytecode enhancer


gavinking:
So I've run into what I consider to be some pretty unexpected behavior of Hibernate ORM when the bytecode enhancer is enabled with `enableLazyInitialization = true`.

There's two aspects of the behavior that, naïvely, look suboptimal to me, at least in the context of HR:

1. By default, lazy collection are marked for field-level lazy fetching. I gotta say I just don't see much advantage to this, how is it better than the collection proxy? Indeed, it's arguably worse, since direct field access doesn't get intercepted. That's tolerable for fields that the user has *explicitly* marked for lazy fetching, but not for this *implicit* behavior.
2. When you *do* fetch the collection, `AbstractEntityPersister.initializeLazyProperty()` doesn't set the field of the entity like what it does for other lazy fields. That's either a bug, or it's due to how we're calling this method in Hibernate reactive. Whatever, it means that `AbstractEntityPersister.initializeLazyProperty()` has a pretty fragile contract.

@Sanne I know you're an expert on this stuff, can you help me understand all this please?

gavinking:
@Sanne the workaround in #375 solves problem (2) at least for HR. But shouldn't this be done in core?

Sanne:
Hi Gavin, I'm not sure on terminology which might be important; and no I'm not the expert in this area, I've just started recently to fix some issues with it in the context of making the existing stuff work in Quarkus. What do you mean by `field-level lazy fetching` exactly?

And by "bytecode enhancer is enabled", I assume you're referring to build time enhancement?

It might help to keep in mind that we can't do proxies in Quarkus, as you can't have a proxy compiled to native image; so we're focusing on bytecode manipulation of the class - which wasn't indeed as stable as I'd like it, but it has other advantages as well over proxies: the type identity is consistent at runtime, which helps with various subtle bugs people had, e.g. integrating with other tools and frameworks.

Regarding field access: we actually intercept field accessors too and re-map them; in particular Panache likes this.

So _conceptually_ I don't feel it's a bad choice, but in practice this is possibly not very mature yet; we certainly had a list of issues manifesting since Quarkus made this model more popular. So please give me some more context on what didn't work - did you have a problem in Quarkus, or a different test?

Sanne:
There's one more novelty to keep in mind; we now support "enhanced proxies": this allows some loading patterns to perform much better.
See Steve's blog for a description:
[...]

PR #375, opened 2020-09-18
workaround for collections in field-level lazy fetching

gavinking:
I feel like this must be a bug in Hibernate ORM that I'm dealing with here.

See discussion in #374.


Sanne:
I'm a bit lost indeed, let me catch up..

Sanne:
You won't need this fix anymore if we merge your proposal in ORM right?

gavinking:
Aaahm.... it wouldn't be *necessary*, though I might still decide to merge it *anyway*. Need to think about that some more.

gavinking:
Actually I'm going to merge this anyway. I can always rip out the stuff that sets the collection field if we turns out we ultimately don't need it.

Issue #247, opened 2020-06-03
query cache


gavinking:
Closely related to #184, we should test the query cache, and if it's working, add the methods to the `Query` interfaces that are needed for managing it.

gavinking:
As reported by @DavideD in #298, the query cache doesn't work when the query result set contains entity instances that aren't themselves cached.

This is likely quite easy to fix.

```
Unexpected access to the database
org.hibernate.AssertionFailure: Unexpected access to the database
	at org.hibernate.reactive.event.impl.DefaultReactiveLoadEventListener.onLoad(DefaultReactiveLoadEventListener.java:102)
	at org.hibernate.event.service.internal.EventListenerGroupImpl.fireEventOnEachListener(EventListenerGroupImpl.java:113)
	at org.hibernate.internal.SessionImpl.fireLoadNoChecks(SessionImpl.java:1184)
	at org.hibernate.internal.SessionImpl.internalLoad(SessionImpl.java:1049)
	at org.hibernate.reactive.session.impl.ReactiveSessionImpl.internalLoad(ReactiveSessionImpl.java:148)
	at org.hibernate.type.EntityType.resolveIdentifier(EntityType.java:697)
```

gavinking:
@DavideD It would be nice to know what exactly was the test you ran that resulted in the above error.

DavideD:
I've created this branch with the test case: https://github.com/DavideD/hibernate-reactive/tree/247-Testcase
This is the test case: https://github.com/DavideD/hibernate-reactive/commit/4b395d5d713df00e94aa632cace3ba7809f3012e
It requires an `import.sql` maybe you know a way to avoid that?

The error is different now but I think it's a result of the latest changes. I didn't make much progress on solving this issue, if you want to have a look at it maybe you can solve it quickly.

DavideD:
I will get back to this now

DavideD:
Done

PR #376, opened 2020-09-18
[#247] [DON'T MERGE] Load query cache results

DavideD:
I think I need some help with #247.

What's happening is that when the query result is cached but one of the entity in the returned list isn't, the code will try to load the result from the db.
This will execute a query that returns a `CompletionStage` and causing
an AssertionFailure exception: https://github.com/hibernate/hibernate-reactive/blob/master/hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/DefaultReactiveLoadEventListener.java#L102

I couldn't think of any simple solution for this scenario, except for rerunning the query ignoring the cache when we throw the exception.
Note that this works but it will log the error on the console because [AssertionFailure](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/AssertionFailure.java) logs an error every time is created.

Thanks

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/CachingReactiveLoader.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/CachingReactiveLoader.java
@@ -5,6 +5,7 @@
  */
 package org.hibernate.reactive.loader;
 
+import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
 import org.hibernate.QueryException;
 import org.hibernate.cache.spi.FilterKey;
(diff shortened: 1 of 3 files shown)


gavinking:
> What's happening is that when the query result is cached but one of the entity in the returned list isn't, the code will try to load the result from the db.

Yep, that's what I figured the challenge with this would be.

gavinking:
> I couldn't think of any simple solution for this scenario, except for rerunning the query ignoring the cache when we throw the exception.

That's perhaps not worse than the behavior of core.

Imagine if there are N cache missed, then we get N queries to the database instead of just one.

So I'm not against the behavior you suggest.

gavinking:
> Note that this works but it will log the error on the console because `AssertionFailure` logs an error every time is created.

Well that's our code. We can make it throw a different kind of exception.

DavideD:
I've sent a new PR (https://github.com/hibernate/hibernate-reactive/pull/377). Closing this one.

Issue #361, opened 2020-09-09
Check cross repo build with GitHub


DavideD:
https://github.com/marketplace/actions/github-action-build-chain-cross-repo-builds

I haven't looked into it yet so I've created this issue so that I won't forget it, but could we use it to test with the latest Hibernate ORM SNAPSHOTS?

DavideD:
It seems that it would require all project to have a workflow on GitHub. This is not the case for ORM.
I think it's faster to create a job on Hibernate Jenkins for now.

DavideD:
Done.

In the end it was simpler to create a GitHub workflow that checks releases regularly and during push/pr requests.

PR #380, opened 2020-09-22
[#361] Track ORM 5.4 snapshots

DavideD:
This approach uses GitHub workflows, the JBoss Snapshots repository to download the latest 5.4 ORM SNAPSHOT and will be triggered by push  or pr on master and every hour via a cron job.

I'm testing and running the examples on all dbs but I think a simpler version that runs on Postgres might be enough. I've created it on a separate branch: https://github.com/DavideD/hibernate-reactive/pull/8

I've added a workflow to test 5.5 as well so that I can see that it fails for it but we can remove it for now.

If we want to build the project when ORM build, I will have to look into Jenkins.

### DIFF
--- a/.github/workflows/tracking-orm-5.4.build.yml
+++ b/.github/workflows/tracking-orm-5.4.build.yml
@@ -1,0 +1,122 @@
+name: Latest ORM 5.4
+
+on:
+  # Trigger the workflow on push or pull request,
+  # but only for the master branch
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # Run every hour at minute 25
+    - cron: '25 * * * *'
+
+jobs:
+  # The examples test the Hibernate ORM Gradle plugin. We use it for bytecode enhancements.
+  run_example_mysql:
+    name: Run examples on MySQL
+    runs-on: ubuntu-latest
+    services:
+      # Label used to access the service container
+      mysql:
+        # Docker Hub image
+        image: mysql
+        env:
+          MYSQL_ROOT_PASSWORD: hreact
+          MYSQL_DATABASE: hreact
+          MYSQL_USER: hreact
+          MYSQL_PASSWORD: hreact
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd="mysqladmin ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 3306:3306
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        with:
+          java-version: 1.8
+        uses: actions/setup-java@v1
+      - name: Print the effective ORM version used
+        run: ./gradlew :example:dependencyInsight --dependency org.hibernate:hibernate-core -PhibernateOrmVersion='[5.4,5.5)' -PskipOrmVersionParsing
+      - name: Run examples on MySQL
+        run: ./gradlew :example:runAllExamplesOnMySQL -PhibernateOrmVersion='[5.4,5.5)' -PskipOrmVersionParsing
+
+  run_example_postgres:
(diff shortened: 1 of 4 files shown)


gavinking:
Excellent.

DavideD:
I guess the disadvantage is that I need to add the snapshots repository to the list of repositories. I don't think it will cause issue considering that we don't have snapshots dependencies otherwise.

DavideD:
For the Vert.X SQL Client I don't think there is a snapshot available but we could test things every hour/day using the latest released version.

For example:
```
./gradlew build -PvertxVersion='[3,4)'
```
I guess at least we would get notified as soon as there is a new broken release

DavideD:
I think I've done with this one:
* I've kept the tests for the examples because they use the Hibernate ORM Gradle plugin
* I've added a workflow for Vert.x 3, it will only test the released versions but I guess is good enough for now
* The workflows should start once per hour, on prs and on push for master. If it gets too noisy we can tweak it later
* It requires the Hibernate ORM SNAPSHOT repository.  We could exclude it from the normal build using a property but the build 
  file is starting to get full of this customizations and I would prefer not to add other. Maybe we can refactor it later as a second 
  step.

The PR check doesn't seem to take longer even with the new workflow.

Cheers

DavideD:
I've removed the 5.5 workflow

gavinking:
> I've removed the 5.5 workflow

do you have one which at least tracks master?

DavideD:
What do you mean?

DavideD:
ORM master is 5.5. It's not working at the moment.

DavideD:
I will create an issue for it

gavinking:
> ORM master is 5.5. It's not working at the moment.

WDYM “not working”?

DavideD:
https://github.com/hibernate/hibernate-reactive/issues/384

There has been an API change

gavinking:
So does that mean it's going to be a problem whenever there is some breaking change on master that isn't backported to 5.4 yet?

gavinking:
> So does that mean it's going to be a problem whenever there is some breaking change on master that isn't backported to 5.4 yet?

Or can we have a special branch of HR that tracks ORM master?

DavideD:
> So does that mean it's going to be a problem whenever there is some breaking change on master that isn't backported to 5.4 yet?

Yes

> Or can we have a special branch of HR that tracks ORM master?

Sure, we can have a reactive branch 5.5  that works with ORM 5.5.
I think the only reason we haven't done that so far is because we were more interested to be aligned to quarkus.

DavideD:
Or vice-versa, keep the reactive master aligned to 5.5 and create a 5.4 branch. Hopefully, it won't take too much work to maintain both.

WDYT?

DavideD:
> I think the only reason we haven't done that so

Also, there was no need becayse we didn't have any breaking API

Sanne:
I haven't had the chance to look at it yet, but I hope we can either fix
the breakage or backport it, so to not need additional branches.

gavinking:
> WDYT?

Not sure, need to think it through.

DavideD:
It seems for reactive to work with 5.5 the fix is pretty simple: https://github.com/DavideD/hibernate-reactive/commit/05c410850aa741350b49746328f3a9f8e5906258

Not sure if I'm missing something.

I have on my repository: https://github.com/DavideD/hibernate-reactive/tree/orm-5.5

I've also updated the workflow for it: https://github.com/DavideD/hibernate-reactive/pull/10

gavinking:
> It seems for reactive to work with 5.5 the fix is pretty simple: [DavideD@05c4108](https://github.com/DavideD/hibernate-reactive/commit/05c410850aa741350b49746328f3a9f8e5906258)

That doesn't look bad at all.

DavideD:
Any reason not to merge this while we wait for the  issue on 5.5 to be backported?

gavinking:
No, no reason that I can see.

DavideD:
Rebased and merged https://github.com/hibernate/hibernate-reactive/pull/380

PR #387, opened 2020-09-29
Upgrade tests and instructions to use PostgreSQL v 13.0

Sanne:


DavideD (approved):
LGTM

DavideD:
Actually, I think you need to update the TestContainer configuration as well

DavideD:
@Sanne check [PostgreSQLDatabase#IMAGE_NAME](https://github.com/hibernate/hibernate-reactive/blob/1cb2041e01e6cb7e8f185e900e2d549abaeee571/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/PostgreSQLDatabase.java#L14)

Sanne:
thanks, didn't know about that one :)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/PostgreSQLDatabase.java
```diff
@@ -11,7 +11,7 @@
 
 	public static PostgreSQLDatabase INSTANCE = new PostgreSQLDatabase();
 
-	public final static String IMAGE_NAME = "postgres:12-alpine";
+	public final static String IMAGE_NAME = "postgres:13.0";
```
DavideD, line 14: Do we need to be extra precise? Wouldn't `13` be good enough?

DavideD: Other than that, feel free to merge it if it passes the checks

Sanne: I prefer using specifi tags in CI, so that stuff won't start failing because of an upgrade and we have no clue about it.

Sanne: For example: our github actions was likely already testing on pgsql v 13 as it was configured to use the latest.

Would you have suspected that new failures could have been caused by t he fact that PostgreSQL 13 was just released last week?

But also, it means our readme was not correct, as it states we're testing with v 12.

DavideD: > Would you have suspected that new failures could have been caused by t he fact that PostgreSQL 13 was just released last week?

Maybe

I understand the argument but I think it's easier to test for the latest 13.x version than update the version each time. I don't expect to have to deal with an exception because of this very often .

Anyway, it's ok for me to merge. 

We have this other workflow that's testing with the latest ORM 5.4 version available: https://github.com/hibernate/hibernate-reactive/blob/master/.github/workflows/tracking-orm-5.4.build.yml

I guess I will add a comment to mention that it also using the latest db

DavideD: @blafond This is a fix for PostgresSQL, but could you create an issue and update the images for MySQL and DB2 as well, please?

Ping me if something is not clear.
Thanks

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-reactive/pull/387

Issue #388, opened 2020-09-29
Reactive not working with the latest snapshot of ORM


DavideD:
https://github.com/hibernate/hibernate-reactive/actions/runs/278608521

I already have a fix.

PR #390, opened 2020-09-29
[#388] Fix the build with the latest ORM snapshot

DavideD:
Mostly because I'm receiving an email every hour.
This will fix the build but I would like to add a different solution when some issues on 5.5 get backported.


DavideD:
It's not necessary anymore

PR #392, opened 2020-10-01
Changes after ORM 5.4.22

### LINKED ISSUE #374: lazy collections and bytecode enhancer

gavinking:
After the update to 5.4.22, the following changes can be made:

- put collections in default fetch group
- let ORM core do the logger syntax highlighting

Fixes #374.


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -36,27 +33,21 @@
...
-	private final boolean formatSQL;
-	private final boolean highlightSQL;
+	private final SqlStatementLogger sqlStatementLogger;
 	private final boolean usePostgresStyleParameters;
 
 	private final Pool pool;
 	private final SqlConnection connection;
 	private Transaction transaction;
 
 	SqlClientConnection(SqlConnection connection, Pool pool,
```
DavideD, line 45: @Sanne This might affect what you where doing for Quarkus

gavinking: It shouldn't, I don't think.

DavideD: The class that in Quarkus extends the SqlClientPool needed some refactoring and these properties were a bit problematic when I talked to him about it. This PR might make things easier.

Anyway, I don't think this is a blocking issue for us. I just wanted the mentioned that this part is changing to him.

gavinking: Yeah, but this is a change that's completely *internal* to `SqlClientPool`.

DavideD: Sure, ok

DavideD:
LGTM

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
Sanne: Thanks, yes this is useful to know.

PR #395, opened 2020-10-02
Allow existing SqlClientPool instance to be used

Sanne:
Some refactoring is necessary to allow the Quarkus integration to adapt to the latest improvements in the Hibernate Reactive SqlClientPool.

I've extracted `BaseSqlClientPool` as a parent abstract class from the existing `SqlClientPool`, to then create two concrete implementations:
 - ManagedSqlClientPool for the case in which the Hibernate SessionFactory will control (manage) its lifecycle (this matches the existing semantic)
 - InjectedSqlClientPool for when an external container wants to inject an existing insance into our SessionFactory (this is what Quarkus needs)

`ProxyConnection` was extracted as well to simplify things a bit, and we can now use the existing `SqlStatementLogger` from ORM.

Preview of the Quarkus upgrade patch:
 - https://github.com/quarkusio/quarkus/commit/2953848f27ca0cea60ae305fd302c9d20b63ef28


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/BaseSqlClientPool.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * A pool of reactive connections backed by a Vert.x {@link Pool}.
+ * <p>
+ * This class may be extended by programs which wish to implement
+ * custom connection management or multitenancy.
+ *
+ * @see SqlClientPoolConfiguration
+ */
+public abstract class BaseSqlClientPool implements ReactiveConnectionPool {
```
gavinking, line 28: I would just call this class `SqlClientPool`. The `Base` bit seems redundant.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/InjectedSqlClientPool.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A pool of reactive connections backed by a Vert.x {@link Pool}.
+ * <p>
+ * This is an alternative to using {@see ManagedSqlClientPool}, for
+ * the case in which one doesn't want Hibernate ORM to manage
+ * the lifecycle of the underlying pool.
+ * <p>
+ * This implementation is meant to be used in Quarkus.
+ */
+public final class InjectedSqlClientPool extends BaseSqlClientPool {
```
gavinking, line 21: Does this class actually have anything to do with dependency injection? If not, then I would like to find a different name for it. `CustomSqlClientPool` or `ExternalSqlClientPool` or `ContainerSqlClientPool` or something like that.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/VertxInstance.java
```diff
@@ -7,10 +7,11 @@
 
 import io.vertx.core.Vertx;
 
+import org.hibernate.reactive.pool.impl.ManagedSqlClientPool;
 import org.hibernate.service.Service;
 
 /**
- * Used by {@link org.hibernate.reactive.pool.impl.SqlClientPool}
+ * Used by {@link ManagedSqlClientPool}
```
gavinking, line 14: Probably better to keep the FQN here instead of using `import`, since you'll sometimes get warnings about an unused import (depending on the tool).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ManagedSqlClientPool.java
```diff
@@ -0,0 +1,205 @@
...
+ * custom connection management or multitenancy.
+ * <p>
+ * This pool is started on SessionFactory start, then stopped when
+ * the SessionFactory is stopped: its lifecycle is managed by Hibernate.
+ * For a non-managed pool whose lifecycle is managed externally,
+ * {@see InjectedSqlClientPool}.
+ *
+ * @see SqlClientPoolConfiguration
+ */
+public class ManagedSqlClientPool extends BaseSqlClientPool implements ServiceRegistryAwareService, Configurable, Stoppable, Startable {
```
gavinking, line 51: How about `DefaultSqlClientPool`?

gavinking (changes requested):

gavinking:
P.S. @Sanne feel free to go ahead and merge this, if you're happy for me to change the names in a separate commit, after thinking them through a bit further.

Sanne:
Thanks, good suggestions. I'll make your changes, then merge, then you can polish it futher if you see any need.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/InjectedSqlClientPool.java
Sanne: +1 for `ExternalSqlClientPool`. I'd avoid the other two as custom doesn't express much, and Quarkus isn't a "container" in the classical sense.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ManagedSqlClientPool.java
Sanne: Ok

hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/VertxInstance.java
Sanne: +1 that wasn't intentional. IDE magic..

Sanne:
Re-pushed to have CI verify things.

gavinking:
Thanks @Sanne, I think this is a good improvement to the design.

Sanne:
Sure let's clarify the doc and its purpose, not arguing about that.

I'm just not sure about spending too much time about explaining the essentials, I'd rather have it point to the better maintained podman references than to have a section to explain how to install a native tool for each OS out there ...

PR #396, opened 2020-10-03
update to Mutiny 0.9.0

gavinking:
And clean up tests.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
@@ -61,18 +61,17 @@ public class MutinySessionTest extends BaseReactiveTest {
 		test(
 				context,
 				populateDB()
-						.onItem().transform( i -> openMutinySession() )
-						.onItem().call( session -> session.find( GuineaPig.class, expectedPig.getId() )
-						.onItem().invoke( actualPig -> {
-							assertThatPigsAreEqual( context, expectedPig, actualPig );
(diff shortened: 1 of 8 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -389,7 +394,7 @@ public void reactiveUpdateVersion(TestContext context) {
 								.call(session::flush)
 								.eventually(session::close)
 						)
-						.chain( () -> selectNameFromId( 5 ) )
+						.replaceWith( selectNameFromId(5) )
```
DavideD, line 397: Only because this is a PR about the style of the code:
shouldn't it be `.replaceWith( selectNameFromId( 5 ) )` (with spaces surrounding the 5)?

gavinking: I don't think there's any reason on earth to put spaces around a numeric literal. It actually makes it *harder* to parse the nested parens, not *easier*.

gavinking:
Hrm. Why do I have failing tests on the CI? They pass on my local machine.

gavinking:
> Hrm. Why do I have failing tests on the CI? They pass on my local machine.

OK, so I need to be a little bit careful about blindly `replaceWith(uni)` vs `chain(()->uni)`. They're not the same. `replaceWith(Uni)` isn't supposed to depend on the receiver being completed.

gavinking:
Alright, I think this is good to go.

gavinking:
Squashed it and merged it.

Issue #367, opened 2020-09-16
Create a test case that declares an ORM persistent unit and a Reactive one


DavideD:
This use case is supported right now but we don't have a test for it.
Something not too complicated: an entity and a couple of unit tests persisting and reading it using JPA/Hibernate and Reactive persistence unit declared in the same persistence.xml.

In OGM we use a rule to isolate among different persistence.xml for tests, it might be helpful for us as well: https://github.com/hibernate/hibernate-ogm/blob/master/core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java

We should test it for Mutiny and CompletionStage.

@blafond This could be a good first issue for you to get accustomed to the project and the way we work

DavideD:
Done!

Thanks

PR #402, opened 2020-10-05
common orm and reactive persistence tests #367

blafond:


hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+		Session ormSession = ormFactory.openSession();
+		ormSession.beginTransaction();
+		ormSession.persist( rose );
+		ormSession.getTransaction().commit();
+		ormSession.close();
+
+		// Check database with Mutiny session and verify 'rose' flour exists
+		Mutiny.Session mutinySession = openMutinySession();
+		// find the entity using the id
+		test( context, mutinySession.find( Flour.class, rose.id)
```
DavideD, line 100: > // find the entity using the id
> test( context, mutinySession.find( Flour.class, rose.id)

I'm not a big fun of comments describing exactly the signature of the method. I think you can remove it.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+				// and check it's the right one
+				.invoke( foundRose -> context.assertEquals( rose, foundRose ) )
+				// and that it's not the wrong one
+				.invoke( noAlmond -> context.assertNotEquals(rose, sunflower))
```
DavideD, line 104: Is it really useful to test that the two objects are different? You have already checked that rose is the expected one.
The only way this assert can fail is if there is something wrong in the test.
It doesn't seem very helpful.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+			  	// Close session
+				.eventually(mutinySession::close)
+				// verify session closed
+				.invoke( sessionClosed -> context.assertFalse(mutinySession.isOpen()) )
```
DavideD, line 108: No need to close the session and it. The super class takes care of it.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import io.vertx.ext.unit.TestContext;
+
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.POSTGRESQL;
+import static org.hibernate.reactive.util.impl.CompletionStages.completedFuture;
+
+public class ORMReactivePersistenceTest extends BaseReactiveTest {
```
DavideD, line 33: I think in this case a brief javadoc explaining what this test is about is useful

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+		StandardServiceRegistryBuilder builder = new StandardServiceRegistryBuilder()
+				.applySettings( configuration.getProperties() );
+
+		StandardServiceRegistry registry = builder.build();
+		ormFactory = configuration.buildSessionFactory( registry );
+	}
+
+	@After
+	public void cleanDb(TestContext context) {
+		ormFactory.close();
```
DavideD, line 62: If an error happens here the db doesn't get cleaned. Maybe we should add a try catch. I think we can ignore the error print it (if an error occurs here is probably due to something else).

DavideD (changes requested):

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
blafond: Something like this?

/**
 * These tests verify that data can be persisted and queried on the same database
 * using both JPA/hibernate and reactive Mutiny or CompletionStage session factories
 */

DavideD: Maybe something like this:
```
Test that it's possible to use an ORM and Reactive session in the same app
<p>
The test creates an entity using the ORM session and then retrieve it using the reactive one.
```

I don't have strong opinion, feel free to pick the one you prefer.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ORMReactivePersistenceTest.java
blafond: print as in....      **context.fail(exception);**     ?

DavideD: Actually, forget about it. We clean the data when we re-run the tests so it's fine the way it is

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-reactive/pull/402

Issue #324, opened 2020-08-20
podman instructions aren't understandable


gavinking:
@DavideD I still don't understand the [podman instructions](https://github.com/hibernate/hibernate-reactive/blob/master/podman.md), and since I can't run Podman on my Mac, I can't fix them.

- What do "Server" and "CLI" mean here? The document doesn't explain. Are they alternatives approaches or do you need to run *both* commands?
- Are there any additional steps that this document assumes you've already taken before running these commands? (e.g. installing Podman)
- Once you've started the test databases, how do you run tests? (*I* think I know, but I'm not the target audience.)
- Why would I even want to use Podman? Does it offer some specific advantage?
- Do we really still need this document now that Docker is working great for us?

DavideD:
We don't need to mention podman but I would like to keep the scripts. We can replace them for docker.
Even if we have the options in the tests I find it handy to have them so that I can copy and paste what I need  if I want to start a machine separately (and to test that the build works without TestContainers).

> What do "Server" and "CLI" mean here? The document doesn't explain. Are they alternatives approaches or do you need to run both commands?

* Server: The db
* CLI: the [Command line interface](https://en.wikipedia.org/wiki/Command-line_interface) that you can use to connect to the db. For example `psql` for postgres or `mysql` for MySQL. This might not be available immediately and it's much easier to connect this way if you need to quickly check the content of the db.

> Are there any additional steps that this document assumes you've already taken before running these commands? (e.g. installing Podman)

We can change it so that everything refers to docker and podman is not mentioned.
We could probably link to the Docker homepage in the readme though.

> Once you've started the test databases, how do you run tests? (I think I know, but I'm not the target audience)

We explain this here: https://github.com/hibernate/hibernate-reactive#if-you-have-podman

> Why would I even want to use Podman? Does it offer some specific advantage?

* Installing Podman in the latest Fedora is pretty straightforward, installing docker requires several steps. I don't think having the instructions in our project for it is a lot of additional work for us and you can use the same options for Docker. Also podman run root-less and doesn't require a daemon.
But podman is meant to be a replacement for docker (you can write alias docker=podman) so we can reword everything and rewrite the instructions for docker so that podman is not mentioned. 

* Do we really still need this document now that Docker is working great for us?

[...]

PR #404, opened 2020-10-08
[#324] Add details on how to run the build with Podman

DavideD:


blafond:
So running through the set-up and test with new instructions

1) Wouldn't it be helpful to add a note on what to expect when a user runs "./gradle test" ?

I'm getting:

Configure project :
Parsing ORM version
Hibernate ORM Version: 5.4.22.Final
Vert.x Version: 3.9.3

BUILD SUCCESSFUL in 1s
11 actionable tasks: 11 up-to-date

2) noticed you changed the command to **./gradlew build** rather than **./gradlew test** ... guessing they are the same in this case?

DavideD:
> Wouldn't it be helpful to add a note on what to expect when a user runs "./gradle test" ?

I don't know. Apart for the versions of the libraries, everything else is the standard  Gradle output.
Once you've reached the `BUILD SUCCESSFUL` message. was there something not clear about them?
They might also change in the future and I would prefer not having to update the file each time.
These files are already quite big and they will become bigger with the addition of new datastores.

> noticed you changed the command to **./gradlew build** rather than **./gradlew test** ... guessing they are the same in this case?

They are basically the same but you are right. We used test everywhere else and I prefer to be consistent.
I will change it.

podman.md
```diff
@@ -1,61 +1,110 @@
-# Running Hibernate Reactive tests on Podman
+# How to start the test databases using Podman
 
-As an alternative to Docker, you can use [podman] to start the databases used 
-by the test suite for Hibernate Reactive.
+The following sections will use [Podman][podman] to start the database of your choice, create the schema
+and create the credentials required to run the tests during the build.
+
+Podman is a daemonless container engine for developing, managing, and running OCI Containers on your Linux System.
+Containers can either be run as root or in rootless mode.
```
gavinking, line 7: A minor comment: `.md` files are supposed to be readable as text, so it's important to maintain a consistent line-length.

gavinking (changes requested):

podman.md
DavideD: ok

DavideD:
I've applied the remarks. I'm going to merge it if there isn't anything else.

README.md
```diff
@@ -160,13 +160,10 @@ directory.
 
 #### If you have Podman
 
-If you use [Podman][], you can start the test database by following 
+If you have [Podman][podman] installed, you can start the test database by following
```
gavinking, line 163: @DavideD I know it's picky, but this line is also really long compared to the rest of the README

gavinking (changes requested):

README.md
DavideD: Should be fixed now

gavinking (approved):

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/404

Thanks

Issue #413, opened 2020-10-20
Remove warnings during doc generation


DavideD:
Example:
```
asciidoctor: WARNING: introduction.adoc: line 510: invalid style for listing block: java
```

```
ReactiveLoaderBasedResultSetProcessor.java:54: warning - Tag @link: reference not found: Loader#processResultSet
```

gavinking:
> asciidoctor: WARNING: introduction.adoc: line 510: invalid style for listing block: java

Well yeah, but I would love to know why the fuck ASCII doctor can't highlight code fragments for me and instead insists that a java listing has to be a whole fucking class!

We should just open a bug against ASCII doctor because that behavior is simply rubbish.

DavideD:
> We should just open a bug against ASCII doctor because that behavior is simply rubbish.
 
This was actually an issue with the IntelliJ plugin

PR #414, opened 2020-10-20
[#413] Fix warnings during docs generation

DavideD:


documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -497,7 +497,7 @@ Whatever you decide, the first step to getting a reactive session is to obtain
 a JPA `EntityManagerFactory` just as you usually would in plain ol' regular JPA,
 for example, by calling:
 
-[java]
+[source, JAVA, indent=0]
```
gavinking, line 500: @DavideD but this doesn't work, does it?

When I tried it I got errors.

DavideD: I don't see any error

gavinking: Ah, well, actually the error is in IntelliJ, so perhaps it's a bug in IntelliJ not in ASCII doctor itself?

gavinking: Indeed, seems to be a bug in the IntelliJ plugin. I had taken it to be a limitation of asciidoc itself.

gavinking (approved):

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: IntelliJ now is giving me two warnings:
* one for `[source, JAVA, indent=0]`
* one for the java code not starting with class or something

I will replace it with `[source,java]` but the other problem is an Intellij thing. Asciidoc doesn't seem to care about it.

DavideD: Ah no wait, forget what I said. IntelliJ was refreshing or something.

It seems an issue with IntelliJ though. I usually use VIM for document changes so I wouldn't know.

gavinking: > It seems an issue with IntelliJ though.

Yes, I agree. So go ahead and merge.

PR #415, opened 2020-10-20
[#324] Remove spaces at the end of the line in podman.md

DavideD:
@Sanne this should fix the issue you were having. 
Was there something else?


Sanne (approved):
Thanks!

Issue #419, opened 2020-10-23
Publish the documentation on hibernate.org during the release


DavideD:
This way we don't have to rely on the JBoss doc server

DavideD:
Done

PR #420, opened 2020-10-23
[#419] Publish the documentation on hibernate.org

DavideD:
I've used this plugin: https://github.com/ajoberstar/gradle-git

Do we still want to publish on the JBoss doc server? For now I've removed it

I've tested it locally but we need to wait for the release to see if it actually works.

Better merge this after we merged the reactive website: https://github.com/hibernate/hibernate.org/pull/155

### DIFF
--- a/release/build.gradle
+++ b/release/build.gradle
@@ -6,6 +6,12 @@ plugins {
     id 'org.ajoberstar.git-publish' version '3.0.0'
 }
 
+ext {
+    if ( !project.hasProperty('docUploadBranch') ) {
+        docUploadBranch = 'staging'
+    }
+}
+
 description = 'Release module'
 // (Optional) before uploading the documentation, you can check
 // the generated website under release/build/hibernate.org with:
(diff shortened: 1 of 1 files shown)


DavideD:
There something wrong with the javadocs on the generated website on staging.
I will check on Monday.

DavideD:
OK, publishing the html directly on hibernate.org doesn't work because Awestruct changes all the links

yrodiere:
> OK, publishing the html directly on hibernate.org doesn't work because Awestruct changes all the links

Did you consider pushing to a separate repository that awestruct is not aware of? You can just copy the data to `_site` after the Awestruct build, using a script.

yrodiere:
I believe you can add directories to be ignored by Awestruct here: https://github.com/hibernate/hibernate.org/blob/staging/.awestruct_ignore

DavideD:
I think what that does is to completely ignore the folders during the generation. But I think this is what I was looking for: http://www.awestruct.org/extensions/indexifier/

In the avoiding some paths section

DavideD:
OK, I think now it is working. I've updated staging as well

DavideD:
This PR only make sense if the Reactive website is available

PR #421, opened 2020-10-25
[#410] Upgrade to Gradle 6.7

### LINKED ISSUE #410: Gradle upgrade to 6.7
It would be useful for, among others:

- be able to test on JDK15
- Toolchain support: allows us more freedom in building and testing with different JDK versions (not necessarily aligned)

In particular I believe we should be always building with the JDK we intend to do releases with, while testing with a range of JDKs (currently the runtime testing JDK also has us build with the same JDK).

There's some more goodies, but less fo of interest:
 - https://docs.gradle.org/current/release-notes.html

Sanne:
Fixes #410


DavideD:
Are the empty lines critical? Do you mind if I remove them before I merge it?

Sanne:
> Are the empty lines critical? Do you mind if I remove them before I merge it?

those files are generated by the "upgrade gradle" task .. I considered ignoring them, but then it will restore the same changes at the next round. So I don't think it's worth it.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/421

Thanks!

Issue #417, opened 2020-10-21
Test with Vertx SQL client 4


DavideD:
Should we create a branch upstream for [vert.x sql client 4](https://github.com/eclipse-vertx/vertx-sql-client/releases/tag/4.0.0-milestone5) and make it work with it?

gavinking:
That sounds like a good idea. And that way we can try adding some simple tests for SQL Server.

DavideD:
The new branch is now upstream:
https://github.com/hibernate/hibernate-reactive/tree/vert.x-4

gavinking:
> The new branch is now upstream:
> https://github.com/hibernate/hibernate-reactive/tree/vert.x-4

Excellent!

PR #425, opened 2020-10-28
#417 Update to Vert.x 4

DavideD:
This is not supposed to be merged.
The idea is to create a separate branch upstream calle `vert.x-4`.


DavideD:
I've removed the workflow for the latest Vert.x sql client 4 because it causes some error

gavinking:
@DavideD so Vert.x 4 has a whole different approach to to managing transactions?

DavideD:
It seems so, the Transaction is not a `SQLClient` anymore so you cannot return it like we did before.
To be fair, I put together this fix quite quickly and I have to give it a better look, but I think it could work for an initial branch

DavideD:
Also `beginConnection()` returns a `Future<Transaction>` instead of a `Transaction`. It makes sense to me but I'm wondering if we should should change our `SQLClientConnection#beginTransaction()` to return a `CompletionStage<Transaction>` instead of `CompletionStage<Void>`. Just a feeling. That's what I would expect by looking at the method for the first time.

But I'm surprise and quite happy to see that all the tests pass with this simple patch.

Sanne:
+1 to track Vertx.next , but remember to keep these dependencies in synch with what's in Quarkus.

DavideD:
Sure, we don't plan to include this in the main branch for now.

DavideD:
It's seems everybody is cool with this, I'm going to push it upstream

Sanne:
upstream? in its own branch?

DavideD:
Yes

DavideD:
Branch created upstream

PR #429, opened 2020-11-05
Test against JDK11 / 15 / 16

yrodiere:
This changes the Gradle configuration and Github workflow to test Hibernate Reactive with JDK 11 / 15 / 16 on each commit to master or pull request.

More specifically, in per-JDK jobs:

* We compile the project with JDK 11, but with `-release 8` so that Java 8 bytecode is produced.
* We run the tests with a JVM version 11, 15 or 16.

Also, this adds a few caches to builds, in the hope that jobs will be faster, since we're starting to have a lot of them.

Finally, this adds downloadable reports to the builds, to more easily get details about test failures.

Example run: https://github.com/yrodiere/hibernate-reactive/actions/runs/347693600

By the way, there is a test failure with JDK 15 and 16: `org.hibernate.reactive.types.BasicTypesAndCallbacksForAllDBsTest#testLocalDateType` fails with the following stack trace:

```
java.lang.AssertionError: Not equals : 2020-11-05T14:34:15.539050417 != 2020-11-05T14:34:15.539050
	at io.vertx.ext.unit.impl.TestContextImpl.reportAssertionError(TestContextImpl.java:362)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:258)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:245)
	at org.hibernate.reactive.types.BasicTypesAndCallbacksForAllDBsTest.lambda$testLocalDateTimeType$19(BasicTypesAndCallbacksForAllDBsTest.java:262)
```

It looks like timestamps returned from the database have microsecond precision whereas nanosecond precision is expected, for some reason. As to why it fails with JDK15 and not with JDK8 or 11... beats me. I didn't try to fix it, I assume you'll have a better idea of what's going on?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -185,7 +185,7 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Create maven package
+    - name: Create artifacts
       run: ./gradlew assemble
     - name: Detect the version of Hibernate Reactive
       id: detect-version
@@ -208,7 +208,7 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Create maven package
+    - name: Create artifacts
       run: ./gradlew assemble
     - name: Install SSH key
       uses: shimataro/ssh-key-action@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,12 @@ jobs:
       uses: actions/setup-java@v1
     - name: Run examples on ${{ matrix.db }}
       run: ./gradlew :example:runAllExamplesOn${{ matrix.db }}
+    - name: Upload reports (if build failed)
+      uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: reports-examples-${{ matrix.db }}
+        path: './**/build/reports/'
 
   test_dbs:
     name: Test with ${{ matrix.db }}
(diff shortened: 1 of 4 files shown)


yrodiere:
cc @Sanne 

I ended up not changing the default build at all: Gradle toolchains are only enabled with specific options on the CI. That's because toolchains are rather impractical for contributors, in particular: automatic detection of system JDKs doesn't work, meaning if you say you want to build with JDK11, Gradle will try to download JDK11, always. Even if you're already using JDK11 to run Gradle. That's easily worked around with a few options in the CI, but it's definitely not something you will want to do when you build the project manually.

DavideD:
> Gradle will try to download JDK11,

Do you mean at every build? Or just the first time?

yrodiere:
> Do you mean at every build? Or just the first time?

Just the first time, it's cached. But still, that's a 100+MB download for something you probably don't need (since you already have a JDK11 somewhere).

And to clarify, I made sure that it does not happen. I was just explaining why I didn't enable toolchains by default.

DavideD:
> And to clarify, I made sure that it does not happen. I was just explaining why I didn't enable toolchains by default.

Yeah, I got it. Thanks

Sanne:
Could we have CI test JDK8 as well?   I don't remember the details, but there were a couple of obscure situations that `-release 8` wouldn't be able to defend against.

DavideD:
I'm going to add JDK 8 to the build and send PR with the test fixed

DavideD:
Superseded by: https://github.com/hibernate/hibernate-reactive/pull/430

PR #430, opened 2020-11-05
Fix test failure on JDK 15+

DavideD:
Built on top of https://github.com/hibernate/hibernate-reactive/pull/429

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,15 @@ ext {
     println "Vert.x Version: " + project.vertxVersion
 
     baselineJavaVersion = JavaLanguageVersion.of( 8 )
+
+    // If the property is set, enable the toolchain and use the selected jdk.
+    // Note that when toolchain is enabled, you also need to specify
+    // the location of the selected jdk plus the location of jdk 11
+    // (auto-download and auto-detect are disabled in gradle.properties).
+    //
+    // Example (with SDKMAN):
+    // ./gradlew build -Ptest.jdk.version=15 \
+    // -Porg.gradle.java.installations.paths=$SDKMAN_CANDIDATES_DIR/java/15.0.1-open,$SDKMAN_CANDIDATES_DIR/java/11.0.7-open
     if ( project.hasProperty( 'test.jdk.version' ) ) {
         // Testing a particular JDK version
         // Gradle doesn't support all JDK versions unless we use toolchains
--- a/build.gradle
+++ b/build.gradle
@@ -59,10 +59,32 @@ ext {
     }
 
     testcontainersVersion = '1.14.3'
-    baselineJavaVersion = 1.8
 
     println "Hibernate ORM Version: " + project.hibernateOrmVersion
     println "Vert.x Version: " + project.vertxVersion
+
+    baselineJavaVersion = JavaLanguageVersion.of( 8 )
+    if ( project.hasProperty( 'test.jdk.version' ) ) {
+        // Testing a particular JDK version
+        // Gradle doesn't support all JDK versions unless we use toolchains
+        javaToolchainEnabled = true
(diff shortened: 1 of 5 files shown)


DavideD:
OK,  adding the JDK 8 is not as straightforward as I thought.
I will have a better look tomorrow

yrodiere:
@DavideD @Sanne The default build already uses JDK8. I don't see the point of adding it again in the JDK matrix?

yrodiere:
@DavideD Also, it looks like you're having trouble to add JDK8 because... for some reason Gradle is producing Java 11 bytecode. I'm investigating.

yrodiere:
@DavideD Got it. I fixed the problem and added a build that tests with JDK8 to the JDK matrix. It's probably a waste of resources, but I suppose we've got plenty at the moment :)

yrodiere:
(@DavideD Beware: I force-pushed to your branch)

DavideD:
@yrodiere That won't work

DavideD:
OK, forget what I said. 

My comment refers to the fact that I'm sure yesterday I've sent the same change and the build failed.
But it was late and I suspect my commit wasn't exactly the same.

Thanks @yrodiere

yrodiere:
@DavideD As discussed, I pushed a few more commits:

1. When testing against JDK15 for example, we don't just run tests with JDK15: we also compile tests with JDK15 and use `-release 15`. That way, we really simulate an application using JDK15 for everything, while relying on our own binaries built for JDK8.
2. When testing against JDK 11 / 15 / 16, we now build the main code with JDK 8, not 11. That way, the produced binaries are identical to what we will actually release.
3. Simplify the Gradle code. Not strictly necessary, but I think it's clearer that way, and more importantly it will be easier to adapt in the future. That's important because I intend to port this code to ORM (next week).

I think @Sanne will be pleased with item 2 in particular.

build.gradle
```diff
@@ -103,9 +113,29 @@ subprojects {
     version = projectVersion
 
     if ( project.javaToolchainEnabled ) {
-        compileJava.options.release = project.baselineJavaVersion.asInt()
-        compileTestJava.options.release = project.baselineJavaVersion.asInt()
-        compileJava.options.encoding = 'UTF-8'
+        // "sourceCompatibility" is not supported with toolchains. We have to work around that limitation.
+        compileJava.configure {
+            if (project.compileJdkVersion.asInt() < 11) {
```
DavideD: Wasn't `release` introduced in 9?

build.gradle
```diff
@@ -59,20 +59,103 @@ ext {
...
+            } else {
+                options.release = project.javaVersions.test.release.asInt()
+            }
+            options.encoding = 'UTF-8'
+        }
+    }
+    else {
+        sourceCompatibility = JavaVersion.toVersion( project.baselineJavaVersion )
+        targetCompatibility = JavaVersion.toVersion( project.baselineJavaVersion )
+        compileJava.options.encoding = 'UTF-8'
```
DavideD, line 157: Can't we just set this option once before the initial if?

DavideD:
LGTM

DavideD:
Really nice

build.gradle
yrodiere: It was. I added a commit. We don't test Java 9 / 10 though: they are not supported anywhere AFAIK.

build.gradle
yrodiere: Right. I amended my last commit and force-pushed.

build.gradle
DavideD: :)
Thanks, I was just checking, I didn't mean that you should fix it.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/430

Sanne:
this is great! thanks again

Issue #399, opened 2020-10-04
tests stopped working on Docker desktop


gavinking:
I've no idea what's going on here, but it seems to have coincided with me updating Docker desktop on Mac to the latest version:

```java
Gradle Test Executor 2 > org.hibernate.reactive.FilterTest > testFilterCollectionFetch FAILED
    org.testcontainers.containers.ContainerLaunchException: Container startup failed
        at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:330)
        at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:311)
        at org.hibernate.reactive.containers.PostgreSQLDatabase.getJdbcUrl(PostgreSQLDatabase.java:39)
        at org.hibernate.reactive.containers.DatabaseConfiguration.getJdbcUrl(DatabaseConfiguration.java:78)
        at org.hibernate.reactive.BaseReactiveTest.constructConfiguration(BaseReactiveTest.java:88)
        at org.hibernate.reactive.FilterTest.constructConfiguration(FilterTest.java:26)

        Caused by:
        org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=postgres:13.0, imagePullPolicy=DefaultPullPolicy())
            at org.testcontainers.containers.GenericContainer.getDockerImageName(GenericContainer.java:1279)
            at org.testcontainers.containers.GenericContainer.logger(GenericContainer.java:613)
            at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:320)
            ... 5 more

            Caused by:
            java.lang.IllegalStateException: Can not connect to Ryuk at localhost:32768
                at org.testcontainers.utility.ResourceReaper.start(ResourceReaper.java:176)
                at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:168)
                at org.testcontainers.LazyDockerClient.getDockerClient(LazyDockerClient.java:14)
                at org.testcontainers.LazyDockerClient.listImagesCmd(LazyDockerClient.java:12)
                at org.testcontainers.images.LocalImagesCache.maybeInitCache(LocalImagesCache.java:68)
                at org.testcontainers.images.LocalImagesCache.get(LocalImagesCache.java:32)
                at org.testcontainers.images.AbstractImagePullPolicy.shouldPull(AbstractImagePullPolicy.java:18)
[...]

PR #431, opened 2020-11-06
disable ryuk for #399

gavinking:
Should we do this?

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -80,6 +80,7 @@ tasks.withType(Test) {
 // Example:
 // gradle test -Pdb=MySQL
 test {
+    environment "TESTCONTAINERS_RYUK_DISABLED", "true"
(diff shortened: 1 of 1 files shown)


DavideD:
I don't mind, if it makes life easier to some people I guess we can include it.

But I don't understand if this is a testcontainers bug. If it is, won't it be solved at some point?

gavinking:
> But I don't understand if this is a testcontainers bug. If it is, won't it be solved at some point?

I believe it is, it when I looked it seemed like it will be .... but .... why do we even need this daemony thing if test containers works perfectly fine without it?

bsideup:
I would advice against it (you don't want to have dangling containers after running the tests, do you? :)).

There was a behaviour change in one of Docker for Mac's releases that we fixed in Testcontainers 1.15.0, so I would suggest upgrading the dependency instead.

gavinking:
Fine, then lets do #435 instead.

Issue #441, opened 2020-11-16
add access to SynchronizeableQuery operations


gavinking:
Perhaps our `Query` interface could just extend `SynchronizeableQuery`. (Need to look more closely.)

See discussion on https://github.com/hibernate/hibernate-orm/pull/3643.

gavinking:
Hrmph, I'm really not sure what to do about this one. We don't have a `NativeQuery` interface at present, and the operations of `SynchronizeableQuery` don't really make sense for HQL queries, so I would really prefer not to pollute `Query`.

There's a couple of options:

- Create a `NativeQuery` interface in `Stage` and `Mutiny` that mixes `SynchronizeableQuery` with `Query`. This is fine I guess, though I did want to avoid it and have just one sort of `Query`.
- Add a variadic parameter to each of the `createNativeQuery()` methods. I sorta lean towards this option, but it makes it difficult to add any additional overloads to these methods later.

gavinking:
> a variadic parameter

Or perhaps a parameter of type `String[]` would be better.

gavinking:
> > a variadic parameter
> 
> Or perhaps a parameter of type `String[]` would be better.

Or perhaps I should buy into my own bullshit and add a `QuerySpaces` class.

gavinking:
> Or perhaps I should buy into my own bullshit and add a `QuerySpaces` class.

It's called `AffectedEntities` FTR.

PR #444, opened 2020-11-16
overload createNativeQuery() to accept query spaces

gavinking:
Here's a proposed solution to #441.

Not sure how I feel about it.

WDYT?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
@@ -133,4 +133,6 @@ public interface ReactiveQuery<R> {
 	ReactiveQuery<R> setCacheRegion(String cacheRegion);
 
 	String getCacheRegion();
+
+	ReactiveQuery<R> setQuerySpaces(String[] querySpaces);
 }
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/AffectedEntities.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/AffectedEntities.java
@@ -1,0 +1,63 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.common;
+
+import org.hibernate.Incubating;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.internal.util.collections.ArrayHelper;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A description of the entities and tables affected by a native query.
+ * <p>
+ * The entities and tables given by an instance passed to
+ * {@code createNativeQuery()} must be synchronized with the database
+ * before execution of the query.
+ *
+ * @author Gavin King
+ */
+@Incubating
+public class AffectedEntities {
+    private static final Class<?>[] NO_ENTITIES = new Class[0];
+    private static final String[] NO_TABLES = new String[0];
+
+    private final String[] queryTables;
+    private final Class<?>[] queryEntities;
+
+    public AffectedEntities(Class<?>... queryEntities) {
+        this.queryTables = NO_TABLES;
+        this.queryEntities = queryEntities;
+    }
+
+    public AffectedEntities(String... queryTables) {
+        this.queryTables = queryTables;
+        this.queryEntities = NO_ENTITIES;
+    }
+
+    public String[] getAffectedTables() {
+        return queryTables;
+    }
+
+    public Class<?>[] getAffectedEntities() {
+        return queryEntities;
+    }
+
+    public String[] getAffectedSpaces(SessionFactoryImplementor factory) {
+        List<String> spaces = new ArrayList<>();
+        for ( String table: getAffectedTables() ) {
+            spaces.add( table );
+        }
+		for ( Class<?> entity: getAffectedEntities() ) {
+			Serializable[] querySpaces = factory.getMetamodel().entityPersister(entity).getQuerySpaces();
+			spaces.addAll( Arrays.asList( (String[]) querySpaces) );
+		}
+        return spaces.toArray(ArrayHelper.EMPTY_STRING_ARRAY);
+    }
+}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -17,6 +17,7 @@ import org.hibernate.collection.internal.AbstractPersistentCollection;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.reactive.common.AffectedEntities;
 import org.hibernate.reactive.common.AutoCloseable;
 import org.hibernate.reactive.common.ResultSetMapping;
 import org.hibernate.reactive.session.ReactiveSession;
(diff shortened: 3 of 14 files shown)


Sanne:
Could we use the `String... querySpaces` rather than `String[] querySpaces` parameter, so that we make it optional to use?

gavinking:
Well that was my first thought but then I didn't like it...

gavinking:
> Well that was my first thought but then I didn't like it...

The issue is that it puts each query space on the same level (and same type) as the query string. And then they look like similar things.

Sanne:
BTW @blafond feel free to join the brainstorming if you have ideas :)

gavinking:
> Wondering if, rather than Strings we should accept something more strongly typed? e.g. a "NativeQueryDefinition", which could (optionally) be used to set some additional future hints - on top of query spaces.

Yes, I agree that we should strongly consider this option. We already have too many overloads here.

gavinking:
I've introduced a rough draft of a `NativeQuery` object.

It's amazing to me just how bad Java is for these sort of simple objects. I wanted to just use constructors, but couldn't because type inference just doesn't work properly with Java constructors. For this sort of thing Ceylon is *sooooo* much superior.

gavinking:
@Sanne do you have any feedback?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
```diff
@@ -0,0 +1,87 @@
...
+        this.queryTables = queryTables;
+        this.queryEntities = queryEntities;
+        this.returnType = returnType;
+    }
+
+    public static <T> NativeQuery<T> create(String queryString, ResultSetMapping<T> resultSetMapping) {
+        return create(queryString, resultSetMapping, NO_TABLES);
+    }
+
+    public static <T> NativeQuery<T> create(String queryString, ResultSetMapping<T> resultSetMapping, String[] queryTables) {
```
Sanne, line 44: Now it might be more suitable to use `...` ?

It feels like you purposefully introduced the mandatory `ResultSetMapping` parameter (or `Class<T> returnType` below) to make it viable?

Sanne: Alternatively, I'd use a builder DSL.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
```diff
@@ -121,4 +121,6 @@
 	ReactiveQuery<R> setCacheRegion(String cacheRegion);
 
 	String getCacheRegion();
+
+	ReactiveQuery<R> setQuerySpaces(String[] querySpaces);
```
Sanne, line 125: Shouldn't we expose this only on native queries?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
gavinking: > Now it might be more suitable to use `...` ?

I agree, and I considered it, but I'm just wondering if it might get in the way of adding additional overloads in future. TBH I'm just not sure.

> Alternatively, I'd use a builder DSL.

I have two objections to that:

1. a kinda principled objection to using a mutable object when an immutable object will do the job, and
2. also the practical objection to how verbose that's going to get: it's already quite difficult to format the code for a native query call in HR with all the surrounding anonymous functions..

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
gavinking: Um, in principle yes, but we don't have a `ReactiveNativeQuery` type at present. (Notice that this is an internal API, not one that is visible to users.)

Sanne: ha, we have some "soft conventions" about what's API. Pretty much anything which isn't explicitly hidden in a "spi" or "impl"/"internal" package is being treated as API in other projects.

That's just how it is on other projects; I don't mean to impose any cargo-cult here, but that made me assume this is API and I suspect it will give the same impression to other users.

Do you think we could improve on this, or not worth it?  cc/ @DavideD

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
Sanne: I generally don't like DSL builder objects on the "hot path" from a performance perspective, but I'm thinking here that the expected usage is to define native queries as application constants.

So I could see myself liking the shape of:

	private static NativeQuery<Object[]> TITLES_AND_AUTHORS_QUERY = NativeQuery.fromSQL("select b.title, a.name from books b join authors a on author_id=a.id order by b.isbn")
		.withResultSetMapping(Object[].class) // Optional : Object[] as default return type for T of NativeQuery<T> ?
		.affectsTableSpaces("authors","books") // Optional
		.create();
 
But N.B. I'm somewhat cheating here, as I see the current API requires a Session to define & validate a `ResultSetMapping`. So maybe this requires a bit more work, but it has the merit of allowing us to push such validations at build time & also improve efficiency.

Ideally this could be assigned to a `Query<T>` rather than a `NativeQuery<T>` and could tie in with more advanced build time validations, or allow people to use a different builder.

> I agree, and I considered it, but I'm just wondering if it might get in the way of adding additional overloads in future. TBH I'm just not sure.

It's precisely to allow for future overloads that I'd go with a builder in this case.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
DavideD: I think for the most part that's what we are doing already. We might need to give it another review before releasing a CR.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
gavinking: > I generally don't like DSL builder objects on the "hot path" from a performance perspective, but I'm thinking here that the expected usage is to define native queries as application constants.

Hrm. I don't hate it.

> But N.B. I'm somewhat cheating here, as I see the current API requires a `Session` to define & validate a `ResultSetMapping`

I'm not actually happy with this, FTR, I actually want to have a code generator create type safe references to things like named queries and `ResultSetMapping`s.

> It's precisely to allow for future overloads that I'd go with a builder in this case.

Yesyes, I understand that. I like constructors for immutability. On the other hand, it *is* possible to make immutable builders, so perhaps we could do it that way.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
gavinking: @Sanne We have API packages and internal packages. I've been putting the APIs of internal packages in the package root, and the implementation in `.impl`. Objectively there's not a lot of difference just because we don't have non-public packages (yet) in Java.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
Sanne: Sounds like we're on the same page, but I'm not sure about the conclusion?

I think I'd go with the Builder DSL, not because it would make all such things possible now, but it would allow us to get there eventually.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
Sanne: We do have non-public packages if you embrace the module system.

gavinking: > We do have non-public packages if you embrace the module system.

Yes sure, I should have said "in the version of Java we're stuck on for now".

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/NativeQuery.java
gavinking: > I'm not sure about the conclusion?

Well I figured I would try it out and see what I think.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQuery.java
gavinking: @Sanne I've met you halfway. Check it out.

gavinking:
Alright, so after much sitting on this and thinking about it, I changed my mind and decided not to introduce a `NativeQuery` object. There's nothing wrong with method overloading here.

But I did introduce an `AffectedEntities` class to make it easier and more elegant to pass these things in.

gavinking:
I'm going to merge this because it's already been sitting around for too long, and I had to deal with conflicts. And I've already done about as much thinking about the API as is reasonable.

PR #456, opened 2020-11-26
Vertx 3 *and* 4 Compatibility

Sanne:
Some horribly looking reflection, but I think it's worth it - especially considering we'll want to drop Vertx 3 compatibility soon. 

WDYT?

Some close methods in Vert.x 4 now return a `io.vertx.core.Future` - we didn't notice this during previous experiments as it's source-compatible, but we catch the mistake now by reconfiguring Gradle to always build with Vertx 4 and then run tests (and tests only) with Vertx 3 on classpath. This trips as it's an incompatible change at bytecode level.

IMO the new close() methods need to be propagated, but it's hard to do so as long as we also want to keep Vert.x 3 compatibility... So I propose we merge this, and then it should be easier to start reviewin this aspect for a better Vert.x 4 integration.

[ developed in pair with @DavideD ]

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/impl/DefaultVertxInstance.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/impl/DefaultVertxInstance.java
@@ -5,7 +5,12 @@
  */
 package org.hibernate.reactive.vertx.impl;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
 import io.vertx.core.Vertx;
+
+import org.hibernate.AssertionFailure;
 import org.hibernate.reactive.vertx.VertxInstance;
 import org.hibernate.service.spi.Startable;
 import org.hibernate.service.spi.Stoppable;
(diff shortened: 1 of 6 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/impl/DefaultVertxInstance.java
```diff
@@ -34,7 +52,16 @@ public Vertx getVertx() {
     @Override
     public void stop() {
         if ( vertx != null ) {
-            vertx.close();
+            try {
+                final Object invoke = CLOSE_METHOD.invoke( vertx );
+                if ( invoke instanceof Future ) {
+                    Future f = (Future) invoke;
+                    f.toCompletionStage().toCompletableFuture().get();
```
Sanne, line 59: Looks like this is causing a deadlock on shutdown of Vert.x v.4

gavinking:
@Sanne please wait til we get #457 sorted out.

DavideD:
Closing this one, see https://github.com/hibernate/hibernate-reactive/pull/458

PR #457, opened 2020-11-26
proposed fix to Vertx 4 tx handling

gavinking:
@DavideD can you check that this is what we should be doing.


DavideD:
I think you are right.
I will see if I can add a test to check this case

DavideD:
I'm going to apply this directly to Sanne's pr

gavinking:
@DavideD merge this?

DavideD:
Rebased and merged

Thanks

gavinking:
Cheers.

Issue #440, opened 2020-11-12
Proposal: switching to Vertx 4 already


Sanne:
It seems Vert.x 4 is already in CR1, with a CR2 coming soon.

We're currently "stuck" on Vert.x v 3.x because we need to stay compatible with the Vert.x versions used in Quarkus; however:
 
- Quarkus plans to upgrade ASAP in 2021, expecting migration completion by March (but this very issue is meant to simplify that work)
- It seems we only need changes in a single class in Hibernate Reactive to be compatible with both
- It would be odd for the Hibernate Reactive community to impose a major upgrade of such a crucial dependency so close to our own 1.0 release.

So, I'd like us to explore being able to "baseline" on Vert.x 4 already: switch the dependency to v4, and document it as the recommended version, use it in examples - while applying some internal magic to keep runtime compatibility with v3 just for a bit longer, only to satisfy integration needs in Quarkus for as long as Quarkus doesn't upgrade as well.

I've already discussed this in chat with @DavideD and he thought this shouldn't be too hard - pending exploration.

This would also allow us to release with the new drivers (MSSQL support?), as I'm assuming we can keep them out of the Quarkus tree.

Sounds good? Asking @gavinking and @blafond in particular, as I already chatted with Davide.

I hope you like the idea, as it would take some barriers away from our velocity, and reduce the complexity of the Vertx upgrade in Quarkus - so helping them as well.

gavinking:
I'm skeptical that this is worth the effort. March is really not far away and there are other things we could spend time on.

Sanne:
I think it's a small investment and will reduce the total volume of work.

So unless you're highly preferring to release with Vertx 3 first, I'd like to give it a shot - we can abort if it gets unnecessarily complicated.

gavinking:
Well I mean I gotta be honest I'm just not really keen on seeing reflection-based code added to the codebase if we're just going to throw it away in four months time when Vert.x 4 is released.

Sanne:
oh I totally agree, but it's only a couple of methods and we'll get rid of them soon as well.

Sanne:
Closing this as it turns out it was a bad idea: too complex.

PR #458, opened 2020-11-27
[#440] Compatibility with Vert.x 3 and 4

DavideD:
Follows https://github.com/hibernate/hibernate-reactive/pull/456

This PR is built on top of @Sanne's one.
I've included @gavinking's [fix for transaction](https://github.com/hibernate/hibernate-reactive/pull/457) in this commit: https://github.com/hibernate/hibernate-reactive/commit/caf34eac71ba77c7c313e221262980b7d1e2e53b

I've also added a minor clean up and an update to the workflows.
The workflows now will test with Vert.x 3.9.4 and 4.0.0.CR2

It's a draft because I need to wait for the new GitHub workflows to complete.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
@@ -31,7 +31,6 @@ import io.vertx.sqlclient.SqlResult;
 import io.vertx.sqlclient.Transaction;
 import io.vertx.sqlclient.Tuple;
 
-import static org.hibernate.reactive.util.impl.CompletionStages.nullFuture;
 import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
 
 /**
@@ -46,7 +45,7 @@ public class SqlClientConnection implements ReactiveConnection {
 
 	private final Pool pool;
 	private final SqlConnection connection;
-	private CompletionStage<Transaction> transaction;
+	private Transaction transaction;
 
 	SqlClientConnection(SqlConnection connection, Pool pool,
 						SqlStatementLogger sqlStatementLogger,
@@ -251,15 +250,10 @@ public class SqlClientConnection implements ReactiveConnection {
 
 	@Override
 	public CompletionStage<Void> beginTransaction() {
-		if (RUNNING_ON_VERTX3) {
-			transaction = CompletionStages.completedFuture( beginTransactionVertx3() );
-		}
-		else {
-			transaction = Handlers
-				.toCompletionStage( handler -> connection.begin( handler ) );
-		}
-
-		return voidFuture();
+		CompletionStage<Transaction> transactionStage = RUNNING_ON_VERTX3
+				? CompletionStages.completedFuture( beginTransactionVertx3() )
+				: Handlers.<Transaction>toCompletionStage( connection::begin );
+		return transactionStage.thenAccept( tx -> transaction = tx );
 	}
 
 	private Transaction beginTransactionVertx3() {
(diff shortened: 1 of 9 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -199,45 +204,87 @@ private void feedback(String sql) {
...
 
 	@Override
 	public void close() {
-		connection.close();
+		if (RUNNING_ON_VERTX3) {
+			try {
+				CLOSE_ON_3.invoke( connection );
+			}
+			catch (IllegalAccessException | InvocationTargetException e) {
+				throw new AssertionFailure( "Error closing connection on Vert.x 3 compatibility mode. Probably best to upgrade Vert.x to version 4", e );
```
gavinking, line 282: Is there any likelihood that upgrading to Vert.x 4 is really a solution to whatever problem is causing this failure? I think this error message is most likely unhelpful.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -199,45 +204,87 @@ private void feedback(String sql) {
...
+				: Handlers.<Transaction>toCompletionStage( connection::begin );
+		return transactionStage.thenAccept( tx -> transaction = tx );
+	}
+
+	private Transaction beginTransactionVertx3() {
+		try {
+			return (Transaction) BEGIN_ON_3.invoke( connection );
+		}
+		catch (IllegalAccessException | InvocationTargetException e) {
+			throw new AssertionFailure( "Error beginning transaction on Vert.x 3 compatibility mode. Probably best to upgrade Vert.x to version 4", e );
```
gavinking, line 259: Same here.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -199,45 +204,87 @@ private void feedback(String sql) {
 	}
 
 	private SqlClient client() {
-		return transaction != null ? transaction : connection;
+		return connection;
```
gavinking, line 207: Confused: how can this change be right?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Shouldn't these exceptions only occurs if the `close()` method doesn't exist or is not accessible?

If it doesn't exist upgrading should solve the issue.
If the problem is an IllegalAccess I'm not sure the fact that we also show the cause should help.

Do you have any suggestions about what to add?
Would something like this be better?
```
SqlClientConnection#close() method not found or not accessible. Check you are using a supported Vert.x version (3 or 4)
```

gavinking: I don't think they can possibly occur if `close()` doesn't exist, since we would not have been able to obtain the method ref.

They can only occur if the `close()` methods throw, it seems to me.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Gosh.. you are right.
Not seeing any failures in the tests made me beleive that everything was working.
I will work on it

DavideD: Interesting... we actually some failures but only with DB2

DavideD:
@Sanne @gavinking In the end I don't think this solution will work.

It seems there are some failures related to how Vert.x handle transactions when we mix the two libraries.
We already have a working branch with Vert.x 4 so it's probably easier to release two separate versions of Hibernate Reactive for the time being.

Sanne:
Fair enough, let's not try to have compatibility for both. I was clearly mistaken int thinking it would be easier.

And yet I think this has been useful; the change of API for close() methods was "source compatible" but not really compatible, we wouldn't have spotted the semantic difference without trying to run this in this way.

It's actually quite worrying in termx of Vert.x API evolution; if the change is this significant, not getting a compilation error makes it hard to get awareness. I think a different method name - triggering a compilation error - would have been better.

Sanne:
BTW invoking close(), and then block to wait on the returned Future gets you into a deadlock, so we need to figure out what's the right way to shutdown on vertx 4.

gavinking:
Well, we can just ask @vietj, right?

gavinking:
Should we close this? We can always reopen it later....

DavideD:
Yes, I will send a new PR with the updates required to the release process

PR #462, opened 2020-12-01
fix doc for getSingleResult() and add getSingleResultOrNull()

gavinking:
Proposed `Query.getSingleResultOrNull()` method.


gavinking:
Apparently I forgot the semantics of `getSingleResult()` :-)

gavinking:
Feedback on this change?

DavideD:
Looks good to me.
Shouldn't also test that we throw the expected exceptions?

DavideD:
> Apparently I forgot the semantics of getSingleResult() :-)

So I'm not the only one :)
I had the same reaction with ORM not so long ago

gavinking:
> So I'm not the only one :)

It's easily forgettable, since it's different to the semantics of `find()`.

Issue #316, opened 2020-08-11
Consider exposing a single method to open Sessions


Sanne:
Today on SessionFactory we're exposing both options:

- `Uni<Session> openSession()`
- `Session createSession()`

And a similar duality is proposed for the `StatelessSession` opening API (pr #310)

I would suggest exposing a single method only:

    Session openSession()

For various reasons:

## Simplicity

I don't believe there's a good reason to expose both

## `openSession` is the better name

It's the same as people are familiar with from ORM "classic", and it reminds that the `Session` resource needs to be closed.

## Implementation flexibility

Different containers will require different connection release strategies, so I'd rather leave this as an implementation detail of the integration code than something that is exposed on API.

I believe that, while we might need some degree of internal flexibility to accommodate for needs of various integration platforms / containers, we'll never need to open a connection immediately so there should be no need to have a blocking version of this method which requires returning a promise.
(Or, if some exotic container is still requiring this, they should fix it).

# Optimal strategy

For Quarkus today we're using the org.hibernate.ConnectionReleaseMode of `AFTER_STATEMENT`, which is not the most efficient but scales well and is a safe choice on many containers (including WildFly), and is therefore the default in Hibernate ORM.

But we have plans to switch to `BEFORE_TRANSACTION_COMPLETION`, a newly introduced policy (which still isn't production ready) as it will be much more efficient, especially in combination with our preferred connection pool Agroal.

I'm not sure if we'll want to use the same default in Reactive, but take it as an example of why leaving ourselves options open could be useful.

gavinking:
So this is something that was introduced in 33a25d0b7d36c3af19de0eb61d27f3344ccd6a4b and bcdb2550d91a966f7f9339ccb694a6f0921b2c0b when I added support for transactions.

What happened was that starting a session now meant obtaining a `SqlConnection` (instead of just using the `SqlClientPool`), which is a non-blocking operation in Vert.x.

So at first I just changed `openSession()` to return the new `Session` in a non-blocking fashion.

But then @aguibert complained about the usability of that, so to fix it, I added a new `createSession()` operation that uses a `ProxyConnection` which obtains its `Connection` the first time we call a non-blocking operation on the `ReactiveConnection`. Now, I didn't love this design at all because `ProxyConnection` works by introducing a new link in the chain of `CompletionStage`s for every operation on a  `ReactiveConnection`. 

So that's why we have two operations: to provide a way of avoiding all these extra `CompletionStage`s if necessary.

[...]

PR #476, opened 2020-12-08
[#475] Fix race condition in ProxyConnection.withConnection()

### LINKED ISSUE #475: Race condition in ProxyConnection
`org.hibernate.reactive.pool.impl.ProxyConnection#withConnection` often triggers `IllegalStateException( "session is currently connecting to database" )` in an application using Quarkus/Mutiny.

Code snippet that triggers the exception:

```
class Foo {
    @Inject
    Mutiny.Session mutinySession;

    @Path("some-endpoint")
    @GET
    Uni<String> someEndpoint() {
        Uni<String> work = mutinySession.find(SomeEntity.class, someId)
                .flatMap(x -> x.toString());
        return mutinySession.withTransaction(tx -> work);
    }
}
```

The pseudo-code above is simplified a lot but shall illustrate the problem. We can neglect concurrency aspects here, because the problem is that both the `Mutiny.Session.find()` and the `Mutiny.session.withTransaction()` trigger `org.hibernate.reactive.pool.impl.ProxyConnection#withConnection`.

In many cases, the first call to `withConnection()` might trigger the creation of a new connection (sets `connected=true`) and the second call to `withTransaction()` hits the `connected == true && connection == null` case resulting in the `IllegalStateException`.

The problem is related to the state management in `withConnection()`:
* The first invocation sets `connected=true` and triggers `sqlClientPool.getConnection()`, which is likely async and can run for some time.
* Any other invocation (same thread or any other thread) of `withConnection()` that arrives before the `getConnection`-stage completed will see `connected==true && connection==null` and trigger the `IllegalStateException`

Another issue is that the fields used to track the state (`closed`, `connection` are not declared `volatile`/guarded by appropriate fences).

I'll come up with a patch in a few minutes.

snazy:
In many cases, the first call to `withConnection()` might trigger the creation of a new connection (sets `connected=true`) and the second call to `withTransaction()` hits the `connected == true && connection == null` case resulting in the `IllegalStateException`.

The problem is related to the state management in `withConnection()`:
* The first invocation sets `connected=true` and triggers `sqlClientPool.getConnection()`, which is likely async and can run for some time.
* Any other invocation (same thread or any other thread) of `withConnection()` that arrives before the `getConnection`-stage completed will see `connected==true && connection==null` and trigger the `IllegalStateException`

Another issue is that the fields used to track the state (`closed`, `connection` are not declared `volatile`/guarded by appropriate fences).

This fix ensures that all invocations of `withConnection()` do either return the already borrowed `ReactiveConnection` in a "happy hot code path" or "queue up" the operations in a single `CompletionStage` that borrows the connection.

Fixes #475

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
@@ -34,7 +34,7 @@ final class ProxyConnection implements ReactiveConnection {
 		this.tenantId = tenantId;
 	}
 
-	private <T> CompletionStage<T> withConnection(Function<ReactiveConnection, CompletionStage<T>> operation) {
+	<T> CompletionStage<T> withConnection(Function<ReactiveConnection, CompletionStage<T>> operation) {
 		if ( !connected ) {
 			connected = true; // we're not allowed to fetch two connections!
 			CompletionStage<ReactiveConnection> connection =
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
@@ -20,9 +20,17 @@ import org.hibernate.reactive.pool.ReactiveConnectionPool;
 final class ProxyConnection implements ReactiveConnection {
 
 	private final ReactiveConnectionPool sqlClientPool;
-	private ReactiveConnection connection;
-	private boolean connected;
 	private final String tenantId;
+
+	// 'connection' must be volatile, because it can be read from a different thread
+	// than the one that writes this field - and read against this field are _not_
+	// guarded with a 'synchronized(stateChange)'.
+	private volatile ReactiveConnection connection;
+
+	// Helper fields to track the state of this proxy-connection.
+	private CompletionStage<ReactiveConnection> connectionCompletionStage;
+	private boolean closed;
+	private final Object stateChange = new Object();
 
 	public ProxyConnection(ReactiveConnectionPool sqlClientPool) {
 		this.sqlClientPool = sqlClientPool;
@@ -35,21 +43,81 @@ final class ProxyConnection implements ReactiveConnection {
 	}
 
 	<T> CompletionStage<T> withConnection(Function<ReactiveConnection, CompletionStage<T>> operation) {
-		if ( !connected ) {
-			connected = true; // we're not allowed to fetch two connections!
-			CompletionStage<ReactiveConnection> connection =
-					tenantId == null ? sqlClientPool.getConnection() : sqlClientPool.getConnection( tenantId );
-			return connection.thenApply( newConnection -> this.connection = newConnection )
-					.thenCompose( operation );
+		ReactiveConnection conn = connection;
+		if (conn != null) {
+			// Happy hot path: the connection's present and it can be used. When 'connection'
+			// is not null, we're good to use it.
+			//
+			// A non-null 'connection' object can "leak" here, even if 'closed==true' in a
+			// rare situation described below.
+			// It feels fine to accept this exception, because 'close()' should only be
+			// called when all operations have finished. And there's also a 2nd "line of
+			// defense": the 'ReactiveConnection' itself is also closed and should error out
+			// in that case.
+			// There's also no guarantee that all consumers of 'connection' have finished when
+			// 'close()' is being called.
+			//
+			// 		Thread A						Thread B
+			//			withConnection():
+			//			read 'connection' field
+			//
+			//			(thread suspended)				close() runs and finishes
+			//
+			//			runs operation.apply(conn)
+			//
+			return operation.apply(conn);
 		}
-		else {
-			if ( connection == null ) {
-				// we're already in the process of fetching a connection,
-				// so this must be an illegal concurrent call
-				throw new IllegalStateException( "session is currently connecting to database" );
+
+		// No connection in the 'connection' field present.
+		//
+		// This can mean:
+		// 1. this ProxyConnection has been closed (duh)
+		// 2. no connection has been acquired from 'sqlClientPool'
+		// 3. a connection is currently being acquired from 'sqlClientPool'
+
+		// Tricky part comes next...
+		// Either no connection has been acquired from the 'sqlClientPool' yet or a connection
+		// is currently being acquired. Need to distinguish both cases and prevent that more than
+		// one connection's being acquired from 'sqlClientPool'.
+		//
+		// Generally, the CompletionStage returned from 'sqlClientPool.getConnection()' is used
+		// to "bundle" all outstanding 'operation's against the not-yet-available connection.
+		//
+		// Note: the 'connectionCompletionStage' is never cleared to 'null', because otherwise
+		// a new race condition might be introduced. However, it is set to null, in 'close()',
+		// because that's also guarded via 'stateChange'.
+		//
+		CompletionStage<ReactiveConnection> completionStage;
+		synchronized (stateChange) {
+			if (closed) {
+				// easy case, just bark and bite
+				throw new IllegalStateException("Proxy connection already closed");
 			}
-			return operation.apply( connection );
+
+			completionStage = connectionCompletionStage;
+			if (completionStage == null) {
+				// First one getting here for this ProxyConnection instance.
+				// Start acquiring a connection from 'sqlClientPool' and store the
+				// CompletionStage instance for future use.
+				CompletionStage<ReactiveConnection> connection =
+						tenantId == null ? sqlClientPool.getConnection() : sqlClientPool.getConnection(tenantId);
+				completionStage = connection.thenApply(newConnection -> {
+					// Again, change to 'connection' guarded via 'stateChange'.
+					synchronized (stateChange) {
+						this.connection = newConnection;
+					}
+					return newConnection;
+				});
+				connectionCompletionStage = completionStage;
+			}
 		}
+
+		// "Queue up" the 'operation'.
+		// If the connection's already available, the 'operation' will likely be executed
+		// via this call site.
+		// If the connection's not yet available, the 'operation' will be executed once
+		// the 'completionStage' to acquire the connection has finished.
+		return completionStage.thenCompose(operation);
 	}
 
 	@Override
@@ -133,9 +201,18 @@ final class ProxyConnection implements ReactiveConnection {
 
 	@Override
 	public void close() {
-		if ( connection != null ) {
-			connection.close();
-			connection = null;
+		ReactiveConnection conn = connection;
+		try {
+			synchronized (stateChange) {
+				closed = true;
+				connectionCompletionStage = null;
+				connection = null;
+			}
+		}
+		finally {
+			if (conn != null) {
+				conn.close();
+			}
 		}
 	}
 }
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
@@ -7,7 +7,9 @@ package org.hibernate.reactive.pool.impl;
 
 import java.sql.ResultSet;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import org.hibernate.reactive.pool.ReactiveConnection;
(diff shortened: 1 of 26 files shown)


DavideD:
Thanks @snazy, I will have a look in detail tomorrow.
Would you be able to also provide a testcase?

You could use one of our existing test case for inspiration: https://github.com/hibernate/hibernate-reactive/tree/master/hibernate-reactive-core/src/test/java/org/hibernate/reactive

gavinking:
Hmm, this is indeed an interesting one.

I guess we wrote this code under the assumption that the pool was only being used from within a single event loop. But I guess we also need to take into account that you can have multiple event loops in Vert.x.

Is that the case in your environment, @snazy?

gavinking:
@vietj @FroMage @cescoffier @Sanne I'm a little embarrassed to admit that this was something we have not really put any thought into. Is the Vert.x connection `Pool` written to allow multithreaded access, and exactly what are the restrictions?

snazy:
Not really, no.
It's Quarkus (with Vert.x).

The issue at play is pretty much this, where the actual work to get the proxied-DB connection is ran asynchronously.
```
  Thread A                                      Thread B

    mutinySession.find()
    --> ProxyConnection.withConnection()
        triggers getConnection()                --> connction to DB opened async...
        sets connected=true
        

    mutinySession.withTransaction()
    --> ProxyConnection.withConnection()
        sees connected==true && connnection==null


                                                <-- finishes get-connection request

        connection=XYZ
```

gavinking:
Wait, I'm reviewing the code, and it doesn't look that bad. Unless I'm mistaken, an instance of `ProxyConnection` always belongs to a particular `Session`. And `Session`s certainly aren't multithreaded, not in even in Hibernate ORM, even less so in reactive.

gavinking:
> Not really, no.
> It's Quarkus (with Vert.x).

> The issue at play is pretty much this, where the actual work to get the proxied-DB connection is ran asynchronously.

@snazy but you shouldn't be using a single `Session` from more than one unit of work, let alone from more than one event loop.

snazy:
Right, I'm not sharing a `Mutiny.Session` between requests. But that doesn't mean that whatever's executing the individual lambdas (not sure whether there's a better term for the individual units of work or operations or whatever - the things you chain together via `.flatMap()` et al) doesn't schedule those onto other threads.

gavinking:
> The problem is related to the state management in `withConnection()`:
> * The first invocation sets `connected=true` and triggers `sqlClientPool.getConnection()`, which is likely async and can run for some time.
> * Any other invocation (same thread or any other thread) of `withConnection()` that arrives before the `getConnection`-stage completed will see `connected==true && connection==null` and trigger the `IllegalStateException`

The second access to the `Session` should happen from within the same reactive stream as the first. You can't just go calling additional methods of the session before the first is completed. Just like in Hibernate ORM, `Session` is single-threaded, except that now a "thread" is a reactive stream, not a literal Java `Thread`.

snazy:
Mean, it's okay for a reactive framework to juggle the individual pieces of work that form a whole request.

I've seen many of these `IllegalStateException` mentioned above and also `IllegalStateException( "Session/EntityManager is closed" )` from `org.hibernate.internal.AbstractSharedSessionContract#checkOpen` before this change. Both no longer happen in my app with the patch applied.

gavinking:
> that doesn't mean that whatever's executing the individual lambdas (not sure whether there's a better term for the individual units of work or operations or whatever - the things you chain together via `.flatMap()` et al) doesn't schedule those onto other threads.

This shouldn't matter, as long as the access to the `Session` all happens "within the chain" (not sure if the right language to use here).

snazy:
> The second access to the Session should happen from within the same reactive stream as the first. You can't just go calling additional methods of the session before the first is completed. Just like in Hibernate ORM, Session is single-threaded, except that now a "thread" is a reactive stream, not a literal Java Thread.

Yep. Both calls to `mutinySession` happen from the same thread in my example. But the actual work that `getConnection()` does is triggered onto another thread. So the "callback" from `getConnection` into `withConnection` happens from "somewhere else" - and that's why the 2nd call to `mutinySession` sees `connected==true && connection==null`.

snazy:
> This shouldn't matter, as long as the access to the Session all happens "within the chain" (not sure of the right language to use here).

Yep. That's what I've assumed and it seems to work: I've got a bunch of quite complex and chained and dynamically constructed "chains" with both `Uni` and `Multi` ... and all requests I've tested so far work fine.

gavinking:
@snazy, we're going to need a reproducer test. Because in my opinion, the links of the chain really are supposed to be executed in sequence (and not concurrently), even with multiple event loops.

snazy:
Sure - let me code a test case.

From a say "functional point of view" (at least as far as I understood all that Vert.x stuff), it's perfectly fine to construct two `Uni`s (say the `Session.withTransaction` and `Session.find`) and plumb those together at a later point in time, because actual work shall start when something subscribes to these `Uni`s and `Multi`s. If I understand that correctly, the connection should be acquired when something subscribes and not when `withTransaction` or `find` run. But I suspect, that would be a bigger refactoring.

Sanne:
I'll wait for the reproducer (and then let @DavideD look as he volunteered first), but just a thought: I don't think this bug needs multiple threads to be triggered; just sequentially using the `ProxyConnection` in a way that we didn't expect could trigger the issue.

Maybe it's enough to reset the `connected` boolean on `close()`?

snazy:
> Maybe it's enough to reset the connected boolean on close()?
No, because it's not closed then.

I've added test-cases to the branch - a commit before the actual change, so the tests can be ran against the current version (force-pushed to move the addition of the tests before the fix).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
```diff
@@ -34,22 +42,82 @@ public ProxyConnection(ReactiveConnectionPool sqlClientPool, String tenantId) {
...
 			}
-			return operation.apply( connection );
 		}
+
+		// "Queue up" the 'operation'.
+		// If the connection's already available, the 'operation' will likely be executed
+		// via this call site.
+		// If the connection's not yet available, the 'operation' will be executed once
+		// the 'completionStage' to acquire the connection has finished.
+		return completionStage.thenCompose(operation);
```
Sanne, line 120: +1 I agree we need to compose subsequent calls into the "being acquired" handle here.

But I'm not sure that synchronization and volatiles are needed here - do you have more details about that or are you just adding them to be one the safe side?

gavinking: Are you really sure. Read the comment in the deleted code. Clearly I thought about this and decided the only way it could be null is in the case of a concurrent call, which it still seems to me is clearly wrong.

I'm still not convinced there is a bug here.

gavinking:
So the code we're arming about is:

```java
private <T> CompletionStage<T> withConnection(Function<ReactiveConnection, CompletionStage<T>> operation) {
		if ( !connected ) {
			connected = true; // we're not allowed to fetch two connections!
			CompletionStage<ReactiveConnection> connection =
					tenantId == null ? sqlClientPool.getConnection() : sqlClientPool.getConnection( tenantId );
			return connection.thenApply( newConnection -> this.connection = newConnection )
					.thenCompose( operation );
		}
		else {
			if ( connection == null ) {
				// we're already in the process of fetching a connection,
				// so this must be an illegal concurrent call
				throw new IllegalStateException( "session is currently connecting to database" );
			}
			return operation.apply( connection );
		}
	}
```

I don't see anything wrong with this. This `ProxyConnection` belongs to a single reactive stream. If you're only accessing the its `Session` from within chain links of that stream, I don't see how it's possible that you could get a second request for a connection while the first one is still incomplete.

snazy:
> I don't see anything wrong with this. This ProxyConnection belongs to a single reactive stream. If you're only accessing the its Session from within chain links of that stream, I don't see how it's possible that you could get a second request for a connection while the first one is still incomplete.

That's the point: only accessing it from within a single chain. I.e.
```
session.withTransaction(tx -> session.find(...)... );
```
works, but
```
Uni<X> x = session.find(...);
session.withTransaction(tx -> x);
```
does not. But there's nothing that prevents users from doing this. So people can easily run into this issue without knowing it albeit all tests passing. It's a nasty thing, because you can't really figure out why that's happening and it only happens, when the pool needs more time to acquire a connection. Users probably don't know and should not need to know how stuff works internally. Finding those call sites in a large code base would be a nightmare.

The docs for `Uni` tell users that nothing will be computed before it's being subscribed - but hibernate-reactive does computation (acquiring the connection) before it's being subscribed.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
snazy: > not sure that synchronization and volatiles are needed here

Yup, those are needed to prevent races. You cannot guarantee that all invocations there will happen on the same thread. So visibility-issues come into play - `volatile` mitigates some of those. x86 platforms are rather "friendly" wrt load/store instructions reordering to people, but other platforms like arm reorder instructions quite aggressively - so being very careful is not bad per se. If there's a chance that one field is accessed concurrently, `volatile` is your friend. I've seen such reordering issues in the past and those are really hard to catch - up to the point that other threads never saw a specific change.

gavinking:
> That's the point: only accessing it from within a single chain. I.e.
> 
> ```
> session.withTransaction(tx -> session.find(...)... );
> ```
> 
> works, but
> 
> ```
> Uni<X> x = session.find(...);
> session.withTransaction(tx -> x);
> ```

Right, so that's the thing: you're not allowed to do that, not in regular JPA, not in Hibernate Reactive. The session is always by nature very much a single-threaded object.

From [the documentation](https://github.com/hibernate/hibernate-reactive/blob/master/documentation/src/main/asciidoc/reference/introduction.adoc#obtaining-a-reactive-session) we have:

> Persistence operations are exposed via a reactive `Session` object. It's very
> important to understand that most operations of this interface are non-blocking,
> and execution of SQL against the database is never performed synchronously.
> Persistence operations that belong to a single unit of work must be chained by
> composition within a single reactive stream.
>
> Also remember that a Hibernate session is a lightweight object that should be
> created, used, and then discarded within a single logical unit of work.
>
> IMPORTANT: That is to say, you should reuse the same session across multiple
> persistence operations within a single reactive stream representing a certain
> unit of work, but don't share a session between different reactive streams!

This is a really important restriction!

> But there's nothing that prevents users from doing this.

Nor is there in JPA.

> So people can easily run into this issue without knowing it albeit all tests passing. It's a nasty thing, because you can't really figure out why that's happening and it only happens, when the pool needs more time to acquire a connection.

OK, so let's investigate if we can do a better job of detecting this condition. It's probably not going to be easy, however. We would need a way to determine which reactive stream a session is being called from, and right now we don't have a good way to do that.

> It's a nasty thing, because you can't really figure out why that's happening and it only happens, when the pool needs more time to acquire a connection. Users probably don't know and should not need to know how stuff works internally. Finding those call sites in a large code base would be a nightmare.

This is true, but your proposed patch actually makes this situation *worse*, since removes one way the user has of finding out that they're doing session management in an unsafe way.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
Sanne: No I'm not really sure - I just did a session in pair with @DavideD trying to create a different test (w/o using threads) and we couldn't trigger the issue, however I didn't get to emulate a limited pool yet.

But if the state is possible, composing them seems fair. But +1 to understand more how this could happen.

Sanne:
> > That's the point: only accessing it from within a single chain. I.e.
> > ```
> > session.withTransaction(tx -> session.find(...)... );
> > ```
> > 
> > 
> > works, but
> > ```
> > Uni<X> x = session.find(...);
> > session.withTransaction(tx -> x);
> > ```
> 
> Right, so that's the thing: you're not allowed to do that, not in regular JPA, not in Hibernate Reactive. The session is always by nature very much a single-threaded object.

I honestly don't follow here @gavinking . The `find` operation is _expressed_ before the withTransaction scope, but it's not _executed_ yet. It should be fine to declare the operation chain anywhere?

And even if I miss some point there, how does this tie into "multiple threads" ?

gavinking:
> I honestly don't follow here @gavinking . The `find` operation is _expressed_ before the withTransaction scope, but it's not _executed_ yet: it's _executed_ within the transaction. 

Well that's not what we decided to do. We did *consider* doing it that way, but it's not what we ultimately implemented.

Currently `Session.find()` does stuff: it alters the internal state of the `Session` synchronously, and it's only the operations against the database that are executed async. (We've discussed that point several times.)

> It should be fine to declare the operation chain anywhere?

Well, no, it's not. Currently you're expected to call operations of the `Session` from "within the chain". 

(We could change that, if we think we need to.)

> And even if I miss some point there, how does this tie into "multiple threads" ?

I'm not sure either. Apparently somehow, if I understand correctly, @snazy has multiple threads hitting the same session, which is very much a no-on.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -160,6 +170,69 @@ public void after(TestContext context) {
...
+			Pool delayingPool = new Pool() {
+				@Override
+				public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
+					pool.getConnection(ar -> {
+						// maybe there's a nicer way to do this asynchronously
+						new Thread(() -> {
+							try {
+								if (!synchronizer.await(60, TimeUnit.SECONDS))
+									handler.handle(Future.failedFuture("synchronizer await failed"));
+								handler.handle(ar);
```
Sanne, line 198: I honestly can't think now of a better way to introduce delay, but I think it's wrong to have the helper thread invoke a method on the original `handler` here as that would violate the assumptions.

Sanne:
> Currently Session.find() does stuff: it alters the internal state of the Session synchronously, and it's only the operations against the database that are executed async. 

Ok thanks! I forgot that and it's much clearer now.

So yes what's missing here is to understand why there's multiple threads at play. I'm trying to reproduce the issue still but need a different test as I'm not liking how the helper thread is then invoking the handler it's not owning:
 - https://github.com/hibernate/hibernate-reactive/pull/476/files#r539512286

snazy:
> somehow, if I understand correctly, @snazy has multiple threads hitting the same session

Nope. At least not my coder-as-user-of-the-libraries-self ;)
"Multiple threads" comes into play in two ways:
1. via the callback from the connection-pool into the `CompletionStage` that informs the `ProxyConnection` that the connection's now available.
1. when the chain's suspended (because it's waiting for things to complete) and the work gets scheduled (resumed) onto a different thread. "Things" could be anything really, not just database stuff - maybe waiting for an external REST service or waiting for an email to be sent or a file operation to complete - and that can happen within "our" transactional chain. Mean, it's the nature of reactive stuff: the implementation decides where (on which thread) work's being executed.

Sanne:
But IFF the whole chain is being rescheduled on a different thread, it's my understanding that Vertx will guarantee both:

- exclusivity, as in no two threads will be allowed to continue working on the same context
- visibility, as in correctly handling over the full state to the new thread with the appriopriate memory barriers

@vietj @FroMage correct?

Apologies I'm new to reactive programming and Vertx.

snazy:
> But IFF the whole chain is being rescheduled on a different thread, it's my understanding that Vertx will guarantee both:
> 
> * exclusivity, as in no two threads will be allowed to continue working on the same context
> * visibility, as in correctly handling over the full state to the new thread with the appriopriate memory barriers

Not sure about both. For a chain that only has `Uni`s in it, I'd assume that your first point is correct. If there's a `Multi` at play in that chain, I'd assume that all bets are off (but: disclaimer: this is rather out of thin air - I've got not evidence, just guessing).

Visibility - not sure here. Mean, it would be just a nice thing to have load/store-fences around the invocations, but you can compose these things using a ton of different frameworks/libraries ending in gazillion different ways.

gavinking:
> "Multiple threads" comes into play in two ways:
> 
> 1. via the callback from the connection-pool into the `CompletionStage` that informs the `ProxyConnection` that the connection's now available.

Well, I don't think so. It seems to me that this would be the same (logical) thread that then continues on and executes the rest of the chain.

> 2. when the chain's suspended (because it's waiting for things to complete) and the work gets scheduled (resumed) onto a different thread. 

Right, that's what I mean by "logical" thread. Even if things get suspended and resumed on a different thread, they're still executed sequentially, and there's no concurrency to deal with.

AFAICS, there should be no need for volatiles or synchronization.

gavinking:
> If there's a `Multi` at play in that chain, I'd assume that all bets are off

I don't *think* that's right. I believe there are indeed *explicit* ways to fork and join within a stream, but I don't *think* it ever just happens transparently. @cescoffier can clarify.

gavinking:
> It seems to me that this would be the same (logical) thread that then continues on and executes the rest of the chain.

@snazy just to be sure: the "callback from the connection-pool" is executed on the event loop itself. That's the whole point of having a reactive driver.

There's no additional thread pools dispatching database requests here.

cescoffier:
Neither `Uni` nor `Multi` does anything "automatic" about the threads. 
What's guaranteed is that a specific _stage_ does not receive concurrently multiple events (items, failure...). 
You can control the subscription and emission thread if you need to. For this use the `runSubscriptionOn` or `emitOn` operators. Both take an executor as parameter.

Sanne:
Debugged Quarkus: looks nice @snazy is right, a single `ProxyConnection` instance is being exposed to multiple threads in practice - however this wasn't our intention.

The problem is that the Session (and the user code) is being executed on the `executor-thread-1`, but then when we wait on a response from the DB, the continuation is fired on the Vertx pool which is handling the connection pool: `vert.x-eventloop-thread-0`.

The usage pattern that Robert suggested implies that
 - "find" operation is created, but not executed yet (on the user thread)
 - we start a transaction -> wait for the TX start to be acknowledged by the DB (on the user thread, "wait" meaning we move on to next step)
 - then we actually trigger the "find", this triggers the "onConnection" operation on `ProxyConnection` (on the user thread)
 - eventually the TX gets confirmed by the DB -> _on the Vertx Event loop_ 

At this point we have:

- concurrent state change on `ProxyConnection` by two threads
- one of them might (or not) see the `connected` flag, but likelyhood is high for the `Connection` reference not being initialized yet, or just not visible yet.

But this seems an integration problem; we need to figure out how to guarantee the in/out operations are executed consistently on the same thread. Perhaps this also entails making `ProxyConnection` safer, but I'd do it in such a way to _detect_ misuse rather than trying to make it work out, as otherwise this will just hide problems, and possibly move the problem to a different area: we can't make the whole persistence context threadsafe.

gavinking:
> The problem is that the Session (and the user code) is being executed on the `executor-thread-1`, but then when we wait on a response from the DB, the continuation is fired on the Vertx pool which is handling the connection pool: `vert.x-eventloop-thread-0`.

Yeah, I noticed this myself last night: but not in Quarkus, this seems to be the same even in our test suite. I had never noticed it before.

And it sort-of came as a surprise to me. Which I guess that just reflects my lack of understanding of how Vert.x works under the covers. Anyway, I think we need to ask for a much better explanation from Julien and Clement because apparently there were subtleties here that I was missing.

> But this seems an integration problem; we need to figure out how to guarantee the in/out operations are executed consistently on the same thread.

I agree.

Or at least on the same *logical* thread, even if it's a single "thread" where control passes between multiple Java `Thread`s.

> but I'd do it in such a way to detect misuse rather than trying to make it work out, as otherwise this will just hide problems

Yes, exactly, I agree, that's what I was trying to say above in [this comment](https://github.com/hibernate/hibernate-reactive/pull/476#issuecomment-741911679).

DavideD:
> Yeah, I noticed this myself last night: but not in Quarkus, this seems to be the same even in our test suite. I had never noticed it before.

Just for curiosity, how did you notice this in our test suite?

gavinking:
> Just for curiosity, how did you notice this in our test suite?

I stuck in some breakpoints and put a watch on the thread names.

snazy:
One possibility to detect when it's used in an unintended way could be to use the synchronized+CompletionStage from the patch but instead of "piggy-backing" onto the saved CompletionStage, throw some exception.

snazy:
I played around with the stuff a bit more and found another interesting race that happens when the underlying pool is exhausted. I simulated it locally with a connection pool with max-size = 1 and then "just" overloading it (keep ctrl+R pressed in a browser window for some seconds to repeat requests). The application uses `io.quarkus.security.identity.SecurityIdentityAugmentor`, which returns an `Uni`. But unlike "normal requests", there's no active context, when `SecurityIdentityAugmentor.augment` is being called. Anyway - my implementation creates a request `ManagedContext` and runs some DB operations. The whole step to augment the security-identity is "guarded" with an `Uni.await().atMost(...)`. In pseudo-code it looks like this:
```
    public Uni<SecurityIdentity> augment(SecurityIdentity identity, AuthenticationRequestContext context) {
        return context.runBlocking(() -> {
                try {
                    requestContext.activate();
                    return mutinySession.withTransaction(tx -> doSomeStuff()).await().atMost(blockingAugmentTimeout);
                } finally {
                    requestContext.terminate();
                }
        });
    }
```

This looks quite innocent, but in case the `doSomeStuff` leads to a timeout-exception and the surrounding `ManagedContext` gets closed. But the "request" to acquire a connection via `ProxyConnection` is still lingering around as a "waiter" in `io.vertx.sqlclient.impl.ConnectionPool`. It does get the callback and uses the DB connection, but the owning hibernate session has already been closed and the transaction aborted.

I think, the lambda/function passed to the `thenApply` in `withConnection` needs to check whether the `ProxyConnection` has been closed and then return some dummy `ClosedReactiveException` which methods always returns a `CompletionStage` that those some descriptive `IllegalStateException`.

WDYT?

snazy:
Oh, and as a side note: the connection-pool implementation in vertx (io.vertx.sqlclient.impl.ConnectionPool in 3.9.x / io.vertx.sqlclient.impl.pool.ConnectionPool in latest master) isn't thread-safe (https://github.com/eclipse-vertx/vertx-sql-client/issues/849)

Sanne:
> Oh, and as a side note: the connection-pool implementation in vertx (io.vertx.sqlclient.impl.ConnectionPool in 3.9.x / io.vertx.sqlclient.impl.pool.ConnectionPool in latest master) isn't thread-safe ([eclipse-vertx/vertx-sql-client#849](https://github.com/eclipse-vertx/vertx-sql-client/issues/849))

well I think that's a consequence of the Vert.x design - the principles I listed above. But I'm glad you opened the issue, would be good to get a confirmation from them. I'm also scheduling a meeting with them to clarify ;)

FroMage:
First thing wrt https://github.com/hibernate/hibernate-reactive/pull/476#issuecomment-741894044 and `withConnection`, @gavinking you'd have the same bug even single-threaded if someone does this:

```java
withConnection(con -> someOperation);
withConnection(con -> someOperation);
```

Supposing `someOperation` takes a while to kick in, the user is doing two calls to `withConnection` before `someOperation` resolves, and didn't wait for the first call to complete before calling `withConnection` again, so you'll get the same bug.

FroMage:
Second, about thread affinity: it looks like you start the request on the Quarkus worker pool, and then complete it on the event loop. If this is indeed Quarkus's worker pool, Vert.x doesn't know about it, and so it can't track/respect thread affinity with its event loop. I guess this would work if the user started its request on the Vert.x worker thread pool, right @vietj?

How we can resolve this, OTOH, I'm not sure.

gavinking:
> ou'd have the same bug even single-threaded if someone does this:

@FroMage, nope, I don't think so, because a `ProxyConnection` is one-per-session and you're not allowed to use a session in that way. You have to serialize it's operations within a single reactive stream. That's what I tried to explain above.

gavinking:
I mean, if you have people making calls on a reactive session like that, you're going to have waaaay worse race conditions than `ProxyConnection`!

snazy:
Seems like the only way to "get around" that is to move all the actual acquire-connection-work to the subscription phase. It might work if you introduce an indirection via some "connection sentinel object" that gets the actual connection injected during subscription. Just an idea.

FroMage:
> I mean, if you have people making calls on a reactive session like that, you're going to have waaaay worse race conditions than ProxyConnection!

I'm saying that you can get the exact same issue with a single thread trivially. Even if people are not supposed to do this, given that there's no detection, it will happen, by accident. Even without the complexity of threads.

gavinking:
Let me try again to explain this: the only way you can get this race condition is if you have someone making a second invocation of the same `Session` *before* the `CompletionStage returned by the previous invocation has been marked completed.

That's something that's clearly disallowed. To even do it you would have to be either:

1. leaking a reference to the `Session` to another thread/reactive stream, or
2. be accidentally discarding the tip of the current stream (a *very, very* common user error).

gavinking:
> I'm saying that you can get the exact same issue with a single thread trivially. Even if people are not supposed to do this, given that there's no detection, it will happen, by accident. Even without the complexity of threads.

Yes, I know. That's condition 2 I just mentioned, and which is explicitly called out as a common error in the HR docs.

FroMage:
> be accidentally discarding the tip of the current stream (a very, very common user error).

precisely my point, I'm surprised this turns out to be a thread issue after all and not just inadvertance.

FroMage:
I wonder if you can't do something to detect that though, like count pending operations and yell if someone drops one?

gavinking:
> I wonder if you can't do something to detect that though, like count pending operations and yell if someone drops one?

Well I've been wondering the same thing since 2 days ago when I wrote:

> OK, so let's investigate if we can do a better job of detecting this condition. It's probably not going to be easy, however. We would need a way to determine which reactive stream a session is being called from, and right now we don't have a good way to do that.

Now, we could do it with something like your MP Contexts stuff, but we don't have that here. And I've not seen any sort of "reactive thread id" in Mutiny. (Perhaps it exists and I have not noticed.)

gavinking:
@snazy 

> Seems like the only way to "get around" that is to move all the actual acquire-connection-work to the subscription phase. 

Well at one stage `openSession()` returned a `Uni`, but everyone hated that so I had to introduce `ProxyConnection`. Though note that `ProxyConnection` is just a wrapper around the "real" connection, and it would be easy as pie to go back to the original design.

You can read the whole messy history of this thing here:

https://github.com/hibernate/hibernate-reactive/issues/316

gavinking:
> @snazy
> 
> > Seems like the only way to "get around" that is to move all the actual acquire-connection-work to the subscription phase.
> 
> Well at one stage `openSession()` returned a `Uni`, but everyone hated that so I had to introduce `ProxyConnection`. Though note that `ProxyConnection` is just a wrapper around the "real" connection, and it would be easy as pie to go back to the original design.

But again: connection management and `ProxyConnection` aren't really the problem here. Hibernate sessions themselves are stateful and are vulnerable to the exact same issue.

gavinking:
Oh, haha, here is where we actually talked about this exception:

<https://github.com/hibernate/hibernate-reactive/pull/317#issuecomment-672125940>

snazy:
I think, the confusion arises from the overall design choices that all these reactive frameworks made. While one framework wants you to initialize up-front (ReactiveX IIRC), others want you to separate declaration and execution (Mutiny/Vertx).

gavinking:
> I think, the confusion arises from the overall design choices that all these reactive frameworks made. While one framework wants you to initialize up-front (ReactiveX IIRC), others want you to separate declaration and execution (Mutiny/Vertx).

@snazy well, the door is definitely still open to making the operations of `Session` *completely* async, in order to support both those approaches. It simply had not come up as a requirement.

snazy:
Okay, I took a stab at that and it seems to work. I've put a bunch of comments in the code and added a few more tests and cleaned up the earlier added tests.

For all the stuff `Mutiny` the declaration/construction "phase" no longer acquires a connection from the pool. But once it's being subscribed, it triggers the async process to acquire a connection. For the other approaches (think: "just" `CompletionStage`) such a separation isn't really possible, because there's no separate "subscription" phase.

The essence of the whole change: The `ProxyConnection`'s `withConnection` method registers a _demand_ for a connection, but the `openConnection` triggers the actual "pool! gimme a connection!" async process. I thought about concurrency and race conditions here and came to the conclusion that the state management in `ProxyConnection` requires synchronized state-evaluation and state-changes. Just the fact that the connection-pool can call at any time and from any thread makes this necessary. On top of that, handling the situation when the `ProxyConnection`'s closed before the pool issues the callback requires synchronized state-evaluation and state-changes as well. "At least" to ensure that there's no connection-leak.

The whole PR is "mostly boring" (encapsulating the "enrichment" via `Uni.onSubscribe` etc). The interesting part is just in `ProxyConnection`. `MutinyUniConnectionActivator` is there for the "`Uni` enrichment". `ReactiveConnection` got a new method `openConnection` (it was easier to implement the changes with that one - and it's harmless if people use it, but also a "no-op"). `MutinyExceptionsTest.testDuplicateKeyException` never subscribed to the `Uni`s, so it got a slight change. `ProxyConnectionRaceTest` got a few more test cases. The only thing I'm not really happy with is the for-test-stuff in `ProxyConnection` - but it's also not overly ugly, but I needed a way to inject a "customized connection pool" into `ProxyConnection` for the tests.

Also verified with a pseudo-scientific web-app use case: a pool with max-size=1, long running database operations and a dummy who keeps ctrl-R pressed (yes, it runs into timeouts, overload, etc - but eventually recovers fine).

Oh, took the freedom to add a separate commit that adds one of my Gradle plugins to rerun all tests (with `-PtestRerun`) and do repeated test runs (via `PtestRepetitions=N`). Happy to move it into a separate PR.

voigtste:
any updates on this? Currently trying to use that in quarkus and heavily hitting the issuer under load. The applied fix for this would be great :-)

DavideD:
Sorry, somehow we missed this new changes (probably because we went on holiday soon after). I should be able to have a look at it in the next couple of days.
In the meanwhile, if you could rebase it to the latest changes. It would be great.

Another thing that would help me is if you could keep the tests in a separate commits. It makes it easier for me to run them on the current master and verify that they actually fail.

Thanks

DavideD:
I gave a look at this PR to see if we could reuse the tests but we did several important changes in the meanwhile and I'm not even sure if the issue can occur anymore.

@snazy I'm going to close this PR, but if you think the issue is still there, could you please create a test that works with the current `master` branch? Being able to apply the tests on master separately makes it much easier to review fixes.

Thanks

PR #477, opened 2020-12-14
Add --add-opens options required for Gradle

yrodiere:
In order to work around illegal accesses in Gradle.

Workaround for https://github.com/gradle/gradle/issues/15538

This should fix the JDK16 build.

Signed-off-by: Yoann Rodière <yoann@hibernate.org>


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/477

Thanks!

PR #484, opened 2020-12-14
[#482] Update to ORM 5.4.26.Final

DavideD:


DavideD:
It required this other PR to work with JDK 16.ea+28: https://github.com/hibernate/hibernate-reactive/pull/484

Issue #478, opened 2020-12-14
Run tests in the Vert.x Loop Thread


DavideD:
Our assumptions was that a test would run in a single Vert.x Event  loop thread. But Vert.x Unit run by default in a JUnit thread.
We need to apply the rule described [in this section of the Vert.x docs](https://vertx.io/docs/vertx-unit/java/#_running_a_test_on_a_vert_x_context
) to make sure that we run the tests as expected.

DavideD:
@gavinking is this correct? ^

gavinking:
I suppose so. I mean, perhaps this is not a critical thing to do, but it seems to me that this is what we always really intended to happen, isn't it?

DavideD:
Yes, I think so.

It's probably not critical because our tests are simple at the moment. It could become critical down the road if we add more complex tests.

If it's only about adding a rule to the base test class I don't see why not.

gavinking:
Yup. Agreed.

DavideD:
Interesting, now that everything runs  in a single Vert.x event loop thread, Vert.x complains because our schemaDrop operations are blocking: https://github.com/hibernate/hibernate-reactive/blob/4e244618617999c5f213c8a47471c6ce4077b6ff/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveGenerationTarget.java#L68

DavideD:
THis is the exception I'm having:
```
io.vertx.core.VertxException: Thread blocked
	at java.base@11.0.7/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@11.0.7/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
	at java.base@11.0.7/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1796)
	at java.base@11.0.7/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3128)
	at java.base@11.0.7/java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1823)
	at java.base@11.0.7/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2043)
	at app//org.hibernate.reactive.provider.service.ReactiveGenerationTarget.release(ReactiveGenerationTarget.java:68)
	at app//org.hibernate.tool.schema.internal.SchemaDropperImpl.doDrop(SchemaDropperImpl.java:131)
[...]

PR #486, opened 2020-12-16
[#478] Run tests in the Vert.x event loop thread

DavideD:
All the tests extending BaseReactiveTest will now run in the Vert.x event loop Thread by default.

Because the schema generation is blocking I had to delegate the session factory to a Vert.x work pool.


Sanne (approved):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/VertxInstance.java
```diff
@@ -24,7 +24,7 @@
  * <pre>{@code
  * new ReactiveServiceRegistryBuilder()
  *     .applySettings( properties )
- *     .addService( VertxInstance.class, () -> myVertx )
+ *     .addService( VertxInstance.class, (VertxInstance) () -> myVertx )
```
gavinking, line 27: @DavideD what's going on here? Why is this typecast now necessary?

DavideD: When I tried this code I had a compilation error and that was the only way of making it work.
Did I miss something?

gavinking: Oh so you're saying it was always necessary and my Javadoc code example was wrong?

DavideD: Yep

gavinking: OK. NP

PR #488, opened 2020-12-16
Thread and EventLoop context Assertions

Sanne:
I'm introducing some assertions, which are now passing after Davide's #478 (They didn't pass before).

I know we don't often use `assert`, but I feel it's a good compromise in this context:

- we're only focusing on Vertx currently
- we need to nail down our design mistakes
- it's not going to bother anyone at runtime, other than tests and people who enable assertions

WDYT ?


Sanne:
Two commits:
 

1. Verify all operations happen on the Vert.x eventloop
2. Verify that once a Session is opened, all subsequent operations happen on the same Thread

DavideD:
LGTM

Issue #480, opened 2020-12-14
atomicity of session operations


gavinking:
Currently persistence operations like `session.persist()` and `session.remove()` aren't really atomic from the point of view of the user:

1. (most) work updating the internal state of the session is done synchronously, then
2 database operations are done asynchronously.

This sorta makes sense when you consider how we intend the session to be used in the context of Mutiny or a chain of `CompletionStage`s, but it has a couple of problems:

- The separation described above started out _fairly_&mdash;but not _perfectly_&mdash;clean, but we've already had to break it when we fixed a couple of bugs.
- It implicitly assumes that users call operations from inside the callbacks instead of calling the operations and then changing together `Uni`s or `CompletionStage`s. That's certainly how we _intend_ people to use the API, but perhaps it's not always the _only_ reasonable way to use it.
- It means that some of the work of the session is executed outside the current vert.x context. 

So, I'm inclined to think we should change this and do the work on updating the session state asynchronously.

PR #490, opened 2020-12-16
New internal "Pool of pools" and disposal

Sanne:
I have not tested it yet properly, and still need to figure out how Quarkus will be able to integrate this.

Fixes #481

PR #493, opened 2020-12-17
run all operations in a Vert.x context

gavinking:
See #480.

Previously, work could be done on the calling thread, outside of a Vert.x context.

This code makes the Mutiny API do work lazily, dispatching it to the Vert.x context.

Open questions:

- Is it too expensive to do this for *every* operation? Should we only do it in `withSession()` / `withTransaction()`, and just make the other operations lazy, but not dispatch them to the Vert.x context?
- How do I do this with `CompletionStage`s?
- Is this code actually correct, @vietj @cescoffier?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -7,6 +7,8 @@ package org.hibernate.reactive.mutiny.impl;
 
 import io.smallrye.mutiny.Uni;
 
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
 import org.hibernate.Cache;
 import org.hibernate.HibernateException;
 import org.hibernate.internal.SessionCreationOptions;
@@ -17,12 +19,15 @@ import org.hibernate.reactive.pool.ReactiveConnectionPool;
 import org.hibernate.reactive.session.impl.ReactiveCriteriaBuilderImpl;
 import org.hibernate.reactive.session.impl.ReactiveSessionImpl;
 import org.hibernate.reactive.session.impl.ReactiveStatelessSessionImpl;
+import org.hibernate.reactive.vertx.VertxInstance;
 
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.metamodel.Metamodel;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static org.hibernate.reactive.common.InternalStateAssertions.assertUseOnEventLoop;
 
@@ -33,38 +38,56 @@ import static org.hibernate.reactive.common.InternalStateAssertions.assertUseOnE
  */
 public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
-	private SessionFactoryImpl delegate;
+	private final SessionFactoryImpl delegate;
+	private final VertxInstance vertxInstance;
+	private final Executor vertxExecutor;
 
 	public MutinySessionFactoryImpl(SessionFactoryImpl delegate) {
 		this.delegate = delegate;
+		vertxInstance = delegate.getServiceRegistry().getService(VertxInstance.class);
+		vertxExecutor = runnable -> {
+			Context context = vertxInstance.getVertx().getOrCreateContext();
+			if ( Vertx.currentContext() == context ) {
+				runnable.run();
+			}
+			else {
+				context.runOnContext( x -> runnable.run() );
+			}
+		};
+	}
+
+	<T> Uni<T> uni(Supplier<CompletionStage<T>> stageSupplier) {
+		return Uni.createFrom().completionStage(stageSupplier).runSubscriptionOn(vertxExecutor);
 	}
 
 	@Override
 	public Mutiny.Session openSession() {
 		SessionCreationOptions options = options();
 		return new MutinySessionImpl(
-				new ReactiveSessionImpl( delegate, options, proxyConnection( options.getTenantIdentifier() ) )
+				new ReactiveSessionImpl( delegate, options, proxyConnection( options.getTenantIdentifier() ) ),
+				this
 		);
 	}
 
 	@Override
 	public Mutiny.Session openSession(String tenantId) {
 		return new MutinySessionImpl(
-				new ReactiveSessionImpl( delegate, options(), proxyConnection( tenantId ) )
+				new ReactiveSessionImpl( delegate, options(), proxyConnection( tenantId ) ),
+				this
 		);
 	}
 
 	Uni<Mutiny.Session> newSession() throws HibernateException {
 		SessionCreationOptions options = options();
-		return Uni.createFrom().completionStage( connection( options.getTenantIdentifier() ) )
+		return uni( () -> connection( options.getTenantIdentifier() ) )
 				.map( reactiveConnection -> new ReactiveSessionImpl( delegate, options, reactiveConnection ) )
-				.map( MutinySessionImpl::new );
+				.map( s -> new MutinySessionImpl(s, this) );
 	}
 
 	Uni<Mutiny.Session> newSession(String tenantId) throws HibernateException {
-		return Uni.createFrom().completionStage( connection( tenantId ) )
+		return uni( () -> connection( tenantId ) )
 				.map( reactiveConnection -> new ReactiveSessionImpl( delegate, options(), reactiveConnection ) )
-				.map( MutinySessionImpl::new );
+				.map( s -> new MutinySessionImpl(s, this) );
 	}
 
 	@Override
@@ -77,7 +100,7 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
 	Uni<Mutiny.StatelessSession> newStatelessSession() throws HibernateException {
 		SessionCreationOptions options = options();
-		return Uni.createFrom().completionStage( connection( options.getTenantIdentifier() ) )
+		return uni( () -> connection( options.getTenantIdentifier() ) )
 				.map( reactiveConnection -> new ReactiveStatelessSessionImpl( delegate, options, reactiveConnection ) )
 				.map( MutinyStatelessSessionImpl::new );
 	}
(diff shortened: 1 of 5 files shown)


gavinking:
Is CI broken?

DavideD:
It seems so

DavideD:
I've restarted the job. Sometimes it has issues with the download of the dependencies.

gavinking:
Ah, good, much better now.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -34,37 +36,47 @@
...
 
 	public MutinySessionFactoryImpl(SessionFactoryImpl delegate) {
 		this.delegate = delegate;
+		vertxInstance = delegate.getServiceRegistry().getService(VertxInstance.class);
+	}
+
+	<T> Uni<T> uni(Supplier<CompletionStage<T>> stageSupplier) {
+		return Uni.createFrom().completionStage(stageSupplier)
+				.runSubscriptionOn( runnable -> vertxInstance.getVertx()
+						.getOrCreateContext().runOnContext( x -> runnable.run() ) );
```
gavinking, line 49: So, @vietj @cescoffier @Sanne just how expensive is this?

cescoffier: What we could do is something like:

```java
runnable -> {
   Context c = vertxInstance.getVertx().getOrCreateContext();
   if (Vertx.currentContext() == c) { 
      runnable.run();
   } else { 
      c.runOnContext(() -> runnable.run());
   }  

}
```

Sanne: I'd need to run some benchmarks but perhaps there's a better way out:

could we first check if we're on a context, then avoid this all if we already are?

For those users who are not running on the vert.x context, it would be more acceptable to have a performance penalty - they probably don't care as much.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -1316,12 +1312,12 @@ public boolean isOrderReturnEnabled() {
...
-		}
-		if ( Mutiny.Session.class.isAssignableFrom( clazz ) ) {
-			return clazz.cast( new MutinySessionImpl( this ) );
-		}
+//		if ( Stage.Session.class.isAssignableFrom( clazz ) ) {
+//			return clazz.cast( new StageSessionImpl( this ) );
+//		}
+//		if ( Mutiny.Session.class.isAssignableFrom( clazz ) ) {
+//			return clazz.cast( new MutinySessionImpl( this ) );
+//		}
```
Sanne, line 1320: was this a leftover of some experiment?

If not, would be nice to add a comment about the reason.

gavinking: I don't think we need it. I think we can just delete that stuff.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
gavinking: Well I would probably still want to make it lazy, in order to get session operation atomicity, but yes, we can definitely check if there's a context first.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -34,37 +38,54 @@
...
+		return Uni.createFrom().completionStage(stageSupplier)
+				.runSubscriptionOn( runnable -> {
+					Context context = vertxInstance.getVertx().getOrCreateContext();
+					if ( Vertx.currentContext() == context ) {
+						runnable.run();
+					}
+					else {
+						context.runOnContext( x -> runnable.run() );
+					}
+				} );
```
gavinking, line 58: Is this what you mean @cescoffier?

cescoffier: exactly. If we are already in the right context, just run the task immediately.

gavinking:
I've just pushed a proposed implementation for the `Stage` API.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -29,23 +35,41 @@
...
+				runnable.run();
+			}
+			else {
+				context.runOnContext( x -> runnable.run() );
+			}
+		};
+	}
+
+	<T> CompletionStage<T> stage(Function<Void, CompletionStage<T>> stageSupplier) {
+		return voidFuture().thenComposeAsync(stageSupplier, vertxExecutor);
```
gavinking, line 56: Everyone, does this look right?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
Sanne: sure. I'd delete it then - the comments made me wonder if there was some unfinished business

gavinking: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
DavideD: I'm not sure if that's the way one is supposed to use the Async.
My expectation was that the executor should start a Thread. But it seems to work and also makes the examples pass when this branch is rebased on top of @Sanne PR: https://github.com/hibernate/hibernate-reactive/pull/490

LGTM

Sanne:
A little update from other related tasks:
 - my new pool PR (#490 ) was failing to run examples as the assertions would start to fail, but it gets fixed by this PR - which is great. I'll update it after this one gets merged.
 - Related Quarkus patches are ready now; see https://github.com/quarkusio/quarkus/pull/13977 and the two issues it fixes.

This took a while as something dodgy is going on with the Reactive MySQL drive; it works fine when forcing it to run on the vertx context, but in the case of the simple operation which was failing this didn't seem necessary; I suspect another threading issue in the driver (only MySQL), but not to worry about now as it's all working fine when running exclusively on the vertx context.

Sanne:
(and these previous PRs are also already integrated in Quarkus, and required):
 - https://github.com/quarkusio/quarkus/pull/13948
 - https://github.com/quarkusio/quarkus/pull/13879

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
Sanne: Yea that line is definitely wrong: it's using the wrong formatting style :-)

Jokes aside, it seems reasonabe. I wouldn't expect any Thread to be started, as we should push such work onto the vert.x eventloop.

Sanne:
@gavinking this patch looks great to me, could we merge it? Asking as it's marked "draft", otherwise I'd have merged it already..

gavinking:
I sorta wanted to run a micro benchmark to assess the cost of doing this on every operation.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
DavideD: I meant in general, when one passes an executor to the thenComposeAsync, not in this particular case.
But it was just a wrong assumption from my side.

DavideD:
> I sorta wanted to run a micro benchmark to assess the cost of doing this on every operation.

I think it's worthy to merge this as it is (if it's only performance you are worried about) and check the performance as a separate issue

gavinking:
Fine then, I'm outvoted :-)

Go ahead!

Sanne:
cool, and I agree with both of you :)  I'll run some more performance tests - (finally will have the time now!) - but it's easier to do so when the stack is aligned with all these fixes.

Also, no matter what the exact performance results will be, I prefer starting from working code like this and then improve within the boundaries of not breaking this threading model.

Sanne:
So many integration tests in Quarkus started to work after we merged this one, but the `Panache Reactive` module is still in trouble - I believe we might need to rethink some of it. (Assuming we keep the assertions I recently introduced).

Take this `persist` method for example:
 - https://github.com/Sanne/quarkus/blob/2b7f0c397d064f6dd1ebd466179aa61d6a33a121/extensions/panache/hibernate-reactive-panache-common/runtime/src/main/java/io/quarkus/hibernate/reactive/panache/common/runtime/AbstractJpaOperations.java#L36

The `getSession` method will of course lead to opening a Session instance within the current working thread, which is not guaranteed to be an eventloop thread.

I suppose one approach is to apply the same pattern as this PR on all Panache Reactive as well? I'm concerned however that Panache's very "stateless looking" nature doesn't give a good place to record _which_ vert.x context we need to return to, to continue any subsequent work within the same high level operation. Perhaps I'm confused, need to keep reading..

gavinking:
Well, I wonder if Panache reactive should go for a more `withSession()` sort of pattern also. Would look a bit different I guess, since it wouldn't be a `Session` that gets passed in, but I guess it's still in principle a pattern that could make sense.

Sanne:
I wonder the same. But how would it look like?  With the Panache API being very entity-centric, trying to turn it all around to have a Session as entry point to express operations on data seems like a big change - this would lead to Panache being less useful than the traditional Hibernate API.

We might want to first try reworking the internal Panache methods to do something similar as in this PR.

Coincidentally, had a bug report this morning which is affected by this:
 - https://github.com/hantsy/quarkus-sample/blob/95954358a145b2f2badfcca032051ba230a0219b/resteasy-reactive/src/main/java/org/example/DataInitializer.java#L17-L36

PR #495, opened 2020-12-17
[#494] Wrong credentials cause thread to wait forever

### LINKED ISSUE #494: Thread waits forever when the credentials are wrong

DavideD:
Fixes #494


DavideD:
Sorry, it seems there is something wrong

PR #496, opened 2020-12-17
[#494] Wrong credentials cause Thread to wait forever

### LINKED ISSUE #494: Thread waits forever when the credentials are wrong

DavideD:
Fixes #494

Note that the example will print the error but will still get stuck.
I think that's because we use `join()` or `await().indefinetly()`


Sanne:
Thanks, yes it solved the issue and I can see the warning logged :+1:

Sanne (approved):

PR #500, opened 2020-12-22
[#481] Have the DefaultSqlClientPool deliver a different Pool instanc…

### LINKED ISSUE #481: don't share `Pool`s between vert.x contexts in vert.x 3
According to @vietj, we're supposed to have a single connection `Pool` per vert.x context.

(At least that's my interpretation of what he told us.)

Note that this model is changing in vert.x 4, so this is a bug that only affects us on vert.x 3.

Sanne:
…e for each Vert.x context

Fixes #481


Sanne:
updated, improved some minor typos and comments

PR #502, opened 2021-01-04
[#497] Examples get stuck when credentials are wrong

### LINKED ISSUE #497: Example get stuck if there is a problem with the connection
It can be tested by using wrong credentials for the connection.

The code will get stuck here (in [MutinyMain](https://github.com/hibernate/hibernate-reactive/blob/master/example/src/main/java/org/hibernate/example/reactive/MutinyMain.java#L51)): 
```
		// obtain a reactive session
		factory.withTransaction(
				// persist the Authors with their Books in a transaction
				(session, tx) -> session.persistAll(author1, author2)
		)
				// wait for it to finish
				.await().indefinitely();
```

As far as I know the exception  is propagated but the await continues indefinitely.

I guess we need to add proper exception management in both examples (MutinyMain and Main).

DavideD:
Fixes #497

It turns out that the problem is that we don't close the factory if an error occurs.
This pr also contains a commit to add the thread name in the log message for tests. I think it can be useful to have it but we can skip the commit if you think is unnecessary.

Cheers

### DIFF
--- a/hibernate-reactive-core/src/test/resources/log4j2.properties
+++ b/hibernate-reactive-core/src/test/resources/log4j2.properties
@@ -9,5 +9,5 @@ logger.hibernate.level = info
 appender.console.name = console
 appender.console.type = Console
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %highlight{%p} %style{%c{1}}{Blue} %m%n
+appender.console.layout.pattern = %highlight{%p} %style{%c{1}}{Blue} [%t] %m%n
 


example/src/main/java/org/hibernate/example/reactive/MutinyMain.java
```diff
@@ -42,107 +43,126 @@ public static void main(String[] args) {
...
+					// delete all the Books in a transaction
+					(session, tx) -> session.createQuery( "delete Book" ).executeUpdate()
+							//delete all the Authors
+							.call( () -> session.createQuery( "delete Author" ).executeUpdate() )
+			)
+					.await().indefinitely();
+
+		}
+		finally {
+			// remember to shut down the connection pool
```
Sanne, line 163: s/connection pool/ Hibernate /

example/src/main/java/org/hibernate/example/reactive/Main.java
```diff
@@ -43,107 +43,125 @@ public static void main(String[] args) {
...
+			factory.withTransaction(
+					// delete all the Books in a transaction
+					(session, tx) -> session.createQuery( "delete Book" ).executeUpdate()
+							// delete all the Authors
+							.thenCompose( $ -> session.createQuery( "delete Author" ).executeUpdate() )
+			)
+					.toCompletableFuture().join();
+		}
+		finally {
+			// remember to shut down the connection pool
```
Sanne, line 162: I'd change this comment to clarify it's not referring to the pool

Sanne:
Thanks, looks great. There's two comments I'd clarify, but I see they are existing comments - so feel free to that later ?

Sanne (approved):

DavideD:
I've updated the comments and remove the test.
I'm going to create a separate issue for adding back the test. It  fails with the latest ORM snapshot because it throws an `UnsupportedOperationException` instead of something saying that the credentials are wrong.
It seems a bug to me: https://github.com/hibernate/hibernate-reactive/runs/1645163396?check_suite_focus=true#step:6:667

Sanne:
ok, fair enough.

DavideD:
Merged!
Fixes #497

PR #509, opened 2021-01-11
[#479] Test exception when sharing session between contexts

### LINKED ISSUE #479: error when user shares session between multiple vert.x contexts
It's currently considered an error to share a `Session` between multiple reactive streams, so we should detect that condition and throw an exception.

It seems that the best we to detect the condition is to look at the vert.x context.

DavideD:
Resolves #479 
  
It's currently considered an error to share a Session between
multiple reactive streams, so we should detect that condition
and throw an exception.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultipleContextTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultipleContextTest.java
@@ -1,0 +1,146 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.concurrent.CompletionException;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.stage.Stage;
+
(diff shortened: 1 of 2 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultipleContextTest.java
```diff
@@ -0,0 +1,146 @@
...
+import io.vertx.ext.unit.TestContext;
+import org.assertj.core.api.Assertions;
+
+/**
+ * It's currently considered an error to share a Session between multiple reactive streams,
+ * so we should detect that condition and throw an exception.
+ */
+public class MultipleContextTest extends BaseReactiveTest {
+
+	private static final String ERROR_MESSAGE = "Detected use of the Reactive Session from a different Thread than the one which was used to open the Reactive Session - this suggests an invalid integration;";
```
gavinking, line 30: What's the ; for?

gavinking: Also, "Reactive" should be lowercase, since it's not a type or a proper noun.

DavideD: The error message is longer than the string I use here and contains a semicolon. I've update the assert to check a shorter version of the string.

I've also replaced 'Reactive' with 'reactive'

gavinking: Thanks!

DavideD:
Resolves #479 

Thanks

PR #510, opened 2021-01-11
Update WrongCredentialsTest

DavideD:
Test wrong credential exception for all dbs.

This is a follow up to #497, it might be better to test the exception for each db in case there are some changes after updating the dependencies


DavideD:
Closing this for now.
For some reason, it doesn't work with MySQL on CI but I cannot reproduce the problem locally: https://github.com/hibernate/hibernate-reactive/pull/510/checks?check_run_id=1687501809#step:5:315

This is the error message:
 ```
java.lang.UnsupportedOperationException: Unsupported authentication method: sha256_password
```

When we actually expect a `MySQLException` with a message saying that the credentials are wrong.

I will create a separate issue

PR #516, opened 2021-01-14
Service lookup for ReactiveConnectionPool to happen once only

### LINKED ISSUE #514: Performance: pool() access from SessionFactoryImpl(s) to not invoke the ServiceRegistry
Both `pool()` implementations in the Stage & Mutiny versions of the ReactiveSessionFactoryImpl should have a faster access the the pool instance.

Accessing Services from the `ServiceRegistry` is quite flexible, but should only be done during initialization, or to handle occasional / non-hot needs.

Sanne:
Fixes #514

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
@@ -7,6 +7,8 @@ package org.hibernate.reactive.pool.impl;
 
 import io.vertx.sqlclient.PoolOptions;
 import io.vertx.sqlclient.SqlConnectOptions;
+
+import org.hibernate.HibernateError;
 import org.hibernate.reactive.provider.Settings;
 import org.hibernate.service.spi.Configurable;
 
@@ -102,7 +104,10 @@ public class DefaultSqlClientPoolConfiguration implements SqlClientPoolConfigura
                     }
                 }
                 else {
-                    params = uri.getQuery().split("&");
+                    final String query = uri.getQuery();
+                    if ( query != null ) {
+                        params = uri.getQuery().split( "&" );
+                    }
                 }
                 for (String param : params) {
                     if ( param.startsWith("user=") ) {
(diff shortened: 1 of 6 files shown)


Sanne:
seems I didn't think this through - it looked too easy.

We can't do it this way as it would trigger an eager initialization of the pool, which implies accessing it from the framework-initializing-thread, which is unlikely to be the event loop.

Sanne:
I'll admit it's getting me very confused that the same object represents both the pool and the connection concept

Sanne:
Finally got back to this. It was actually straight-forward, when I had time to look better.

Waiting for CI

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
```diff
@@ -127,6 +132,10 @@ public SqlConnectOptions connectOptions(URI uri) {
             }
         }
 
+        if ( username == null ) {
+            throw new HibernateError( "The username parameter on the JDBC-style connection URL is required. For PostgreSQL, for example, use 'jdbc:postgresql://localhost/hreact?user=none'." );
```
DavideD, line 136: Is this really required?

gavinking: No, it's not. It can be specified by property. So that'a a very misleading error message.

Sanne: I had vertx refuse the configuration when it's not set. See the unit tests I've added.

gavinking: @Sanne, you certainly need to specify a username *somehow*, but you equivalently can specify it using the property `javax.persistence.jdbc.user`.

I know it works because we do it that way in the example program.

Sanne: yea something is fishy in the area, I don't deny that.

But _in this specific line_ if that `username` is null, you get an NPE from Vertx.  

I can't work on this today but sure I'll have another look, wondering if the `javax.persistence.jdbc.user` takes another path.

gavinking: That line *can't* be null if you've set the property, because the local var is initialized above as:

    String username = user;

The logic is correct, and I know for sure it works.

gavinking: See #530.

Issue #44, opened 2020-01-20
@ElementCollection


gavinking:
I think "element" collections are a rubbish feature, and even worse here since we don't have lazy fetching, but people are going to bitch and moan if we don't provide this particular special way for them to shoot themselves in the foot.

Low priority.

cambierr:
Hi @gavinking 

Could you ellaborate on 

> I think "element" collections are a rubbish feature, (...)

I'm not that experienced yet with JPA / Hibernate but I do not understand why you consider `@ElementCollection` that bad ? is there something I missed or don't know about that feature ? 

Thank you :)

gavinking:
Well I just think it’s quite uncommon to have tables with two columns that don’t eventually wind up needing additional columns as the system evolves. (Even just a version or a time stamp.) At that point, you’ll need to refactor your Java code. So it’s simply a brittle and inflexible design.

(And the same goes for @ManyToMany, by the way.)

Furthermore, this two-column table doesn’t have a very well-defined primary key, which stinks from a data modeling perspective.

I’ve been telling people not to use this for >10 years, but of course people don’t listen.

cambierr:
Interesting indeed, I agree with you on the blocked evolution path when using that ` @ElementCollection` feature.

Regarding the `@ManyToMany`, do you mean that you'd instead recommend a `@OneToMany` on the joining table instead of a direct `@ManyToMany` on the target table ?  for the same "evolution" reason I guess ? 

Thank you for your reply, that's very interesting

gavinking:
Right, my advice is that you should usually have an entity for every table, and that most relationships should be bidirectional @ManyToOne/@OneToMany or @OneToOne.

blafond:
Thought I'd take a look at this one, and run with it if it's not too hard?  Got a lot to learn here as well as ORM...

DavideD:
The PR has been merged. Not sure why this is still open.

PR #517, opened 2021-01-14
Element collection string

blafond:
This is a cleaned up version of changes required to support ElementCollection's in reactive.

The initial commit set supports List<String> values.  The included EagerElementCollectionTest tests find collection, remove element, replace element, replace collection and remove Entity containing the collection to test cascade remove of the collection's generated table entity.

Additional work to complete ElementCollection support includes:
- Changes and tests for other collection types Map, Set, etc...
- Support for @Embedded entity collections

Also found issues trying to do edits for the OneToMany use-case... so I updated the test for follow-on discussion.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
@@ -90,7 +90,7 @@ public class ReactiveBasicCollectionPersister extends BasicCollectionPersister i
 								if ( hasIdentifier ) {
 									loc = writeIdentifier( st, collection.getIdentifier( entry, i ), loc, session );
 								}
-								getIdentifierType().nullSafeSet( st, id, 1, session );
+								loc = writeElement( st, collection.getElement( entry ), loc, session );
 
 								// FIXME: Set the other parameters as well
 							}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
@@ -22,6 +22,7 @@ import org.hibernate.mapping.Collection;
 import org.hibernate.persister.collection.BasicCollectionPersister;
 import org.hibernate.persister.spi.PersisterCreationContext;
 import org.hibernate.pretty.MessageHelper;
+import org.hibernate.reactive.adaptor.impl.PreparedStatementAdaptor;
 import org.hibernate.reactive.pool.ReactiveConnection;
 import org.hibernate.reactive.session.ReactiveConnectionSupplier;
 import org.hibernate.reactive.util.impl.CompletionStages;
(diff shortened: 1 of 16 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,179 @@
...
+				valuePairList.add( id );
+				valuePairList.add( entries.next() );
+			}
+
+			// In order to insert element collection values create a native query with placeholders followed by
+			// value sets
+			return reactiveConnection
+					.update(
+							getSQLInsertElementCollectionString(
+									valuePairList.size() / 2 ),
```
DavideD, line 89: Why do you need `valuePairList.size() / 2 ` ? I don't understand the reason.
Does ORM do the same?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		return CompletionStages.voidFuture();
+	}
+
+	// FIXME: this method currently only works for List<String> element collections and uses a VALUE_STR = "(?, ?)"
+	// will not work for List, Map, etc... of @Embedded entity (i.e. complex type)
+	private String getSQLInsertElementCollectionString(int numRows) {
+		String defaultInsertString = getSQLInsertRowString();
+		StringBuilder buf = new StringBuilder( defaultInsertString.length() + ( numRows + 2 ) * VALUE_STR.length() );
```
DavideD, line 103: Are you replacing VALUE_STR because the vert.x driver has a different syntax for parameters?

DavideD: If that's the case I don't think this is the right approach, we should replace the parameters when the query is created in `getSQLInsertRowString()`.

DavideD: Do you even need this part? I think we already do the conversion of the parameters in: [SqlConnection#prepartedQuery](https://github.com/hibernate/hibernate-reactive/blob/master/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java#L163)

gavinking: Right, parameter placeholder replacement is handled by the connection layer. We don't need to do it here.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
blafond: I followed the basic logic within AbstractCollectionPersister and converted to using the reactiveConnection.update(..) call.

https://github.com/hibernate/hibernate-orm/blob/15e59b3690debe7f08b41b2ab73eb42b273a42dc/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L1286
is an example for recreate()...   Also used remove() and insertRows() logic to re-implement.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
blafond: k.. thx for the direction

blafond: > If that's the case I don't think this is the right approach, we should replace the parameters when the query is created in `getSQLInsertRowString()`.

getSQLInsertRowString() just returns a generated/cached AbstractCollectionPersister string created in it's constructor.

blafond: Note these strings are what you see when you enable SQL logging (i.e. you see placeholders  .. (?, ?..) etc.)

DavideD: Well.. yes, it's printing the query with the placeholders for the parameters.
Not sure why in this specific case you are trying to do something different

gavinking:
FTR this work relates to issue #44.

DavideD:
@blafond I'm going to close this PR for now. No point in notifying everybody while we are still working on it

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionTest.java
```diff
@@ -0,0 +1,275 @@
...
+	protected Configuration constructConfiguration() {
+		Configuration configuration = super.constructConfiguration();
+		configuration.addAnnotatedClass( Person.class );
+		return configuration;
+	}
+
+	@Before
+	public void populateDb(TestContext context) {
+		String[] phoneArray = new String[]{ "999-999-9999", "111-111-1111", "123-456-7890" };
+		thePerson = new Person( 7242000, "Claude", phoneArray);
```
DavideD, line 44: Because the purpose of this test is to test the insert, I wouldn't put it in a `@Before` method. I would create a dedicated test.

PR #518, opened 2021-01-14
[#507] Rewrite loops methods to improve performance

### LINKED ISSUE #507: Performance: org.hibernate.reactive.util.impl.CompletionStages.loop(IntStream, Function)	1064
The method `CompletionStage org.hibernate.reactive.util.impl.CompletionStages.loop(IntStream, Function)` is very hot at runtime, and is allocating too many short lived objects.

Investigate alternative options.

The primary problematic point seems to be its use from `org.hibernate.reactive.loader.ReactiveResultSetProcessor.initializeEntity(Object, boolean, SharedSessionContractImplementor, PreLoadEvent, Iterable)`.

DavideD:
Resolves #507

I rewrote the methods we use for loops so that we don't need the dependency to asyncUtil.
I assumed it was responsible for the creation of the object causing the issue.

I'm not sure if there is more we can do considering that we will have to loop through the objects one way or the other.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveResultSetProcessor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveResultSetProcessor.java
@@ -67,7 +67,7 @@ public interface ReactiveResultSetProcessor {
 		);
 
 		final Object[] hydratedState = entityEntry.getLoadedState();
-		return CompletionStages.loop(
+		return CompletionStages.loopWithoutTrampoline(
 				IntStream.range( 0, hydratedState.length )
 						.filter( i-> hydratedState[ i ] instanceof CompletionStage ),
 				i -> ( (CompletionStage<Object>) hydratedState[ i ] )
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
@@ -84,44 +84,74 @@ public class CompletionStages {
 	}
 
 	public static <T extends Throwable, Ret> Ret returnNullorRethrow(Throwable x) throws T {
-		if (x != null ) {
+		if ( x != null ) {
 			throw (T) x;
 		}
 		return null;
 	}
 
 	public static <T extends Throwable, Ret> Ret returnOrRethrow(Throwable x, Ret result) throws T {
-		if (x != null ) {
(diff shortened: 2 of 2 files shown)


gavinking:
But is that actually a correct implementation of trampoline? Doesn't it just create a long chain?

gavinking:
The purpose of `loop()` isn't convenience: it's there to avoid blowing up the stack if we do something like iterate over a result set with 500 objects in it.

DavideD:
> But is that actually a correct implementation of trampoline? Doesn't it just create a long chain?

No, it's not. I will have another look at it

DavideD:
But do we even need to use recursion?
I've tried this solution that iterate over the objects: https://github.com/hibernate/hibernate-reactive/pull/518/commits/343feae20bf2de25940773c41b0847d712a0124a

gavinking:
Well anytime you're building a long chain of these things, I understand that you run the risk of the equivalent of a stack overflow, the same kind of problem you get anytime you replace a loop with recursion (and so the standard solution is to replace recursion with a trampoline).


But honestly it was @FroMage who ran into this and was advocating that we use trampoline, and I'm not quite clear on the details. Probably @cescoffier can also explain it better.

Sanne:
Since the Trampoline pattern is awful in terms of efficiency of code, let's limit its use to where we're sure it's necessary

gavinking:
> let's limit its use to where we're sure it's necessary

OK, well how about this: you've identified the call to `loop()` at `ResultSetProcessor: 70` as a hot spot, and after careful inspection I agree that this probably doesn't need to be trampolined, since each entity should only have $O(1)$ associations.

So I think it would be OK to try removing the trampoline _from that particular site_, and see if that improves your benchmark.

WDYT?

Sanne:
yes that's what I'm after. Except it's not clear to me if it's safe to remove the Trampoline here as I don't fully understand it :)

If you say it's safe, remove it from this specific point.  The other points are not worthy of our attention ATM.

DavideD:
I've updated the PR and now the loop without trampoline only run for `ResultSetProcessor`

FroMage:
Trampolines are only useful for loops or recursion. If you do a small number of chains it should be fine, but you'll still pay in terms of stack size.

Sanne (approved):
Thanks all, let's merge this one as it seems effective to removes some noise from the VM

gavinking:
Wait, what, I thought we had agreed not to!

gavinking:
> Wait, what, I thought we had agreed not to!

Oh I see, I missed that it had been updated.

Issue #508, opened 2021-01-09
Performance: find an alternative to PostgreSQL's need for the Parameters helper


Sanne:
The `Parameters` class is being used to convert SQL into PostgreSQL specific format, but this is being flagged as significant overhead during profiling. It gets highlighted both during CPU and during memory profiling.

A better solution should be investigated, or ideally have us generate the right syntax to begin with.

gavinking:
> A better solution should be investigated, or ideally have us generate the right syntax to begin with.

I mean, we *really* tried to do this about a year ago, but frankly it turned out to be a real mess, and ultimately I just got fed up with it.

gavinking:
What I'm saying is that it might not be worth fixing.

gavinking:
(Though one thing worth trying might be to cache the translated strings in a `WeakHashMap`.)

Sanne:
I'd prefer we customize the ORM's strategy for generating the SQL - let's generate it using the right syntax rather than patching it after? Especially as in that way it would end up into the existing caches.

gavinking:
> I'd prefer we customize the ORM's strategy for generating the SQL 

There *soooo many* places that this touches, and so many classes we would need to extend and override things that sometimes aren't public that this is just a mighty pain. Trust me, I tried.

DavideD:
> There soooo many places that this touches, and so many classes we would need to extend and override things that sometimes aren't public that this is just a mighty pain. Trust me, I tried.

I agree

But what about processing the sql as soon as it is generated?

For example overriding `AbstractEntityPersister#generateUpdateString` to something like:
```
	@Override
	public String generateUpdateString(boolean[] includeProperty, int j, boolean useRowId) {
		String sql = super.generateUpdateString( includeProperty, j, useRowId );
		return sql != null
				? Parameters.process( sql )
				: null;
	}
```

gavinking:
Well, that's probably *better*, but it's still a lot of places, once you take into account all the variations of loaders and batch loaders, and off the top of my head I don't think it would help with queries (though perhaps there is an alternative solution for queries).

DavideD:
Something is wrong. I will send a new one when I've fixed it

DavideD:
Sorry, I closed the issue instead of the PR

DavideD:
I have a working branch that seems to improve the performance. I will update the PR soon

PR #519, opened 2021-01-18
[#508] Process SQL string during generation

DavideD:
Follows the conversation we had here: #508

It doesn't seem too bad but I'm not sure I've covered all the use cases. It might be a way to see which test we still miss.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/bulk/impl/ReactiveBulkIdStrategy.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/bulk/impl/ReactiveBulkIdStrategy.java
@@ -25,6 +25,7 @@ import org.hibernate.hql.internal.ast.tree.FromElement;
 import org.hibernate.hql.internal.ast.tree.UpdateStatement;
 import org.hibernate.hql.spi.id.AbstractMultiTableBulkIdStrategyImpl;
 import org.hibernate.hql.spi.id.AbstractTableBasedBulkIdHandler;
+import org.hibernate.hql.spi.id.IdTableInfo;
 import org.hibernate.hql.spi.id.MultiTableBulkIdStrategy;
 import org.hibernate.hql.spi.id.global.GlobalTemporaryTableBulkIdStrategy;
 import org.hibernate.hql.spi.id.local.IdTableInfoImpl;
@@ -37,6 +38,7 @@ import org.hibernate.persister.entity.Queryable;
 import org.hibernate.reactive.bulk.StatementsWithParameters;
 import org.hibernate.reactive.pool.ReactiveConnection;
 import org.hibernate.reactive.pool.ReactiveConnectionPool;
+import org.hibernate.reactive.pool.impl.Parameters;
 import org.hibernate.reactive.util.impl.CompletionStages;
 import org.hibernate.sql.Delete;
 import org.hibernate.sql.Update;
@@ -68,6 +70,7 @@ public class ReactiveBulkIdStrategy
 	private final boolean db2;
 	private final Set<String> createdGlobalTemporaryTables = new HashSet<>();
 	private final List<String> dropTableStatements = new ArrayList<>();
+	private final Parameters parameters;
 
 	private StandardServiceRegistry serviceRegistry;
 
@@ -82,6 +85,7 @@ public class ReactiveBulkIdStrategy
 	ReactiveBulkIdStrategy(Dialect dialect) {
 		super( new ReactiveIdTableSupport( dialect ) );
 		db2 = dialect instanceof DB297Dialect;
+		parameters = Parameters.create( dialect );
 	}
 
 	@Override
@@ -195,7 +199,8 @@ public class ReactiveBulkIdStrategy
 							Collections.addAll( parameterList, assignment.getParameters() );
 						}
 					}
-					statements.add( update.toStatementString() );
+					String sql = parameters.process( update.toStatementString(), parameterList.size() );
+					statements.add( sql );
 //					if ( useSessionIdColumn() ) {
 //						parameterList.add( SESSION_ID );
 //					}
@@ -207,6 +212,26 @@ public class ReactiveBulkIdStrategy
 
 			this.statements = ArrayHelper.toStringArray( statements );
 			this.parameterSpecifications = parameterSpecifications.toArray( new ParameterSpecification[0][] );
+		}
+
+		@Override
+		protected String generateIdInsertSelect(
+				String tableAlias, IdTableInfo idTableInfo, ProcessedWhereClause whereClause) {
+			String sql = super.generateIdInsertSelect( tableAlias, idTableInfo, whereClause );
+			return parameters.process( sql );
+		}
+
+		@Override
+		protected String generateIdSubselect(
+				Queryable persister, AbstractCollectionPersister cPersister, IdTableInfo idTableInfo) {
+			String sql = super.generateIdSubselect( persister, cPersister, idTableInfo );
+			return parameters.process( sql );
+		}
+
+		@Override
+		protected String generateIdSubselect(Queryable persister, IdTableInfo idTableInfo) {
+			String sql = super.generateIdSubselect( persister, idTableInfo );
+			return parameters.process( sql );
 		}
 
 		@Override
@@ -330,6 +355,24 @@ public class ReactiveBulkIdStrategy
 
 			this.statements = ArrayHelper.toStringArray( statements );
 			this.parameterSpecifications = parameterSpecifications.toArray( new ParameterSpecification[0][] );
+		}
+
+		@Override
+		protected String generateIdInsertSelect(String tableAlias, IdTableInfo idTableInfo, ProcessedWhereClause whereClause) {
+			String sql = super.generateIdInsertSelect( tableAlias, idTableInfo, whereClause );
+			return parameters.process( sql );
+		}
+
+		@Override
+		protected String generateIdSubselect(Queryable persister, IdTableInfo idTableInfo) {
+			String sql = super.generateIdSubselect( persister, idTableInfo );
+			return parameters.process( sql );
+		}
+
+		@Override
+		protected String generateIdSubselect(Queryable persister, AbstractCollectionPersister cPersister, IdTableInfo idTableInfo) {
+			String sql = super.generateIdSubselect( persister, cPersister, idTableInfo );
+			return parameters.process( sql );
 		}
 
 		@Override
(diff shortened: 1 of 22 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/TableReactiveIdentifierGenerator.java
```diff
@@ -227,17 +227,17 @@ protected int determineIncrement(Properties params) {
...
-				+ " where " + valueColumnName + "=?  and " + segmentColumnName + "=?";
+		return "update " + renderedTableName + " set " + valueColumnName + "=$1"
+				+ " where " + valueColumnName + "=$2  and " + segmentColumnName + "=$3";
 	}
 
 	protected String buildInsertQuery() {
 		return "insert into " + renderedTableName + " (" + segmentColumnName + ", " + valueColumnName + ") "
-				+ " values (?, ?)";
+				+ " values ($1, $2)";
 	}
```
gavinking, line 241: These changes don't look right: `TableReactiveIdentifierGenerator` isn't specific to Postgres.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/bulk/impl/ReactiveBulkIdStrategy.java
```diff
@@ -209,6 +212,26 @@ public TableBasedUpdateHandlerImpl(SessionFactoryImplementor factory, HqlSqlWalk
 			this.parameterSpecifications = parameterSpecifications.toArray( new ParameterSpecification[0][] );
 		}
 
+		@Override
+		protected String generateIdInsertSelect(
+				String tableAlias, IdTableInfo idTableInfo, ProcessedWhereClause whereClause) {
+			String sql = super.generateIdInsertSelect( tableAlias, idTableInfo, whereClause );
+			return Parameters.process( sql );
```
gavinking, line 219: Why so many little local variables named `sql`?

gavinking:
Well, I agree that this approach is a lot better than what we wound up with when we tried to do it as part of the SQL *generation*.

Still a lot of places that get touched by this, however, and only going to be *more* once we get proper support for collections.

I would only do this if we know for sure it has a measurable impact on performance.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/bulk/impl/ReactiveBulkIdStrategy.java
DavideD: I find it easier for debugging

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/TableReactiveIdentifierGenerator.java
DavideD: Yes, you are right. I forgot to check that we are actually on Postgres before parsing. I'm working on it

DavideD:
I've updated the PR and it should work now. We can wait to merge it but at least we can compare the benchmarks.
\cc @Sanne

Sanne:
Very interesting, thanks. Tomorrow I'll check the performance impact.

But I'm surprised this was possible, are we sure that Reactive is overriding all SQL generation from ORM? I didn't expect that.

And it's sad that we have to override so many more methods from ORM. Remember we'll eventually want to define a proper SPI contract, to make this project maintaineable - which is why I expected that ORM core was going to need to expose an explict "post-SQL-generation-hook", or similar.

DavideD:
> are we sure that Reactive is overriding all SQL generation from ORM? I didn't expect that.

That's the drawback of this solution.We can only be sure of the use cases we test.
But these kind of bugs are quite easy to fix and we will find them soon enough.

DavideD:
> "post-SQL-generation-hook", or similar.

Let's first check if it's worth doing, then we can talk about a better solution.

Sanne:
Sure, that makes sense.

Sanne:
There's still some parsing happening at runtime:

```
   void java.lang.StringBuilder.<init>(int)	39
      void org.hibernate.reactive.pool.impl.Parameters$Parser.<init>(String, int)	27
         void org.hibernate.reactive.pool.impl.Parameters$Parser.<init>(String, int, Parameters$1)	27
         String org.hibernate.reactive.pool.impl.Parameters.process(String, int)	27
         CompletionStage org.hibernate.reactive.loader.ReactiveLoader.executeReactiveQueryStatement(String, QueryParameters, List, SharedSessionContractImplementor)	27
         CompletionStage org.hibernate.reactive.loader.ReactiveLoader.doReactiveQueryAndInitializeNonLazyCollections(String, SharedSessionContractImplementor, QueryParameters, boolean, ResultTransformer)	27
         CompletionStage org.hibernate.reactive.loader.ReactiveLoader.doReactiveQueryAndInitializeNonLazyCollections(String, SharedSessionContractImplementor, QueryParameters)	27
         CompletionStage org.hibernate.reactive.loader.entity.impl.ReactivePlanEntityLoader.load(Serializable, Object, SharedSessionContractImplementor, LockOptions, Boolean)	27
         CompletionStage org.hibernate.reactive.persister.entity.impl.ReactiveAbstractEntityPersister.reactiveLoad(Serializable, Object, LockOptions, SharedSessionContractImplementor, Boolean)	27
         CompletionStage org.hibernate.reactive.persister.entity.impl.ReactiveAbstractEntityPersister.reactiveLoad(Serializable, Object, LockOptions, SharedSessionContractImplementor)	27
         CompletionStage org.hibernate.reactive.event.impl.DefaultReactiveLoadEventListener.loadFromDatasource(LoadEvent, EntityPersister)	27
         CompletionStage org.hibernate.reactive.event.impl.DefaultReactiveLoadEventListener.doLoad(LoadEvent, EntityPersister, EntityKey, LoadEventListener$LoadType)	27
         CompletionStage org.hibernate.reactive.event.impl.DefaultReactiveLoadEventListener.load(LoadEvent, EntityPersister, EntityKey, LoadEventListener$LoadType)	27
```

gavinking:
Well that's a query, so I guess we expect that, no?

Sanne:
well I was hoping that after this, the prepares SQLs would be amended just once - and then reused after having been fixed.

This is not a Query but it's a load event; see the `ReactivePlanEntityLoader.load` method in the stack above. We should be able to use the SQL constant which is being generated when the EntityLoaders are first created?

DavideD:
I've updated this PR and it should now process the queries in advance when possible.

When filters are enabled I have to wait until they have been placed in the query before processing it and because this happens just before execution it means that the query cannot be processed in advance.

It seems there are some visible changes in performance when this patch is applied so, even if it requires some changes, it might be worthy.

@gavinking Could you have another look at this? Thanks
@Sanne maybe it's worthy to check this again considering that in master we have the other performance fixes applied

Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
```diff
@@ -13,70 +16,144 @@
...
+			return sql;
+		}
+
+		@Override
+		public String processLimit(String sql, Object[] parameterArray, boolean hasOffset) {
+			return sql;
+		}
+	};
+
+	public static Parameters create(Dialect dialect) {
```
gavinking, line 38: Could we rename this to, I dunno, `for()`, or `get()`, or `instance()`? Because it doesn't actually instantiate anything.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -160,17 +160,15 @@
 
 	public CompletionStage<RowSet<Row>> preparedQuery(String sql, Tuple parameters) {
 		feedback(sql);
-		String processedSql = usePostgresStyleParameters ? Parameters.process( sql, parameters.size() ) : sql;
```
gavinking, line 163: Can't we just remove the `usePostgresStyleParameters` field now?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveLoader.java
```diff
@@ -102,8 +110,23 @@
 		// Adding locks and comments.
 		sql = preprocessSQL( sql, queryParameters, session.getFactory(), afterLoadActions );
 
+		Object[] parameterArray = toParameterArray( queryParameters, session );
+
+		boolean hasFilter = session.getLoadQueryInfluencers().hasEnabledFilters();
+		if ( hasFilter ) {
+			// If the query is using filters, we know that they have been applied when we
+			// reach this point. Therefore is sage to process the query.
```
gavinking, line 118: "Therefore it's safe"

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/ReactiveLoader.java
```diff
@@ -102,8 +110,23 @@
...
+			sql = parameters().process( sql );
+		}
+		else if ( LimitHelper.useLimit( limitHandler, queryParameters.getRowSelection() ) ) {
+			// if the query is not using filters, it should have been preprocess before
```
gavinking, line 122: "preprocessed"

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -87,6 +88,11 @@
 public interface ReactiveAbstractEntityPersister extends ReactiveEntityPersister, OuterJoinLoadable, Lockable {
 	Logger log = Logger.getLogger( JoinedSubclassEntityPersister.class );
 
+	default Parameters parameters() {
+		Dialect dialect = getFactory().getJdbcServices().getDialect();
```
gavinking, line 92: I really hate unnecessary local variables. Especially this one.

gavinking (changes requested):
Looks fine. Left a few minor comments.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
DavideD: Good point, I will change it to `for`

DavideD:
Thanks @gavinking, I've applied all the remarks and it should be ready to be merged.
Should we wait for the result of the benchmark before actually doing it?

@Sanne^

gavinking:
Cheers

Sanne:
I can run it on the benchmark again tomorrow if you think that's useful. Sorry I was off today.

DavideD:
> I can run it on the benchmark again tomorrow if you think that's useful. Sorry I was off today.

I really don't want to make you work on a Saturday. I'm sure it can wait until next week.
Tale your time and relax a bit.

This project is not going to love you back :)

Sanne:
I've measured it, and looks very good. Allocation rate is down 10% *of the whole platform* , so that's significantly less garbage for the VM to handle.

Please apply it, if you think it's ready.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/519

@blafond Make sure to rebase your branch on top of this

DavideD:
Thanks

gavinking:
Great!

PR #523, opened 2021-01-25
don't throw exception when id is Short

qiuhw:
Hi, this patch fix the problem when id is Short the generation of id throws exception as `org.hibernate.HibernateException: cannot generate identifiers of type Short`.


gavinking:
OK, LGTM!

Thanks.

qiuhw:
Clode this due to the reason given here #524

PR #524, opened 2021-01-25
don't throw exception when id is Short

qiuhw:
Hi, this patch fix the problem when id is Short the generation of id throws exception as org.hibernate.HibernateException: cannot generate identifiers of type Short.

The other same titled pull request   #523   was forgot to import ShortType since I was modifying the code directly on GitHub UI as my local fork was still pulling down dependencies, don't know how to change the pull request there instead of this one, so I recreate this.


DavideD:
Thanks @qiuhw. Looks fine but I would like to add a test case before merging it. I will do it in a bit

DavideD:
I've added some tests to your original commit here: https://github.com/hibernate/hibernate-reactive/issues/526

I'll merge it once CI confirms that everything works.

Thanks a lot

PR #531, opened 2021-02-01
[#44] Add support for @ElementCollection

blafond:
Added support for List<String>, Set<String> and List<{Embeddable Entity}>

Tests include:
-  insert, remove, replace, clear elements
- duplicate and null elements.
- clearing of generated collection tables on delete of primary table

Will be logging issues to address support for
- @OrderColumn
- Map<> types ElementCollections
- Embeddable entity with a collection of elements
- additional tests for each

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRecreateAction.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRecreateAction.java
@@ -1,0 +1,79 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.engine.impl;
+
+import java.io.Serializable;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.HibernateException;
+import org.hibernate.action.internal.CollectionAction;
+import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.event.service.spi.EventListenerGroup;
+import org.hibernate.event.spi.EventType;
+import org.hibernate.event.spi.PostCollectionRecreateEvent;
+import org.hibernate.event.spi.PostCollectionRecreateEventListener;
+import org.hibernate.event.spi.PreCollectionRecreateEvent;
+import org.hibernate.event.spi.PreCollectionRecreateEventListener;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.reactive.engine.ReactiveExecutable;
+import org.hibernate.reactive.persister.collection.impl.ReactiveCollectionPersister;
+
+public class ReactiveCollectionRecreateAction extends CollectionAction implements ReactiveExecutable {
+
+	public ReactiveCollectionRecreateAction(
+			final PersistentCollection collection,
+			final CollectionPersister persister,
+			final Serializable key,
+			final SharedSessionContractImplementor session) {
+		super( persister, collection, key, session );
+	}
+
+	@Override
+	public CompletionStage<Void> reactiveExecute() {
+		final ReactiveCollectionPersister persister = (ReactiveCollectionPersister)getPersister();
+		final SharedSessionContractImplementor session = getSession();
+		final Serializable key = getKey();
+
+		final PersistentCollection collection = getCollection();
+
+		preRecreate();
+
+		return persister.recreateReactive( collection, key, session ).thenAccept( v -> {
+			session.getPersistenceContextInternal().getCollectionEntry( collection ).afterAction( collection );
+			evict();
+			postRecreate();
+		} );
+	}
+
+	@Override
+	public void execute() throws HibernateException {
+		// Unsupported in reactive see reactiveExecute()
+		throw new UnsupportedOperationException( "Use reactiveExecute() instead" );
+	}
+
+	private void preRecreate() {
+		final EventListenerGroup<PreCollectionRecreateEventListener> listenerGroup = listenerGroup( EventType.PRE_COLLECTION_RECREATE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PreCollectionRecreateEvent event = new PreCollectionRecreateEvent( getPersister(), getCollection(), eventSource() );
+		for ( PreCollectionRecreateEventListener listener : listenerGroup.listeners() ) {
+			listener.onPreRecreateCollection( event );
+		}
+	}
+
+	private void postRecreate() {
+		final EventListenerGroup<PostCollectionRecreateEventListener> listenerGroup = listenerGroup( EventType.POST_COLLECTION_RECREATE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PostCollectionRecreateEvent event = new PostCollectionRecreateEvent( getPersister(), getCollection(), eventSource() );
+		for ( PostCollectionRecreateEventListener listener : listenerGroup.listeners() ) {
+			listener.onPostRecreateCollection( event );
+		}
+	}
+}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
@@ -1,0 +1,183 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.engine.impl;
+
+import java.io.Serializable;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.AssertionFailure;
+import org.hibernate.HibernateException;
+import org.hibernate.action.internal.CollectionAction;
+import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.event.service.spi.EventListenerGroup;
+import org.hibernate.event.spi.EventType;
+import org.hibernate.event.spi.PostCollectionRecreateEvent;
+import org.hibernate.event.spi.PostCollectionRecreateEventListener;
+import org.hibernate.event.spi.PostCollectionRemoveEvent;
+import org.hibernate.event.spi.PostCollectionRemoveEventListener;
+import org.hibernate.event.spi.PreCollectionRecreateEvent;
+import org.hibernate.event.spi.PreCollectionRecreateEventListener;
+import org.hibernate.event.spi.PreCollectionRemoveEvent;
+import org.hibernate.event.spi.PreCollectionRemoveEventListener;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.reactive.engine.ReactiveExecutable;
+import org.hibernate.reactive.persister.collection.impl.ReactiveCollectionPersister;
+import org.hibernate.reactive.util.impl.CompletionStages;
+import org.hibernate.stat.spi.StatisticsImplementor;
+
+public class ReactiveCollectionRemoveAction extends CollectionAction implements ReactiveExecutable { ;
+	private final Object affectedOwner;
+	private final boolean emptySnapshot;
+
+	public ReactiveCollectionRemoveAction(
+			final PersistentCollection collection,
+			final CollectionPersister persister,
+			final Serializable key,
+			final boolean emptySnapshot,
+			final SharedSessionContractImplementor session) {
+		super( persister, collection, key, session );
+		if ( collection == null ) {
+			throw new AssertionFailure( "collection == null");
+		}
+		this.emptySnapshot = emptySnapshot;
+		// the loaded owner will be set to null after the collection is removed,
+		// so capture its value as the affected owner so it is accessible to
+		// both pre- and post- events
+		this.affectedOwner = session.getPersistenceContextInternal().getLoadedCollectionOwnerOrNull( collection );
+	}
+
+	@Override
+	public CompletionStage<Void> reactiveExecute() {
+		final Serializable key = getKey();
+		final SharedSessionContractImplementor session = getSession();
+		final ReactiveCollectionPersister reactivePersister = (ReactiveCollectionPersister) getPersister();
+		final CollectionPersister corePersister = getPersister();
+		final PersistentCollection collection = getCollection();
+		final StatisticsImplementor statistics = session.getFactory().getStatistics();
+
+		CompletionStage<Void> removeStage = CompletionStages.voidFuture();
+
+		if ( !emptySnapshot ) {
+			// an existing collection that was either non-empty or uninitialized
+			// is replaced by null or a different collection
+			// (if the collection is uninitialized, hibernate has no way of
+			// knowing if the collection is actually empty without querying the db)
+			removeStage = removeStage.thenAccept( v -> preRemove() )
+					.thenCompose( v -> reactivePersister
+							.removeReactive( key, session )
+							.thenAccept( ignore -> {
+								evict();
+								postRemove();
+							})
+					);
+		}
+		if( collection != null ) {
+			return removeStage.thenAccept(v -> {
+				session.getPersistenceContextInternal().getCollectionEntry( collection ).afterAction( collection );
+				evict();
+				postRemove();
+				if ( statistics.isStatisticsEnabled() ) {
+					statistics.updateCollection( corePersister.getRole() );
+				}
+			} );
+		}
+		return removeStage.thenAccept(v -> {
+			evict();
+			postRemove();
+			if ( statistics.isStatisticsEnabled() ) {
+				statistics.updateCollection( corePersister.getRole() );
+			}
+		} );
+
+	}
+
+//	@Override
+//	public int compareTo(Object o) {
+//		return 0;
+//	}
+
+//	@Override
+//	public Serializable[] getPropertySpaces() {
+//		return new Serializable[0];
+//	}
+//
+//	@Override
+//	public AfterTransactionCompletionProcess getAfterTransactionCompletionProcess() {
+//		return null;
+//	}
+//
+//	@Override
+//	public BeforeTransactionCompletionProcess getBeforeTransactionCompletionProcess() {
+//		return null;
+//	}
+//
+//	@Override
+//	public void afterDeserialize(SharedSessionContractImplementor session) {
+//
+//	}
+
+	@Override
+	public void execute() throws HibernateException {
+		// Unsupported in reactive see reactiveExecute()
+		throw new UnsupportedOperationException( "Use reactiveExecute() instead" );
+	}
+
+	private void preRemove() {
+		final EventListenerGroup<PreCollectionRemoveEventListener> listenerGroup = listenerGroup( EventType.PRE_COLLECTION_REMOVE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PreCollectionRemoveEvent event = new PreCollectionRemoveEvent(
+				getPersister(),
+				getCollection(),
+				eventSource(),
+				affectedOwner
+		);
+		for ( PreCollectionRemoveEventListener listener : listenerGroup.listeners() ) {
+			listener.onPreRemoveCollection( event );
+		}
+	}
+
+	private void postRemove() {
+		final EventListenerGroup<PostCollectionRemoveEventListener> listenerGroup = listenerGroup( EventType.POST_COLLECTION_REMOVE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PostCollectionRemoveEvent event = new PostCollectionRemoveEvent(
+				getPersister(),
+				getCollection(),
+				eventSource(),
+				affectedOwner
+		);
+		for ( PostCollectionRemoveEventListener listener : listenerGroup.listeners() ) {
+			listener.onPostRemoveCollection( event );
+		}
+	}
+
+	private void preRecreate() {
+		final EventListenerGroup<PreCollectionRecreateEventListener> listenerGroup = listenerGroup( EventType.PRE_COLLECTION_RECREATE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PreCollectionRecreateEvent event = new PreCollectionRecreateEvent( getPersister(), getCollection(), eventSource() );
+		for ( PreCollectionRecreateEventListener listener : listenerGroup.listeners() ) {
+			listener.onPreRecreateCollection( event );
+		}
+	}
+
+	private void postRecreate() {
+		final EventListenerGroup<PostCollectionRecreateEventListener> listenerGroup = listenerGroup( EventType.POST_COLLECTION_RECREATE );
+		if ( listenerGroup.isEmpty() ) {
+			return;
+		}
+		final PostCollectionRecreateEvent event = new PostCollectionRecreateEvent( getPersister(), getCollection(), eventSource() );
+		for ( PostCollectionRecreateEventListener listener : listenerGroup.listeners() ) {
+			listener.onPostRecreateCollection( event );
+		}
+	}
+
+}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
@@ -29,7 +29,7 @@ import org.hibernate.reactive.persister.collection.impl.ReactiveCollectionPersis
 import org.hibernate.reactive.util.impl.CompletionStages;
 import org.hibernate.stat.spi.StatisticsImplementor;
 
-public class ReactiveCollectionRemoveAction extends CollectionAction implements ReactiveExecutable { ;
+public class ReactiveCollectionRemoveAction extends CollectionAction implements ReactiveExecutable {
 	private final Object affectedOwner;
 	private final boolean emptySnapshot;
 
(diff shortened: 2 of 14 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+								locOffset = writeIndex( st, collection.getIndex( entry, i, this ), locOffset, session );
+							}
+							locOffset = writeElement( st, collection.getElement( entry ), locOffset, session );
+
+							collection.afterRowInsert( this, entry, i );
+						}
+				);
+				loop = loop.thenCompose( total -> reactiveConnection
+						.update( getSQLInsertRowString(), paramValues )
+						.thenApply( insertCount -> total + insertCount ) );
```
gavinking, line 297: I guess we had better trampoline this loop.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/collection/impl/ReactiveBatchingCollectionInitializerBuilder.java
```diff
@@ -24,7 +24,7 @@
 public abstract class ReactiveBatchingCollectionInitializerBuilder {
 
 	public static ReactiveBatchingCollectionInitializerBuilder getBuilder(SessionFactoryImplementor factory) {
-		switch ( factory.getSettings().getBatchFetchStyle() ) {
+		switch ( factory.getSessionFactoryOptions().getBatchFetchStyle() ) {
```
gavinking, line 27: What's the reason for this change?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRecreateAction.java
```diff
@@ -0,0 +1,79 @@
...
+import org.hibernate.event.spi.EventType;
+import org.hibernate.event.spi.PostCollectionRecreateEvent;
+import org.hibernate.event.spi.PostCollectionRecreateEventListener;
+import org.hibernate.event.spi.PreCollectionRecreateEvent;
+import org.hibernate.event.spi.PreCollectionRecreateEventListener;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.reactive.engine.ReactiveExecutable;
+import org.hibernate.reactive.persister.collection.impl.ReactiveCollectionPersister;
+
+public class ReactiveCollectionRecreateAction extends CollectionAction implements ReactiveExecutable {
```
gavinking, line 25: It seems to me that this should extend `CollectionRecreateAction`, to be consistent with what we do for entity-related actions, for example:

    public class ReactiveEntityDeleteAction extends EntityDeleteAction implements ReactiveExecutable

And that way we wouldn't need the copy/pasted `preRecreate()`/`postRecreate()` methods here.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
```diff
@@ -0,0 +1,183 @@
...
+import org.hibernate.event.spi.PreCollectionRecreateEventListener;
+import org.hibernate.event.spi.PreCollectionRemoveEvent;
+import org.hibernate.event.spi.PreCollectionRemoveEventListener;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.reactive.engine.ReactiveExecutable;
+import org.hibernate.reactive.persister.collection.impl.ReactiveCollectionPersister;
+import org.hibernate.reactive.util.impl.CompletionStages;
+import org.hibernate.stat.spi.StatisticsImplementor;
+
+public class ReactiveCollectionRemoveAction extends CollectionAction implements ReactiveExecutable { ;
```
gavinking, line 32: Likewise, this would extend `CollectionRemoveAction`. (Unless there's something I'm missing.)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+									loc = writeIndex( st, collection.getIndex( entry, i, this ), loc, session );
+								}
+								loc = writeElement( st, collection.getElement( entry ), loc, session );
+
+								collection.afterRowInsert( this, entry, i );
+							}
+					);
+					loop = loop.thenCompose( total -> reactiveConnection
+							.update( getSQLInsertRowString(), paramValues )
+							.thenApply( insertCount -> total + insertCount ) );
```
gavinking, line 127: This should also be a trampoline, I guess.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+								else {
+									writeElementToWhere( st, entry, loc, session );
+								}
+							}
+						}
+				);
+				loop = loop.thenCompose( total -> reactiveConnection
+						.update( getSQLDeleteRowString(), paramValues )
+						.thenApply( deleteCount -> total + deleteCount ) );
+			}
```
gavinking, line 201: Trampoline.

gavinking (changes requested):
Excellent, this all looks very good to me, except I would like to see

1. use of trampolining via `CompletionStages.loop()`
2. inheritance of `Action`s from the non-reactive versions in order to eliminate copy/paste.


Thanks!!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
DavideD: See [CompletionStages#loop](https://github.com/hibernate/hibernate-reactive/blob/4844db0393e3738a5185da0ab3c709049a27137d/hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java#L155)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+							.thenApply( insertCount -> total + insertCount ) );
+				}
+			}
+		}
+
+		return loop.thenAccept( total -> LOG.debugf( "Done inserting collection: %s rows inserted", total ) );
+	}
+
+	public CompletionStage<Integer> removeReactive(Serializable id, SharedSessionContractImplementor session)
+			throws HibernateException {
```
DavideD, line 136: Shouldn't we be consistent with ORM and return `CompletionStage<Void>` instead?
See [AbstractCollectionPersister#remove](https://github.com/hibernate/hibernate-orm/blob/20f6d4f5dbccd352d5a3e219121136cc53506f89/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L1206)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	@Override
+	public CompletionStage<Void> reactiveUpdateRows(
+			PersistentCollection collection,
+			Serializable id,
+			SharedSessionContractImplementor session) throws HibernateException {
+
+		// Currently there is no "updateRows()" method in AbstractCollectionPersister
+		// Not sure if this method, reactiveUpdateRows() is needed.
```
DavideD, line 214: This comment is wrong: [BasicCollectionPersister#updateRows](https://github.com/hibernate/hibernate-orm/blob/5.4/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L1746)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForEmbeddableEntityTypeListTest.java
```diff
@@ -0,0 +1,815 @@
...
+	@Embeddable
+	public static class Phone {
+
+		private String number;
+
+		private String country;
+
+		// FIXME: add nested @Embeddable
+		// @ElementCollection(fetch = FetchType.EAGER)
+		// private List<AlternativePhone> alternatives = new ArrayList<>();
```
DavideD, line 768: Can you remove this comments from the version we are going to merge?

DavideD: This is probably going to be a different test class anyway

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,303 @@
...
+
+		if ( !isInverse && collection.isRowUpdatePossible() ) {
+
+			LOG.debugf( "Updating rows of collection: %s#%s", getNavigableRole().getFullPath(), id );
+
+			// update all the modified entries
+			// NOTE this method call uses a JDBC connection and will fail for Map ElementCollection type
+			// May also have to override getSQLDeleteString() to correctly parameterize to handle the SQL for the PG use case
+			// SEE getSQLInsertRowString() in this class for an example
+			int count = doUpdateRows( id, collection, session );
```
DavideD, line 230: This is not the correct implementation anyway so it's probably going to fail if somebody reaches this code. I think we should throw an `UnsupportedOperationException` instead like we do in reactiveDeleteRows.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
```diff
@@ -55,4 +70,85 @@ protected ReactiveCollectionInitializer createSubselectInitializer(SubselectFetc
...
+			PersistentCollection collection, Serializable id, SharedSessionContractImplementor session) {
+		return CompletionStages.voidFuture();
+	}
+
+	private ReactiveConnection getReactiveConnection(SharedSessionContractImplementor session) {
+		return ( (ReactiveConnectionSupplier) session ).getReactiveConnection();
+	}
+
+	@Override
+	public CompletionStage<Integer> removeReactive(Serializable id, SharedSessionContractImplementor session)
```
DavideD, line 85: It should return `CompletionStage<Void>` like in ORM

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
```diff
@@ -55,4 +70,85 @@ protected ReactiveCollectionInitializer createSubselectInitializer(SubselectFetc
...
+			Expectation expectation = Expectations.appropriateExpectation( getDeleteAllCheckStyle() );
+			boolean callable = isDeleteAllCallable();
+
+			List<Object> params = new ArrayList<>();
+			params.add( id );
+			String sql = getSQLDeleteString();
+
+			return reactiveConnection.update( sql, params.toArray( new Object[0] ) );
+		}
+		return CompletionStages.completedFuture( -1 );
```
DavideD, line 109: Even if we decide to return a `CompletionStage<Integer>` in this case, ORM returns 0 when it doesn't update any row. Not sure we should return -1 here.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
```diff
@@ -55,4 +70,85 @@ protected ReactiveCollectionInitializer createSubselectInitializer(SubselectFetc
...
+			return CompletionStages.voidFuture();
+		}
+
+		if ( !isRowDeleteEnabled() ) {
+			return CompletionStages.voidFuture();
+		}
+
+		// FIXME: Not sure what to implement here for OneToMany use-case
+		// Maybe nothing due to the ReactiveCollectionUpdate/Remove/Recreate Action classes perform necessary operations
+		// on the collection (basically delete and replace rather than operating on individual rows
```
DavideD, line 141: I think you can remove these comments, You had to add these methods because they are in the interface and I suspect, I'm not sure, we are going to work on this when adding support for `@ManyToMany` or tests associations with maps

DavideD: I think this would be enough:
```
// TODO: See AbstractCollectionPersister#insertRows
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Similar to {@link org.hibernate.persister.collection.AbstractCollectionPersister} in ORM
+ * <p>
+ *     The reactive classes already extend something so we cannot have an abstract class like in ORM.
+ * </p>
+ */
+public interface ReactiveCollectionPersister {
+
+	/**
+	 * Reactive version of {@link org.hibernate.persister.collection.AbstractCollectionPersister}.recreate()
```
DavideD, line 23: Is this a common way to link to the javadoc?
I usually do:
```
{@link org.hibernate.persister.collection.AbstractCollectionPersister#recreate}
```
Same for the rest of the javadoc.

DavideD: We could also remove the parameters from the javadoc since they don't have any helpful comment (I don't think they need one).

DavideD: This also applies to the javadoc of the other methods.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	CompletionStage<Void> recreateReactive(PersistentCollection collection, Serializable id, SharedSessionContractImplementor session);
+
+	/**
+	 * Reactive version of {@link org.hibernate.persister.collection.AbstractCollectionPersister}.remove()
+	 * @param id
+	 * @param session
+	 * @return Integer completion stage
+	 */
+	CompletionStage<Integer> removeReactive(Serializable id, SharedSessionContractImplementor session);
```
DavideD, line 37: This is the only method to return `CompletionStage<Integer>` instead of `CompletionStage<Void>` and it's also inconsistent with ORM.

DavideD (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/collection/impl/ReactiveBatchingCollectionInitializerBuilder.java
blafond: Just to fix the deprecation of getSettings(). I can change it back.

gavinking: Oh I see. OK, that's fine.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRecreateAction.java
blafond: This will require a change to the ORM actions which are currently marked as ```final```.  I'll test locally and create a PR that'll have to be back-ported to 5.4 at some point.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
blafond: got it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java
blafond: The best I can tell is:   ```{@link org.hibernate.persister.collection.AbstractCollectionPersister#insertRows()}```

thx

DavideD: As long as it is clickable

blafond: I can't find a way to create a clickable link to a method.

Best I could do is: ```   * Reactive version of {@link org.hibernate.persister.collection.AbstractCollectionPersister} recreate() method```

gavinking: links to methods need to list the parameter types.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionRemoveAction.java
```diff
@@ -0,0 +1,183 @@
...
+//
+//	@Override
+//	public BeforeTransactionCompletionProcess getBeforeTransactionCompletionProcess() {
+//		return null;
+//	}
+//
+//	@Override
+//	public void afterDeserialize(SharedSessionContractImplementor session) {
+//
+//	}
```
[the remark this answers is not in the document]
blafond: yes

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveCollectionUpdateAction.java
```diff
@@ -0,0 +1,211 @@
...
+//
+//	@Override
+//	public BeforeTransactionCompletionProcess getBeforeTransactionCompletionProcess() {
+//		return null;
+//	}
+//
+//	@Override
+//	public void afterDeserialize(SharedSessionContractImplementor session) {
+//
+//	}
```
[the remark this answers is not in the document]
blafond: yes

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForEmbeddableEntityTypeListTest.java
blafond: yes.. removed.  Already started on setting up the different test class.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -260,44 +211,96 @@ protected String getSQLDeleteString() {
...
-							}
-							locOffset = writeElement( st, collection.getElement( entry ), locOffset, session );
+	/*
+	 * generates and returns parameter values for a reactiveRecreate() and reactiveInsertRows() methods
+	 * compatible with getSQLInsertRowString() and query string
+	 */
+	private Object[] paramValues(Object entry, PersistentCollection collection, Serializable id, SharedSessionContractImplementor session) {
+		return PreparedStatementAdaptor.bind(
+				st -> {
+					int i = 0;
```
DavideD, line 266: Something is wrong here.
In the ORM code the index `i` was incremented in the loop: https://github.com/hibernate/hibernate-orm/blob/cfc7b972501bf26713098abbbc36265a9b1ed46d/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L1614

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -260,44 +211,96 @@ protected String getSQLDeleteString() {
...
+					int loc = writeKey( st, id, 1, session );
```
DavideD, line 267: In ORM, offset is initialized like this:
```
offset += expectation.prepare( st );
```
Shouldn't we do the same or something similar?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
```diff
@@ -106,7 +106,7 @@ private ReactiveConnection getReactiveConnection(SharedSessionContractImplemento
 
 			return reactiveConnection.update( sql, params.toArray( new Object[0] ) );
 		}
-		return CompletionStages.completedFuture( -1 );
+		return CompletionStages.completedFuture( 0  );
```
DavideD, line 109: Shouldn't this return `CompletionStage<Void>`?

DavideD (commented):
I have some questions about these new changes.
Not the end of the world but the commit message is  a bit misleading, there are other changes in this commit: it's not only about the trampoline.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
blafond: only reason is the ```reactiveConnection.update( sql, params.toArray( new Object[0] ) );``` returns ```CompletionStage<Integer>```

Guess we could convert to Void?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java
blafond: k, I'll work through the changes

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java
DavideD: OK, but all the other methods are Void, why this should be different?
I think I mentioned this inconsistency before. The method in `ReactiveCollectionPersister` should return `CompletionStage<Void>`: https://github.com/hibernate/hibernate-reactive/blob/384eb90a3a7e24f6e2b14e87d8d98689e4137075/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java#L31

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveCollectionPersister.java
blafond: fixed and pushed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
blafond: looking

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,308 @@
...
+
+		CompletionStage<Integer> loop = CompletionStages.zeroFuture();
+		ReactiveConnection reactiveConnection = getReactiveConnection( session );
+
+		return CompletionStages.total(
+				entryArray(collection),
+				entry -> loop.thenCompose( total -> reactiveConnection
+						.update( getSQLInsertRowString(), paramValues( entry, collection, id ,session ) )
+						.thenApply( insertCount -> total + insertCount ) ) )
+				.thenAccept( total -> LOG.debugf( "Done inserting rows: %s inserted", total ) );
```
DavideD, line 225: I haven't tested it, but I think this should be:
```
		return CompletionStages
				.total( entryArray( collection ), 
						entry -> reactiveConnection.update( getSQLInsertRowString(), paramValues( entry, collection, id,session ) ) )
				.thenAccept( total -> LOG.debugf( "Done inserting rows: %s inserted", total ) );
```

DavideD: By the way, if you create a `total` method in `CompletionStages` that accepts an Iterator or Iterable you can remove the conversion you need to do in `entryArray`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -0,0 +1,308 @@
...
+		ReactiveConnection reactiveConnection = getReactiveConnection( session );
+
+		boolean deleteByIndex = hasIndex && !indexContainsFormula;
+
+		return CompletionStages.total(
+				entryArray( collection, deleteByIndex ),
+				entry -> loop.thenCompose( total -> reactiveConnection
+						.update( getSQLDeleteRowString(), paramValues( entry, id ,session, deleteByIndex ) )
+						.thenApply( deleteCount -> total + deleteCount ) ) )
+				.thenAccept( total -> LOG.debugf( "Done removing rows: %s removed", total ) );
```
DavideD, line 168: return CompletionStages
				.total( entryArray( collection, deleteByIndex ),
						entry -> reactiveConnection.update( getSQLDeleteRowString(), paramValues( entry, id ,session, deleteByIndex ) ) )
				.thenAccept( total -> LOG.debugf( "Done removing rows: %s removed", total ) );

DavideD: You can remove entryArray if you add a new `total` method in `CompletionStages` accepting an iterable

DavideD (changes requested):

blafond:
Closing.. replaced by https://github.com/hibernate/hibernate-reactive/pull/549

PR #535, opened 2021-02-02
[#508] Fix ReactivePlanEntityLoader#hasFilters

DavideD:
It should return true when there are filters


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/535

Issue #398, opened 2020-10-04
finish javadoc for org.hibernate.reactive.provider.service package


gavinking:
Some of the classes in this package don't have any Javadoc at all. And in a couple of cases it's non obvious what they are for or when they're used.

@Sanne I think most of these are your classes, though a couple of them are probably mine.

Sanne:
thanks, I can polish these next week.

gavinking:
Great, thanks.

DavideD:
@Sanne any update on this? If you are busy I can take care of it

Sanne:
@DavideD yes if you could steal this from me that would be best. Thanks!

gavinking:
Thanks man.

PR #548, opened 2021-02-08
[#398] Update javadoc

DavideD:
I've added the javadoc to the classes in org.hibernate.reactive.provider.


hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveQueryTranslatorFactory.java
```diff
@@ -14,6 +14,11 @@
 import org.hibernate.hql.spi.QueryTranslatorFactory;
 import org.hibernate.reactive.session.impl.ReactiveQueryTranslatorImpl;
 
+/**
+ * Facade for generation the generation of reactive {@link QueryTranslator} and {@link FilterTranslator} instances.
```
gavinking, line 18: "for generation the generation"

gavinking (changes requested):

DavideD:
Thanks.

I've fixed the javadoc and merge it

PR #557, opened 2021-02-10
[#556] Fix parameter processing when query has filters and pagination

### LINKED ISSUE #556: Parameter index clash when combining filters and limits
As reported in https://github.com/quarkusio/quarkus/issues/14779#issuecomment-775970486, we have a query like `SELECT FROM Foo`, in this case:

```java
Person.findAll(Sort.by("id")).filter("Person.isAlive").filter("Person.hasName",
                            Parameters.with("name", "Stef"));
```

And we have the following filters defined:

```java
@FilterDef(name = "Person.hasName", defaultCondition = "name = :name", parameters = @ParamDef(name = "name", type = "string"))
@FilterDef(name = "Person.isAlive", defaultCondition = "status = 'LIVING'")
@Filter(name = "Person.isAlive")
@Filter(name = "Person.hasName")
```

and the Panache layer calls:

```java
        jpaQuery.setMaxResults(maxResults);
...
            Filter filter = em.enableFilter(entry.getKey());
            for (Entry<String, Object> paramEntry : entry.getValue().entrySet()) {
                filter.setParameter(paramEntry.getKey(), paramEntry.getValue());
            }
            filter.validate();
```

So using `setMaxResults`, and filters causes the query to be translated to:

```
select person0_.id as id1_16_, person0_.address_id as address_5_16_, person0_.name as name2_16_, person0_.status as status3_16_, person0_.uniqueName as uniquena4_16_ from Person2 person0_ where person0_.status = 'LIVING' and person0_.name = $1 order by person0_.id limit $1
```

If you notice, the filter param value passed by name turns into an indexed param which clashes with the indexed parameter used for the `limit`.

DavideD:
Fixes #556

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/hql/impl/ReactiveQueryLoader.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/hql/impl/ReactiveQueryLoader.java
@@ -114,12 +114,20 @@ public class ReactiveQueryLoader extends QueryLoader implements CachingReactiveL
 		final boolean cacheable = factory.getSessionFactoryOptions().isQueryCacheEnabled()
 				&& queryParameters.isCacheable();
 
-		String processedSQL = parameters().process( getSQLString() );
+		// Hibernate Reactive process filters just before running the query.
+		// Our parameters processor is not smart enough to work on the same query twice,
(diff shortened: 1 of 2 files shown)


DavideD:
Actually I want to change a couple of things before merging it....

PR #559, opened 2021-02-10
[#556] Fix parameter processing when query has filters and pagination 

### LINKED ISSUE #556: Parameter index clash when combining filters and limits

DavideD:
Fixes #556

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
@@ -1,0 +1,302 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.Objects;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.Id;
+import javax.persistence.NamedQuery;
+
+import org.hibernate.annotations.Filter;
+import org.hibernate.annotations.FilterDef;
+import org.hibernate.annotations.ParamDef;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.containers.DatabaseConfiguration;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.stage.Stage;
+import org.hibernate.reactive.testing.DatabaseSelectionRule;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import io.vertx.ext.unit.TestContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test the combination of filters, max results and first result.
+ */
+public class FilterWithPaginationTest extends BaseReactiveTest {
+
+	/**
+	 * There's probably a bug in the Vert.x Db2 SQL client.
+	 * See: https://github.com/eclipse-vertx/vertx-sql-client/issues/887
+	 */
+	@Rule
+	public DatabaseSelectionRule skipDB2 = DatabaseSelectionRule.skipTestsFor( DatabaseConfiguration.DBType.DB2 );
+
+	FamousPerson margaret = new FamousPerson( 1L, "Margaret Howe Lovatt", Status.LIVING, "the woman who lived in a half-flooded home with a dolphin.");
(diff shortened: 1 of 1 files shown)


DavideD:
Something doesn't work with MySQL... I'm checking

DavideD:
It seems there were conflicts with existing tests because I'm using an entity person.
I will create a separate issue for it

gavinking:
> It seems there were conflicts with existing tests because I'm using an entity person.

just add a `@Table` annotation and done.

DavideD:
You'd think that would be enough but it's not

DavideD:
Now something is failing with DB2 ... uff

DavideD:
> just add a @Table annotation and done

Sorry, you meant a different table name than `Person`, yes, that might have worked.

gavinking:
Right, exactly.

DavideD:
I think I've found a bug in the vert.x DB2 SQL client: https://github.com/eclipse-vertx/vertx-sql-client/issues/887

@mkouba Maybe you can confirm this?

DavideD:
For now I'm going to skip the tests for DB2

DavideD:
I think this can be merged now

hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/hql/impl/ReactiveQueryLoader.java
```diff
@@ -114,12 +114,20 @@ public Parameters parameters() {
 		final boolean cacheable = factory.getSessionFactoryOptions().isQueryCacheEnabled()
 				&& queryParameters.isCacheable();
 
-		String processedSQL = parameters().process( getSQLString() );
+		// Hibernate Reactive process filters just before running the query.
+		// Our parameters processor is not smart enough to work on the same query twice,
+		// so we wait to process a query if filters are enabled. It's going to be process
+		// just before the execution
+		boolean hasFilters = session.getLoadQueryInfluencers().hasEnabledFilters();
```
gavinking, line 121: @DavideD this looks perfectly fine to me

except you know how I feel about unnecessary local variables 😡😁

DavideD: I swear I'm not doing on purpose : -D

It's just that the next line reads so much better with a shorter variable

gavinking: Haha I'm one of those crazy people who thinks "more readable" ≈ "less tokens".

DavideD: I suppose it could have been a method.... well at some point I want to change this part anyway to make it easier to track where we need this stuff.

gavinking: Yeah FTR I'm also always happy to move the tokens somewhere else. 

Extract Method >> Extract Variable.

DavideD: Alright, gimme a sec

DavideD:
I've removed the local variable, rebased and Squash the commits related to the test together.

gavinking:
Cheers.

PR #563, opened 2021-02-11
orphan removal

### LINKED ISSUE #506: orphan removal unsupported
Hi, we are facing an issue when using `orphanRemoval`, cascade goes into `onDelete` instead of `reactiveOnDelete`, and it throws an `UnsupportedOperationException`

## Example

Product
```
@Entity
@Table(name = "PRODUCT")
public class Product {

    @ManyToOne(targetEntity = Shop.class, cascade = MERGE)
    @JoinColumn(name = "SHOP_ID", referencedColumnName = "ID")
    private Shop shop;
   
    ...
}
```

Shop
```
@Entity
@Table(name = "SHOP")
public class Shop {

    @OneToMany(mappedBy = "shop", cascade = ALL, fetch = EAGER, orphanRemoval = true)
    private List<Product> productList = new ArrayList<>();

    ...
}
```

Main application
```
EntityManagerFactory emf = Persistence.createEntityManagerFactory("mysql");
Stage.SessionFactory sessionFactory = emf.unwrap(Stage.SessionFactory.class);

sessionFactory.withTransaction((session, transaction) -> {

        Shop s1 = Shop.builder().name("s1").build();
        s1.addProduct(Product.builder().name("ap1").build());
        s1.addProduct(Product.builder().name("ap2").build());
        s1.addProduct(Product.builder().name("ap3").build());
        s1.addProduct(Product.builder().name("ap4").build());

        // create
        return session.merge(s1)
                    .thenCombine(session.flush(), (s, v) -> s);
})
.thenCompose(t1 -> {
        return sessionFactory.withTransaction((session, transaction) -> {
            
            // get
            return session.find(Shop.class, t1.getId())
                    .thenCompose(s2 -> {

                        // update
                        s2.setName("s2");
                        s2.getProductList().clear();
                        s2.addProduct(Product.builder().name("bp5").build());
                        s2.addProduct(Product.builder().name("bp6").build());
                        s2.addProduct(Product.builder().name("bp7").build());

                        // update
                        return session.merge(s2);
                    })
                    .thenCombine(session.flush(), (s, v) -> s);
        });
});
```

## Stacktrace

```
java.util.concurrent.CompletionException: javax.persistence.PersistenceException: org.hibernate.HibernateException: javax.persistence.PersistenceException: org.hibernate.HibernateException: java.lang.UnsupportedOperationException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:824)
	at java.util.concurrent.CompletableFuture.uniHandleStage(CompletableFuture.java:834)
	at java.util.concurrent.CompletableFuture.handle(CompletableFuture.java:2155)
	at java.util.concurrent.CompletableFuture.handle(CompletableFuture.java:110)
	at org.hibernate.reactive.session.impl.ReactiveSessionImpl.doFlush(ReactiveSessionImpl.java:800)
	at org.hibernate.reactive.session.impl.ReactiveSessionImpl.reactiveAutoflush(ReactiveSessionImpl.java:787)
	at org.hibernate.reactive.stage.impl.StageSessionImpl$Transaction.flush(StageSessionImpl.java:409)
	at org.hibernate.reactive.stage.impl.StageSessionImpl$Transaction.lambda$execute$2(StageSessionImpl.java:391)
	at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:952)
	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
	...
Caused by: java.lang.UnsupportedOperationException
	at org.hibernate.reactive.event.impl.DefaultReactiveDeleteEventListener.onDelete(DefaultReactiveDeleteEventListener.java:86)
	at org.hibernate.event.service.internal.EventListenerGroupImpl.fireEventOnEachListener(EventListenerGroupImpl.java:113)
	at org.hibernate.internal.SessionImpl.fireDelete(SessionImpl.java:924)
	... 93 more
```

## Sample application

https://github.com/pqab/orphanRemoval-example

## Environment Information

- Version: 1.0.0.Alpha11

gavinking:
Probably needs more tests.

Fixes  #506

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
@@ -1,0 +1,130 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import io.vertx.ext.unit.TestContext;
+import org.hibernate.cfg.Configuration;
+import org.junit.Test;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import java.util.HashSet;
+import java.util.Set;
+
+import static javax.persistence.CascadeType.PERSIST;
+import static javax.persistence.CascadeType.REMOVE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class OrphanRemovalTest extends BaseReactiveTest {
+
+    @Override
+    protected Configuration constructConfiguration() {
+        Configuration configuration = super.constructConfiguration();
+        configuration.addAnnotatedClass( Version.class );
+        configuration.addAnnotatedClass( Product.class );
+        configuration.addAnnotatedClass( Shop.class );
+        return configuration;
+    }
+
+    @Test
+    public void testOrphan(TestContext context) {
+        Shop shop = new Shop("shop");
+        Product product = new Product("ap1", shop);
+        product.addVersion(new Version(product));
+        shop.addProduct(product);
+        shop.addProduct(new Product("ap2", shop));
+        shop.addProduct(new Product("ap3", shop));
+        shop.addProduct(new Product("ap4", shop));
+
+        test(context,
+                getSessionFactory()
+                        .withTransaction((session, transaction) -> session.persist(shop))
+                        .thenCompose(v -> getSessionFactory().withTransaction((session, transaction) -> session.find(Shop.class, shop.id)
+                                .thenCompose( result -> session.fetch(result.products).thenApply(products -> result) )
+                                .thenAccept(result -> {
+                                    // update
+                                    result.products.clear();
+                                    result.addProduct(new Product("bp5", result));
+                                    result.addProduct(new Product("bp6", result));
+                                    result.addProduct(new Product("bp7", result));
+                                })
+                        ))
+                        .thenCompose(v -> getSessionFactory().withTransaction((session, transaction) ->
+                                session.createQuery("select count(*) from Product", Long.class)
+                                        .getSingleResult().thenAccept(result -> context.assertEquals(3l, result))
+                        ))
+                        .thenCompose(v -> getSessionFactory().withTransaction((session, transaction) ->
+                                session.createQuery("select name from Product" )
+                                        .getResultList().thenAccept(list -> assertThat( list ).containsExactlyInAnyOrder( "bp5", "bp6", "bp7"))
+                        ))
+        );
+    }
+
+    @Entity(name = "ProductVersion")
+    public static class Version {
+        @Id @GeneratedValue
+        private long id;
+
+        @ManyToOne
+        private Product product;
+
+        public Version(Product product) {
+            this.product = product;
+        }
+        Version() {}
+    }
+
+    @Entity(name = "Product")
+    public static class Product {
+
+        @Id @GeneratedValue
+        private long id;
+        private String name;
+
+        Product(String name, Shop shop) {
+            this.name = name;
+            this.shop = shop;
+        }
+
+        Product() {}
+
+        @ManyToOne
+        private Shop shop;
+
+        @OneToMany(mappedBy = "product", cascade = {PERSIST, REMOVE})
+        private Set<Version> versions = new HashSet<>();
+
+        public void addVersion(Version version) {
+            versions.add(version);
+        }
+    }
+
+    @Entity(name = "Shop")
+    public static class Shop {
+
+        @Id @GeneratedValue
+        private long id;
+        private String name;
+
+        Shop(String name) {
+            this.name = name;
+        }
+
+        Shop() {}
+
+        @OneToMany(mappedBy = "shop", cascade = {PERSIST, REMOVE}, orphanRemoval = true)
+        private Set<Product> products = new HashSet<>();
+
+        public void addProduct(Product product) {
+            products.add(product);
+        }
+    }
+
+}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
@@ -47,18 +47,18 @@ public class ReactiveActionQueue {
 	static {
 		EXECUTABLE_LISTS_MAP = new LinkedHashMap<>( 8 );
 
-//		EXECUTABLE_LISTS_MAP.put(
-//				OrphanRemovalAction.class,
-//				new ListProvider<OrphanRemovalAction>() {
-//					ExecutableList<OrphanRemovalAction> get(ReactiveActionQueue instance) {
-//						return instance.orphanRemovals;
-//					}
-//					ExecutableList<OrphanRemovalAction> init(ReactiveActionQueue instance) {
-//						// OrphanRemovalAction executables never require sorting.
-//						return instance.orphanRemovals = new ExecutableList<>( false );
-//					}
-//				}
-//		);
+		EXECUTABLE_LISTS_MAP.put(
+				ReactiveOrphanRemovalAction.class,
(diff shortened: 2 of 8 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
```diff
@@ -0,0 +1,102 @@
...
+                                    // update
+                                    result.productList.clear();
+                                    result.addProduct(new Product("bp5", result));
+                                    result.addProduct(new Product("bp6", result));
+                                    result.addProduct(new Product("bp7", result));
+                                })
+                        ))
+                        .thenCompose(v -> getSessionFactory().withTransaction((session, transaction) ->
+                                session.createQuery("select count(*) from Product", Long.class)
+                                        .getSingleResult().thenAccept(result -> context.assertEquals(3l, result))
```
DavideD, line 57: Would it be too much to ask to check the content of the list and not just the size?

DavideD: I mean, the content of the table

gavinking: Well, if it's of length 3 instead of 4 or 7, I think we can be pretty certain orphan-delete worked.

Remember we're not trying to write tests for the whole of Hibernate core here.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
```diff
@@ -0,0 +1,126 @@
...
+            this.name = name;
+            this.shop = shop;
+        }
+
+        Product() {}
+
+        @ManyToOne
+        private Shop shop;
+
+        @OneToMany(mappedBy = "product", cascade = {PERSIST, REMOVE}, orphanRemoval = true, fetch = FetchType.EAGER)
```
gavinking, line 97: So there's still a slight problem here. If I remove the `EAGER` fetch here, then removal doesn't cascade. Need to look into that further.

gavinking: Note: that's not an orphan removal; it's a cascade removal *from* the orphan.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
DavideD: I don't agree with this but OK

gavinking: Well I guess I'm not sure what it is you want me to test?

That we didn't delete the wrong `Product`s? How could it possibly happen that we deleted the wrong four ids by accident, unless there was something seriously very broken in core?

Or something else?

DavideD: > How could it possibly happen that we deleted the wrong four ids by accident, unless there was something seriously very broken in core

This is actually why I write tests, make sure that somebody at some point doesn't mess up things.
We also don't always use the exact ORM code, or we might not use it in the future and having some more accurate tests  now makes me sleep better at night.

In this case in particular, it's not difficult to check the content of the list with an assertion after changing the `select count(*)` to a `select  name`.
Not sure why you are so against it, but I'm OK merging this if you feel so strong about it.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
gavinking: Alright, I know what's going on here now.

gavinking: So the original core code calls:

    entityResolver.resolve( (EntityType) types[i], value, session, entity, overridingEager )

But we call:

    ((ReactiveSession) session1).reactiveGet( entityType.getReturnedClass(), (Serializable) value )

which returns `null` when the entity has been scheduled for deletion.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
gavinking: I think we have to be able to trust in the Hibernate test suite (which is *massive*). If we start adding all sorts of tests for things that aren't implemented in Hibernate Reactive itself, we're going to quickly end up with a test suite that is as slow as the Hibernate test suite. I know that's not a problem for us *now*, since the tests run really fast *for now* but you would be surprised how quick and easy it is to bloat out a test suite and eventually find yourself always waiting for around for the build to finish. (This has happened to me over and over.)

Adding tests for all sorts of random things that are unlikely to break isn't smart. Smart is spending time thinking up things which *are* likely to break, and which *are* implemented with code in HR, and writing tests for those things.

Like here: instead of writing a more verbose test for the thing I could already see was working for sure, I asked myself "well, what happens to the associations belonging to the orphan itself?" and I wrote a test for that instead.

And guess what: buried treasure! My intuition lead me to something that was really quite broken.

gavinking:
Alright, so I found something that was pretty broken. And what I've done to fix it almost surely fixes a several other bugs that we have just been lucky enough to not run into yet. Sadly, I'm not quite sure where to find them, so I don't know how to write tests for them, which sucks.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
DavideD: Agree to disagree.

And I don't find this test to be much more verbose:
```
session.createQuery("select name from Product", Long.class)
     .getResultList().thenAccept( list -> assertThat( list ).containsExactlyInAnyOrder( "bp5", "bp6", "bp7") )
```

But as I said before. Feel free to merge it.

DavideD:
> Sadly, I'm not quite sure where to find them, so I don't know how to write tests for them, which sucks.

Sorry, I don't think I understand this sentence. Do you mean you don't know how to test for some bugs?
As long as this issue is solved I think we are good. We can add some tests later.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrphanRemovalTest.java
DavideD: > And guess what: buried treasure! My intuition lead me to something that was really quite broken.

I cannot run your intuition on CI ... yet :)

gavinking: I don't know what `assertThat()` is. It's not a method of `TestContext`.

DavideD:
```
import static org.assertj.core.api.Assertions.assertThat;
```

gavinking: `Assertions.assertThat` apparently.

gavinking: OK, I added that assertion.

gavinking:
> Do you mean you don't know how to test for some bugs?

I have no clear idea what the bugs I fixed even _are_.

DavideD:
> I have no clear idea what the bugs I fixed even are.

:-D
I'm not too worry about bugs we are not aware of ... I happy to have tests for the one we know.

DavideD:
Actually... maybe @gbadner can help with this? About how to test for those bugs

DavideD:
> Actually... maybe @gbadner can help with this? About how to test for those bugs

or better... for the code you've changed

gavinking:
Looking again, I think the main thing is something to do with null one-to-one associations.

Anyway the bug I was trying to fix is fixed so I'm going to merge this.

PR #564, opened 2021-02-11
Upgrade to testcontainers 1.15.2

DavideD:


gavinking:
@DavideD this change seems to have broken my ability to run tests on MySQL from within IntelliJ. (It _does_ still work from gradle.)

Better roll it back for now.

gavinking:
Excuse me, I might be wrong. Now it's not working on the older version either (it was a few minutes ago). WTF.

gavinking:
Can someone else please try running the tests in IntelliJ with `-Pdocker -Pdb=mysql` because it's really quite broken for me.

gavinking:
FTR it works on MySQL without `-Pdocker`. Feels to me like `VertxMySqlContainer` is not doing its job quite right.

gavinking:
Alright so if I take `VertxMySqlContainer` out of the picture it mostly works; there's just one error right at the start of the test suite.

So apparently `VertxMySqlContainer` is busted.

DavideD:
I can test it now... give me a second.
Seems weird though because CI uses testcontiners and docker

gavinking:
> Seems weird though because CI uses testcontiners and docker

Right, on the command line there's no problem.

DavideD:
I think there's something wrong with the build.

Does this work for you?:
```
./gradlew -Pdocker testDbMySQL
```

DavideD:
Do you want me to look into it?

DavideD:
I have the same issue with testing with Intellij

gavinking:
What's strange is it *just started* happening.

DavideD:
It seems that for me it doesn't even work from the terminal but I've never used testcontainers before so I'm not sure if it's relevant

DavideD:
Is it only MySQL? Does it work if you use Db2?

gavinking:
> Does it work if you use Db2?

Nup.

gavinking:
Perhaps we just need to up the wait time.

gavinking:
Waaaaaaaiiiit. Now I see what's going on. It's not that anything's broken, it's just that the Vert.x blocked thread checker is now way more verbose and prints lots of errors at the start while testcontainers is trying to fetch the image!

The tests actually run fine.

PR #566, opened 2021-02-12
Maria DB support

### LINKED ISSUE #471: tests+dialect for MariaDB
Testcontainers has [a module for MariaDB](https://www.testcontainers.org/modules/databases/mariadb/).

MariaDB is for the most part just a flavor of MySQL, and we support it via the Vert.x driver for MySQL. But I think that's not quite good enough. There are some (very) minor differences in the Hibernate `Dialect`, and so we should treat it as a separate case.

So we should set up tests to run on Maria, and make sure we detect it and use the correct dialect by default, instead of making the user specify it explicitly.

gavinking:
Fixes #471.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerManyToOneAssociationTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerManyToOneAssociationTest.java
@@ -101,7 +101,7 @@ public class EagerManyToOneAssociationTest extends BaseReactiveTest {
 	@DiscriminatorValue("N")
 	//@Inheritance(strategy = InheritanceType.JOINED)
 	public static class Book {
-		public static final String TABLE = "Book";
+		public static final String TABLE = "Bookz";
 
 		@Id
 		private Integer id;
@@ -152,7 +152,7 @@ public class EagerManyToOneAssociationTest extends BaseReactiveTest {
 	@Table(name = Author.TABLE)
 	public static class Author {
 
-		public static final String TABLE = "Author";
+		public static final String TABLE = "Authorz";
 
 		@Id
 		private Integer id;
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/VertxMariaContainer.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/VertxMariaContainer.java
@@ -5,29 +5,28 @@
  */
 package org.hibernate.reactive.containers;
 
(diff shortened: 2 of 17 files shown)


gavinking:
Couple of very minor things to clean up before I merge.

gavinking:
Oh, and we need to set up the tests to run on the CI. @DavideD can you make that happen?

DavideD:
Sure, I will do it when we have merged this

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerManyToOneAssociationTest.java
```diff
@@ -101,7 +101,7 @@ public void manyToOneIsNull(TestContext context) {
 	@DiscriminatorValue("N")
 	//@Inheritance(strategy = InheritanceType.JOINED)
 	public static class Book {
-		public static final String TABLE = "Book";
+		public static final String TABLE = "Bookz";
```
DavideD, line 104: Just a curiosity, how come you also need to change the table name?

gavinking: Well MariaDB seems to be a lot pickier about that stuff for some reason I wasn't able to fully nail down. All I know is I wish it supported `cascade`.

gavinking: (In `drop table` I mean.)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/VertxMariaContainer.java
```diff
@@ -0,0 +1,107 @@
...
+import org.testcontainers.containers.MariaDBContainer;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Normally the MySQLContainer checks for readiness by attempting to establish a JDBC connection
```
DavideD, line 21: s/`MySQLContainer`/`MariaDBContainer`/

gavinking: Yup, already noticed.

gavinking:
> Sure, I will do it when we have merged this

OK, I'll merge it now then.

PR #568, opened 2021-02-12
[#471] Add MariaDB to workflows, podman.md and build.gradle

### LINKED ISSUE #471: tests+dialect for MariaDB

DavideD:
Resolves #471

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        db: [ 'MySQL', 'PostgreSQL', 'DB2' ]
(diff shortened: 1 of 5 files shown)


DavideD:
Waiting for CI to finish the build before merging this

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/568

PR #576, opened 2021-02-16
[#575] Set default port for MariaDB

### LINKED ISSUE #575: Default port not set for MariaDB
When the URL for MariaDB doesn't contain the port we don't set the default value.
I also noticed that the build get stuck if a test fails during the setup.

I have a fix almost ready.

DavideD:
Fixes #575 

This also throws an exception now if we don't have a default port for the scheme in the URI.

@gavinking I've noticed we still have the class [UriPoolConfiguration](https://github.com/hibernate/hibernate-reactive/blob/master/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UriPoolConfiguration.java) around.

I don't think we are using it now and if I remember correctly we plan to use it when they added some changes in Vert.x.
I'm wondering if we should remove it for now or at least add a Javadoc to it.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
@@ -141,9 +141,14 @@ public abstract class BaseReactiveTest {
 
 		final Async async = context.async();
 		Handler<AsyncResult<org.hibernate.SessionFactory>> result = r -> {
-			sessionFactory = r.result();
-			poolProvider = registry.getService( ReactiveConnectionPool.class );
-			async.complete();
+			if ( r.failed() ) {
+				context.fail( r.cause() );
+			}
+			else {
+				sessionFactory = r.result();
+				poolProvider = registry.getService( ReactiveConnectionPool.class );
+				async.complete();
+			}
 		};
 
 		vertxContextRule.vertx().executeBlocking( sfPromise, result );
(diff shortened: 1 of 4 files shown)


gavinking:
> I'm wondering if we should remove it for now or at least add a Javadoc to it.

well, no, we don't want to remove it, it's part of the test for the `hibernate.vertx.pool.configuration_class` setting.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -141,9 +141,14 @@ public void before(TestContext context) {
 
 		final Async async = context.async();
 		Handler<AsyncResult<org.hibernate.SessionFactory>> result = r -> {
```
gavinking, line 143: Must say I'm not at all a fan of the assigning-an–anonymous-function-to-a-local-variable idiom. I would inline this local. But that's orthogonal to this patch.

gavinking (approved):
LGTM

DavideD:
Sorry, my bad

> > I'm wondering if we should remove it for now or at least add a Javadoc to it.
> 
> well, no, we don't want to remove it, it's part of the test for the `hibernate.vertx.pool.configuration_class` setting.

RIght, I didn't realize it was in the test folder.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/576

PR #577, opened 2021-02-16
minor cleanup

gavinking:


gavinking:
Moar local variable hate.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -134,24 +130,20 @@ public void before(TestContext context) {
 		// when running in a Vert.x event loop thread.
 		// Vert.x can deal with blocking code using `executeBlocking` or more
 		// more in general with Work threads.
```
DavideD, line 132: well... since you are fixing this, could you also update the comment, please? `more` is repeated twice.
Otherwise looks good to me.

gavinking: I don't actually know what that sentence means.

DavideD: It's a reminder that we need to call `executeBlocking` because the Schema Generation is blocking  and  would  normally causes exception when executed in a Vert.x event loop thread.
To deal with it we need to execute it in a `executeBlocking`.
I guess we can remove the part about the Work threads.

DavideD:
Thanks!

PR #579, opened 2021-02-16
support for Map<> `@ElementCollection`

### LINKED ISSUE #532: add support for Map<> type ElementCollection
Support for Map<> element collections as a minimum, requires work in ```ReactiveBasicElementCollectionPersister```'s ```reactiveUpdatRows()``` to rewrite ```doUpdateRows()``` logic using a ```ReactiveConnection``` instead of JDBC.

See work in #44

blafond:
Fixes #532

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
@@ -30,6 +30,7 @@ import org.jboss.logging.Logger;
 
 import static org.hibernate.reactive.util.impl.CompletionStages.total;
 import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
+import static org.hibernate.reactive.util.impl.CompletionStages.zeroFuture;
 
 /**
  *
@@ -104,12 +105,24 @@ public class ReactiveBasicCollectionPersister extends BasicCollectionPersister i
 		Iterator<?> entries = collection.entries(this );
 		return total(
 				entries,
-				(entry, index) -> reactiveConnection.update(
-						getSQLInsertRowString(),
-						insertRowsParamValues( entry, index+1, collection, id, session )
-				)
-				//TODO: compose() a reactive version of collection.afterRowInsert()
+				(entry, index) -> {
+					if ( collection.entryExists( entry, index ) ) {
+						return reactiveConnection.update(
+								getSQLInsertRowString(),
+								insertRowsParamValues( entry, index, collection, id, session )
+						).thenApply( currentTotal -> afterRowInsert( currentTotal, collection, entry, index ) );
+					}
+					else {
+						return zeroFuture();
+					}
+				}
 		).thenAccept( total -> LOG.debugf( "Done inserting rows: %s inserted", total ) );
+
+	}
+
+	Integer afterRowInsert(Integer currentTot, PersistentCollection collection, Object entry, int i) {
+		collection.afterRowInsert( this, entry, i );
+		return currentTot;
 	}
 
 	/**
@@ -144,11 +157,10 @@ public class ReactiveBasicCollectionPersister extends BasicCollectionPersister i
 			return voidFuture();
 		}
 
-		ReactiveConnection reactiveConnection = getReactiveConnection( session );
 		Iterator<?> deletes = collection.getDeletes(this, !deleteByIndex() );
 		return total(
 				deletes,
-				(entry, index) -> reactiveConnection.update(
+				(entry, index) -> getReactiveConnection( session ).update(
 						getSQLDeleteRowString(),
 						deleteRowsParamValues( entry, index+1, id, session )
 				)
@@ -169,13 +181,13 @@ public class ReactiveBasicCollectionPersister extends BasicCollectionPersister i
 			SharedSessionContractImplementor session) throws HibernateException {
 
 		if ( !isInverse && collection.isRowUpdatePossible() ) {
-			// TODO:  convert AbstractCollectionPersister.doUpdateRows() to reactive logic
-			// update all the modified entries
 			// NOTE this method call uses a JDBC connection and will fail for Map ElementCollection type
 			// Generally bags and sets are the only collections that cannot be mapped to a single row in the database
 			// So Maps, for instance will return isRowUpdatePossible() == TRUE
-
-			throw new UnsupportedOperationException();
+			// A map of type Map<String, String> ends up calling this method, however
+			// for delete/insert of rows is performed via the reactiveDeleteRows() and
+			// reactiveInsertRows()... so returning a voidFuture()
+			return voidFuture();
 		}
 		return voidFuture();
 	}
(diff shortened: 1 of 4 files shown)


gavinking:
Awesome, thanks!!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -222,11 +228,16 @@ protected String getSQLUpdateRowString() {
 		return CompletionStages.total(
 				collection.entries( this ),
 				entry -> {
-					CompletionStage<Integer> st = reactiveConnection.update(
-							getSQLInsertRowString(),
-							insertRowsParamValues( entry, index, collection, id, session ) );
-					collection.afterRowInsert( this, entry, index.getAndIncrement() );
-					return st;
+					if ( collection.needsInserting( entry, index.get(), elementType ) ) {
```
DavideD, line 231: I think you need something like this here:
```
int i = index.getAndIncrement();
if ( collection.needsInserting( entry, i, elementType ) ) {
...
}
```
Or you only increment the index when the collection needs inserting.
This behaviour would be different than what happens in ORM: https://github.com/hibernate/hibernate-orm/blob/f451ed46d9d2856f9dab6a4ebb985a4d3ea8fee5/hibernate-core/src/main/java/org/hibernate/persister/collection/AbstractCollectionPersister.java#L1614

DavideD: I suspect there are no failures now because we don't have a test using `@Embeddable`

DavideD (changes requested):
This is great! Looks good overall and I'm a bit surprised the patch is this simple.

Still, there are a couple of things not quite right.

We still need at least a test that uses an `@Embeddable` with some attributes before we can merge it.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -155,11 +155,14 @@ protected String getSQLUpdateRowString() {
...
 
-		CompletionStage<Integer> loop = CompletionStages.zeroFuture();
 		ReactiveConnection reactiveConnection = getReactiveConnection( session );
 
 		boolean deleteByIndex = !isOneToMany() && hasIndex && !indexContainsFormula;
 
+		if( !collection.getDeletes( this, !deleteByIndex ).hasNext()) {
+			return CompletionStages.voidFuture();
+		}
+
```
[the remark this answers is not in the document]
blafond: Yes it was an attempt at opt...  I'll remove the check.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
blafond: ... right. In ORM version, the index is incremented whether or not entry `needsInserting(entry...)` so I'm cleaning that up. Note that `i` is 0 based, so the incrementing can't happen before the check.. putting it outside of the check.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForBasicTypeMapTest.java
```diff
@@ -0,0 +1,602 @@
...
+ * {@code
+ *     class Person {
+ *         @ElementCollection
+ *         Map<String, String> phones;
+ *     }
+ * }
+ * </p>,
+ *
+ */
+public class EagerElementCollectionForBasicTypeMapTest  extends BaseReactiveTest {
```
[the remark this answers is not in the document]
blafond: yep. fixed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForBasicTypeMapTest.java
```diff
@@ -0,0 +1,602 @@
...
+//		String[] sortedExpected = Arrays.stream( expectedPhones).sorted()
+//				.sorted( String.CASE_INSENSITIVE_ORDER )
+//				.collect( Collectors.toList() )
+//				.toArray( new String[expectedPhones.length] );
+//		List<String> sortedActual = person.getPhones().values().stream()
+//				.sorted( String.CASE_INSENSITIVE_ORDER )
+//				.collect( Collectors.toList() );
+//		Assertions.assertThat( sortedActual )
+//				.containsExactly( sortedExpected );
+//	}
```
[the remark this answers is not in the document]
blafond: removed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
DavideD:
```
int i = index.getAndIncrement();
```
This method will first return the value **and then** increment it.
Meaning the `i` will start from 0.
If this is what you mean, because index actually start from 1 :-)

DavideD: You won't use index anymore after this, so it should be correct

blafond: dah.. yep.. thx

blafond: Which @Embeddable scenario are you referring to?  

this one:
```
		@ElementCollection(fetch = FetchType.EAGER)
		private Map<String, Phone> phones = new HashMap<>();
```
or the @Embeddable with it's own @ElementCollection?

DavideD: Yes, I mean this case:
```
@ElementCollection(fetch = FetchType.EAGER)
private Map<String, Phone> phones = new HashMap<>();
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -185,8 +188,11 @@ protected String getSQLUpdateRowString() {
 			// NOTE this method call uses a JDBC connection and will fail for Map ElementCollection type
 			// Generally bags and sets are the only collections that cannot be mapped to a single row in the database
 			// So Maps, for instance will return isRowUpdatePossible() == TRUE
-
-			throw new UnsupportedOperationException();
+			// A map of type Map<String, String> ends up calling this method, however
+			// for delete/insert of rows is performed via the reactiveDeleteRows() and
+			// reactiveInsertRows()... so returning a voidFuture()
+			// TODO: create issue to test entity (embedded?) Map<?, ?> object types
```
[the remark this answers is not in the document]
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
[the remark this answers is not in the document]
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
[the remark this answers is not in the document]
blafond: done

DavideD (approved):
Thanks @blafond.

I think we can merge this. 
There are a couple of things that I think we need to change but I will have a look at them while working on the `ReactiveOneToManyPersister` because in ORM  these two collection persisters share some of the code.

PR #580, opened 2021-02-18
tests for list ordering

### LINKED ISSUE #550: Support OrderColumn annotation
It might already work but I don't think we've tested it.

See example 196 in the Hibernate ORM documentation: https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#collections-value

```
@ElementCollection
@OrderColumn(name = "order_id")
private List<String> phones = new ArrayList<>();

person.getPhones().remove( 0 );
```

Maybe it's a good first issue

gavinking:
fixes #550

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForBasicTypeSetTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForBasicTypeSetTest.java
@@ -43,7 +43,7 @@ import org.assertj.core.api.Assertions;
  * </p>,
  *
  * @see EagerElementCollectionForBasicTypeListTest
- * @see EagerElementCollectionForEmbeddableEntityTypeListTest
+ * @see EagerElementCollectionForEmbeddableTypeListTest
  */
 public class EagerElementCollectionForBasicTypeSetTest extends BaseReactiveTest {
 
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForEmbeddableTypeListTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerElementCollectionForEmbeddableTypeListTest.java
@@ -5,16 +5,16 @@
  */
 package org.hibernate.reactive;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.Collectors;
 import javax.persistence.ElementCollection;
+import javax.persistence.Embeddable;
 import javax.persistence.Entity;
 import javax.persistence.FetchType;
 import javax.persistence.Id;
-import javax.persistence.Table;
 
 import org.hibernate.cfg.Configuration;
 import org.hibernate.reactive.mutiny.Mutiny;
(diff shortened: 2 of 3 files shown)


gavinking:
@blafond @DavideD there's something I noticed when I was working on this that I apologize for not noticing before: in these tests we have lots of duplicated tests, one for `Stage`, and one for `Mutiny`. 

We *definitely* don't need to do this in future. Indeed, it's bad that we have done this (though I left it alone for now).

Look, I know there's a lot of irrational/magical thinking around testing in our industry but objectively:

- the more code you have, the harder it is to change your program
- therefore, code duplication is bad
- tests are code
- therefore, duplicated tests are bad QED

It is simply not the case that "more tests" => "better tested code". Nononononono. To have well-tested code you need to *think carefully about what sort of things might break* and target your tests toward those things, always actually *aiming to break your code*. 

Sure, I can spend hours writing thousands of tests for minor variations of stuff that is very similar to stuff I already have tests for and then walk around telling everyone "ooooh look how well-tested my code is", but I would be a liar, because I wasn't actually *trying to break anything*. All I've done is make my test suite slow and redundant and difficult to maintain. All I've done is created problems for the future.

I don't need a test suite where 300 tests fail when I break one thing. Actually, I hate it when that happens.

DavideD:
I think I need some time to answer to this.

gavinking:
Sure.

All I'm saying is it's OK to use our insight into _how our code is written_ when deciding which tests to write. We know for sure that `StageSession` and `MutinySession` are very trivial wrappers, and so, once we have basic tests for their basic operations, we don't need to duplicate every other test for both of them.

DavideD:
> All I'm saying is it's OK to use our insight into _how our code is written_ when deciding which tests to write. We know for sure that `StageSession` and `MutinySession` are very trivial wrappers, and so, once we have basic tests for their basic operations, we don't need to duplicate every other test for both of them.

That's not true. If this were your only comment,  I wouldn't have any problem discussing it.
But you are saying that I write tests for the sake of adding code to our code base without thinking about it.

I might be open to discuss the fact that testing both API might not be necessary and we maybe need to rethink the way we write the tests (I know that build time will become a problem, eventually).

But the way you voice your concerns doesn't put me in the best mood to participate to this conversation.

gavinking:
> But you are saying that I write tests for the sake of adding code to our code base without thinking about it.

No, I was espousing a general principle. "You" in that comment never refers to you personally. You can read "you" as "one", if you prefer.

gavinking:
i.e.

The middle paragraph can be reworded:

> It is simply not the case that "more tests" => "better tested code". Nononononono. To have well-tested code I need to think carefully about what sort of things might break and target my tests toward those things, always actually aiming to break my code.

blafond:
I sorta agreeing with you both .. bare with me.  

My first real reactive issue was to implement support for the complex world of the @ElementCollection. The current suite of tests for each element type cover a wide range of scenarios for editing. The Action's from the Queue are executed for each edit during flush() and the logic for updating the DB via the persisters.  The edits result in recreate, remove, insert, delete and update operations and sometimes a mix of more than one operation.  Re-working the ORM version of these edits hasn't been straight forward... at least for me.  Seems that having tests for each scenario for each type at this point makes sense.

My understanding of the Stage -vs- Mutiny is about testing that the corresponding functions/methods that execute the user functions. I haven't dug into the Mutiny framework yet... but there has yet to be a case where a direct swapping out of methods didn't work.  So I'm in favor of just keeping one set of tests in one element collection type. 

This would reduce the # tests for each type from 30 -- > 15

gavinking:
> My first real reactive issue was to implement support for the complex world of the @ElementCollection. The current suite of tests for each element type cover a wide range of scenarios for editing.

I mean, for sure there are going to need to be quite a few tests for this. I mean, just to begin with, `ReactiveCollectionPersister` has five different reactive operations, and we need to be sure these work correctly for `List` vs `Set` vs `Map` vs array, and with variations like `@OrderColumn`.

> Re-working the ORM version of these edits hasn't been straight forward... at least for me. 

If it was straightforward we would have already done it. It was a missing feature because we knew it wasn't that easy.

> Seems that having tests for each scenario for each type at this point makes sense.

Absolutely.

> This would reduce the # tests for each type from 30 -- > 15

Right, that's what I'm talking about.

DavideD:
It seems to me that the main issue is about testing the Session.Mutiny API.

I'm not completely on board with this but I think having a place where discuss it might be more constructive.

DavideD:
> It seems to me that the main issue is about testing the Session.Mutiny API.
> 
> I'm not completely on board with this but I think having a place where discuss it might be more constructive.

Sorry, I meant to say that I'm going to create an issue about it

DavideD:
> No, I was espousing a general principle. "You" in that comment never refers to you personally. You can read "you" as "one", if you prefer.

That might be a general principle but you have written it in a comment addressed to me, in regard to an issue I helped to solve on a project I'm contributing with others.
You didn't write it on your personal twitter account or on your blog.

The context you have decided to share this story and the words you have chosen make it personal.

Even if, maybe, you didn't mean to.

PR #582, opened 2021-02-18
minor cleanups to ReactiveBasicCollectionPersister

gavinking:
1. fix some generics-related warnings
2. `afterRowInsert()` would need to be called in `compose()` but it's not implemented yet anyway
3. suck the ugly `AtomicInteger` into the util method
4. use static imports

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
@@ -6,8 +6,8 @@
 package org.hibernate.reactive.persister.collection.impl;
 
 import java.io.Serializable;
+import java.util.Iterator;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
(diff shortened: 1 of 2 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
```diff
@@ -101,29 +100,23 @@ protected String getSQLUpdateRowString() {
...
-				.thenAccept( total -> LOG.debugf( "Done inserting rows: %s inserted", total ) );
+		Iterator<?> entries = collection.entries(this );
+		return total(
+				entries,
+				(entry, index) -> reactiveConnection.update(
+						getSQLInsertRowString(),
+						insertRowsParamValues( entry, index+1, collection, id, session )
+				)
+				//TODO: compose() a reactive version of collection.afterRowInsert()
+		).thenAccept( total -> LOG.debugf( "Done inserting rows: %s inserted", total ) );
```
gavinking, line 112: @blafond careful with stuff like this. Calling `afterRowInsert()` after calling something which returns a `CompletionStage` almost certainly doesn't do what you intend it to do.

(Apologies for not catching this when reviewing the PR.)

blafond: noted. thx

DavideD:
Nice clean up. LGTM

PR #587, opened 2021-02-20
lazy initialization for @ElementCollection

### LINKED ISSUE #551: Test lazy fetching of @ElementCollection
Follows: #44 
So far we only have tests for the eager case but we need to check that it also work with `Session#fetch`.

gavinking:
also added some Javadoc

fixes #551


gavinking:
I've also implemented support for subselect fetching!

Issue #589, opened 2021-02-20
one-sided @OneToMany


gavinking:
A `@OneToMany` association without `mappedBy` pointing to a matching `@ManyToOne` is actually a bad feature that you should essentially never use, however, it's legal, and it's only a very little bit of work to implement it now, so we should do it.

Issue #581, opened 2021-02-18
@OrderColumn


gavinking:
`ReactiveBasicCollectionPersister.reactiveUpdateRows()` is currently unimplemented, which means an `@ElementCollection` with an `@OrderColumn` blows up.

gavinking:
@Blafond this looks straightforward to implement, wanna take a look?

gavinking:
After #590, I believe this is most likely working. But we need to write tests for it.

gavinking:
> But we need to write tests for it.

I wrote a couple, but we need more.

And we should figure out what is the story with `@OrderColumn` on a unidirectional `@OneToMany`. I suspect the bug I found also exists in core, but we should check that.

DavideD:
> (skip a test on DB2; for some reason this test exposes a bug in the driver)

@gavinking Don't we need an issue to keep track of this?

gavinking:
I'm going to report it to Vert.x, but it's pretty intermittent. I've no idea how to make an isolated test for it.

DavideD:
Mmmh... that's nasty.
I'm prone to say that we should create an issue for it anyway to keep track of it and eventually update the tests.

But it's your call

gavinking:
Nah, it's not our problem, and it's clear from the error message that it's something they know about already:

```
ERROR errors [vert.x-eventloop-thread-0] failed to execute statement [select orderedemb0_.id as id1_0_0_, orderedemb0_.title as title2_0_0_ from EAuthor orderedemb0_ where orderedemb0_.id=?]
ERROR errors [vert.x-eventloop-thread-0] could not load an entity: [org.hibernate.reactive.OrderedEmbeddableCollectionTest$Author#1]
java.util.concurrent.CompletionException: java.lang.IllegalStateException: Needed to have 6 in buffer but only had 0. In JDBC we would normally block here but need to find a non-blocking solution
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346) ~[?:?]
	at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:632) ~[?:?]
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088) ~[?:?]
[...]

PR #590, opened 2021-02-20
initial support for unidirectional @OneToMany and @OrderColumn

gavinking:
see #589


gavinking:
So this implementation seems to work for `@OneToMany`s with no `@OrderColumn`, which we currently state is unsupported (see #581).

I have not tested it for `Map`s, but I suspect it mostly works for the case of `Map`.

gavinking:
I've now done a bunch of work on this branch:

- I've ported over all the missing code from the collection persisters. TBH I'm not at all clear on what all of this code does, and so I don't really know how to test it, but at least we have a very sensible-looking implementation to start writing tests for.
- I've done some work on the design of the hierarchy, to make sure we don't have code duplication. It's all quite clean.
- I've wired in the `Expectation` stuff, which is an important double-check on our code, and is also important for users to detect when someone else has modified the database underneath them.
- By side-effect, I've made batching work, I believe.
- I've removed some logging which was unnecessary and resulted in longer chains of `CompletionStage`s.
- I've written a test for `@OneToMany @OrderColumn` which basically works except for a bug which I highly suspect also exists in core. We'll need to write a bunch more tests for `@OrderColumn`, which should be working now.

gavinking:
> * I've written a test for `@OneToMany @OrderColumn` which basically works except for a bug which I highly suspect also exists in core. We'll need to write a bunch more tests for `@OrderColumn`, which should be working now.

I added a couple more tests for this, and, as I suspected, `@ OrderColumn ` now works fine in conjunction with `@ManyToMany` and `@ElementCollection`.

gavinking:
Closing in favor of #591.

PR #591, opened 2021-02-21
unidirectional @OneToMany and @OrderColumn

### LINKED ISSUE #589: one-sided @OneToMany

gavinking:
Fixes #589 and #581.

Supersedes #590.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,15 +194,10 @@ At present a few limitations remain.
 
 ### Mapping annotations
 
-The following mapping annotations are not yet supported:
+The following Hibernate mapping annotations are not yet supported:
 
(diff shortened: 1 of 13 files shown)


gavinking:
I'm going to merge this one soon because I don't want it to turn into a whole separate dev branch. You guys need to be basing your wok on this stuff.

I'm not totally happy with the tests but at least there's quite a few of them, and perhaps I'll open an issue to improve them.

PR #601, opened 2021-02-22
Vert.x 4 updates

DavideD:


gavinking:
Does anyone think we should de a beta release of HR for vert.x 4?

DavideD:
I suppose we could. 
I would require a change in the artifact or groupid though.

Is there a plan about using Vert.x 4 in Quarkus?

gavinking:
> Is there a plan about using Vert.x 4 in Quarkus?

Well, yes, but the whole conversation has all got a bit confusing due to the debate about `jakarta.`.

gavinking:
> It will require a change in the artifact or group id.

Hrm. I suppose you're right. Well, is there any reason to not create a `hibernate-reactive-core-vertx4` artifact id?

DavideD:
None that I can think of. It was merely an observation.

Issue #411, opened 2020-10-19
improvement to hi/lo generator


gavinking:
In Hibernate core fetching of a new hi value blocks concurrent access to the hi/lo generator.

@cescoffier suggests the following idiom to do the same thing in Vert.x

<blockquote>

I have a specific component used by the rest of the connector. 
This component exposes an asynchronous API, let's call this method `getId` and you pass a callback invoked with the retrieve id. 
The component handles a queue of requests. In this queue, you store the current Vert.x context and the callback. When you needs an id, you do something like:

```java
public void getId(Consumer<Id> callback) {
   if (idIsAvailable()) {
      callback.call(available);
   } else {
      Context ctxt = Vertx.getOrCreateContext();
      queue.add(new Request(context, callback);
      retrieveIfNeeded();
   }
}
```

`retrieveIfNeeded` verify is there is a retrieval in progress:

```java
private void retrieveIfNeeded() {
   if (!inProgess ) {
        retrieve();
   }
}

private void retrieve() {
    inProgress = true;
    retrieveTheIds(ids -> {
         invokeCallbackAndClearTheQueue(ids);
         inProgress = false;
   );
}
```

When invoking the callbacks, we make sure it's invoked on the right context:

```java
ctxt.runOnContext(() -> callback.call(id));
```

</blockquote>

cescoffier:
BTW, do you need some kind of periodic task or enforce a delay between 2 retrievals?

gavinking:
> BTW, do you need some kind of periodic task or enforce a delay between 2 retrievals?

No, this is something that happens when we run out of lo values.

PR #605, opened 2021-02-24
improving hi/lo id generation

gavinking:
Don't issue concurrent requests for new hi values.

see #411

@Sanne @cescoffier  does this look right to you guys?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/TableReactiveIdentifierGenerator.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/TableReactiveIdentifierGenerator.java
@@ -18,7 +18,6 @@ import org.hibernate.internal.util.StringHelper;
 import org.hibernate.jdbc.TooManyRowsAffectedException;
 import org.hibernate.reactive.pool.impl.Parameters;
 import org.hibernate.reactive.provider.Settings;
-import org.hibernate.reactive.id.ReactiveIdentifierGenerator;
 import org.hibernate.reactive.pool.ReactiveConnection;
 import org.hibernate.reactive.session.ReactiveConnectionSupplier;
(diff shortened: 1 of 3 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/TableReactiveIdentifierGenerator.java
```diff
@@ -89,7 +94,42 @@ private synchronized long next(long hi) {
...
 		}
+		else {
+			synchronized (this) {
+				CompletableFuture<Long> result = new CompletableFuture<>();
+				if (queue == null) {
+					// make a queue for any concurrent streams
+					queue = new ArrayList<>();
+					// go off and fetch the next hi value from db
+					nextHiValue(session).thenAccept( id -> {
+//						Vertx.currentContext().runOnContext(v -> {
```
gavinking, line 105: I'm not sure if this is needed or not.

cescoffier (approved):
The pattern looks good.

gavinking:
@Sanne if you don't respond I'm going to just merge it ;-)

DavideD:
He is on PTO, today.
I suppose we can merge it. We already have a test somewhere for this isn't it?

DavideD:
> We already have a test somewhere for this isn't it?

Well, I suppose lot of tests will fail if this doesn't work

gavinking:
> He is on PTO, today.
> I suppose we can merge it. We already have a test somewhere for this isn't it?

Yeah, and I know something useful is being tested there, because I had to struggle more than I expected to get the tests to pass when making this change. (I had not originally considered that there could be more queued id requests than the block size.)

What we *don't* have is any tests for "concurrent" streams. But I'm totally not going to try adding them right now.

DavideD:
Sounds good to me

PR #607, opened 2021-02-24
[#313] Support for Generated properties

### LINKED ISSUE #313: @Generated properties
Fields mapped with `@org.hibernate.annotations.Generated` don't work in Hibernate Reactive because we need reactified versions of `processInsertGeneratedProperties()`, `processUpdateGeneratedProperties()`, and `processInsertGeneratedProperties()`.

I don't believe this is a super-high priority, but it's worth keeping track of.

DavideD:
Fixes #313

- Support for `@Generated` and `@GeneratorType`
- Does not support `@Version @Source(SourceType.DB) Date version` (I will create a separate issue)
- I've found a bug when using `InheritanceType#JOINED` and Identity generated id. So I've disabled that test for now
- We can remove some methods that I've copied from ORM when 5.4.29 has been released, but I would merge this anyway and fix it later


gavinking (approved):
LGTM.

The only thing I want to change is to make the chain a bit shorter in the case where there are no generated properties by not chaining useless void stages. But I can make that change after you merge b/c I know how I want to do it.

DavideD:
Sounds good.
Should I add the `@Generated ` annotation to the list of useful annotations in the doc?

gavinking:
> Should I add the `@Generated ` annotation to the list of useful annotations in the doc?

OK. I do think it's useful to let people know about all these nice features of Hibernate that they can use that aren't in JPA.

gavinking:
So we need a release of core before we can merge this, right?

DavideD:
No, this PR will work without the core release but I've copied some methods: https://github.com/hibernate/hibernate-reactive/pull/607/files#diff-2f66a2de01610794b46c16fbf21dc4b2ec7e5dbce83e7cd7c7ef8e47b9cafdb3R200

gavinking:
Oh, right, OK. Then merge it.

DavideD:
But if you prefer to wait to merge it, I will update the PR so that's ready to be merged when the next release of ORM is out

gavinking:
No, I hate having unmerged PRs hanging around.

DavideD:
Mee too.

Merged https://github.com/hibernate/hibernate-reactive/pull/607

Thanks

gavinking:
Great!! This feature has been bugging me for a long time.

PR #614, opened 2021-02-25
[#609] Fix type support for GenerationStrategy.IDENTITY

### LINKED ISSUE #609: Id of type Integer causes exception with generated identity id
It happens because the value returned by the insert is a `Long` and we are using it on an `Integer` field 

```
	@Test
	public void test(TestContext context) {
		final TestHibernateDBO rose = new TestHibernateDBO();
		rose.name = "Rosa";

		Mutiny.Session mutinySession = openMutinySession();
		test( context, mutinySession.persist( rose ).call( mutinySession::flush )
				.chain( () -> getMutinySessionFactory().withSession( session -> session.find( TestHibernateDBO.class, rose.id ).invoke( result -> {
					context.assertEquals( rose, result );
				} ) ) ) );
	}

	@Entity
	@Table(name = "test_hibernate")
	public static class TestHibernateDBO {

		@Id
		@GeneratedValue(strategy = GenerationType.IDENTITY)
		@Column(name = "test_hibernate_id")
		public Integer id;

		@Column(name = "name")
		public String name;

		public TestHibernateDBO() {
		}

		@Override
		public boolean equals(Object o) {
			if ( this == o ) {
				return true;
			}
			if ( o == null || getClass() != o.getClass() ) {
				return false;
			}
			TestHibernateDBO that = (TestHibernateDBO) o;
			return Objects.equals( name, that.name );
		}

		@Override
		public int hashCode() {
			return Objects.hash( name );
		}
	}
```

Throws:
```
[Exception 0] javax.persistence.PersistenceException: org.hibernate.PropertyAccessException: Could not set field value [1] value by reflection : [class org.hibernate.reactive.ORMReactivePersistenceTest$TestHibernateDBO.id] setter of org.hibernate.reactive.ORMReactivePersistenceTest$TestHibernateDBO.id
```

DavideD:
Fixes #609 

This will allow Identity type identifier of type Short, Long and Integer on all the db
It seems that the other types will create error when the table is created and won't work anyway.

I've also noticed that we weren't testing the property `hibernate.jdbc.use_get_generated_keys` but we run different queries when
it is enabled. I followed the approach used for Postgres and enabled Db2 as well.

@gavinking Do we really need to run different queries? It seems that if we follow the path of the enabled property we can get the id with just one query: https://github.com/hibernate/hibernate-reactive/compare/master...DavideD:609-integer-id?expand=1#diff-2f66a2de01610794b46c16fbf21dc4b2ec7e5dbce83e7cd7c7ef8e47b9cafdb3R395

Or was it done like that because it didn't work for Db2?


DavideD:
By the way, initially I was planning to go for a more involved solution so that I could reuse the Hibernat ORM method `IdentifierGeneratorHelper.getGeneratedIdentity`. It became a bit more complex than I wanted for this issue but we might want to do it at some point.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -390,22 +391,55 @@ void preInsertInMemoryValueGeneration(Object[] fields, Object object,
...
+		Class clazz = delegate().getIdentifierType().getReturnedClass();
+		if ( clazz == Long.class ) {
+			return generatedId;
+		}
+		if ( clazz == Integer.class ) {
+			return generatedId.intValue();
+		}
+		if ( clazz == Short.class ) {
+			return generatedId.shortValue();
+		}
```
gavinking, line 441: @DavideD this sorta duplicates the logic in `IdentifierGeneration.assignIdIfNecessary()`. I think we should find a way to unify both into one method.

gavinking: At the _very least_ it should be moved to a `static` method of `IdentifierGeneration`.

gavinking (changes requested):
Looks great except for that one thing.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
DavideD: Makes sense.

I actually noticed that, but I when I looked at `IdentifierGeneration` I didn't feel like this method would fit there.
I will give  a second look

DavideD: No, you are right. I will update the PR

gavinking (approved):
Great, thanks!

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/614
Thanks

PR #625, opened 2021-02-28
add example with all-handwritten SQL and stateless session

gavinking:


DavideD:
This looks great.
The only thing I would change is to not call the folder `example` and `example2` but move them under a folder `examples`: 
```
/examples
    -- /stateful
    -- /stateless
```
But I guess it's not critical

gavinking:
Yep, you're right, good point. But we could make that change after merging this (assuming we decide to merge it).

gavinking:
@DavideD What I wanted to ask you is if you can easily set this up to be tested as part of the CI build, like the other example.

DavideD:
Yes, it shouldn't be a problem to add it to the workflow.
Any reason why we shouldn't merge it?

DavideD:
Maybe we could add a quick note about the difference between the two sessions? Or at least a link to somewhere where it is described?

I know that it is explained in the javadoc but I suspect people will be confused about which approach to use when they see the two examples

gavinking:
Right, exactly, that's my reservation about this.

I think we at least need a README in each example dir.

example2/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+// The following rules define a task to run
+// the different API available.
+//
+// They require the selected db ready
+// to accept connections.
+//
+// Examples:
+// gradle runExampleMySQLMain runExamplePostgreSQLMutinyMain
+def mainClasses = ['Main', 'MutinyMain']
+def dbs = ['PostgreSQL']
```
DavideD, line 47: No MySQL?

DavideD: Ah, right

DavideD:
I've added the new example to the workflow with this commit: https://github.com/hibernate/hibernate-reactive/pull/632/commits/b1bc6fbe5a2a406cba372f870e8c658fd44d7917

You can Cherry pick the commit and add it to your branch or we can merge my PR after this one (it might need a rebase if you have done some changes).

Your choice

gavinking:
Hey, if you wanna just take it over, be my guest! :-)

DavideD:
Well... no not really, but I think we can merge it and sort out the details about the README and the folders later.
I was under the impression you wanted to change something, if that's not case I'm going to merge it

gavinking:
OK, let's do it. And then add the READMEs and change the dir layout and add your stuff.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/625
Thanks

PR #630, opened 2021-03-01
alternative impl of StatelessSession batching

gavinking:
This approach maintains "statelessness" of the SS by using two of them!


gavinking:
@DavideD WDYT of this? Any better idea?

DavideD:
Yeah sorry I haven't commented before.
The thing is that I find a bit hard to follow what's going on with these two constructors.

But, otherwise, I like how the class looks like and I think we can merge this

gavinking:
Well the issue is this: all the infrastructure in core calls back to the session to obtain a connection.

So when we're doing something where we want to enable batching, we need to switch out the connection for a different one.

There's three ways I could do that:

1. switch the connection pointed to by our instance variable, which is sort of nasty and results in bad behavior if the user does something they're not allowed to do, or
2. use two instances of `StatelessSession` with different connections, or
3. proxy the whole damn `StatelessSession`.

The third option seems like too much work. Option 2 seems better than 1.

Do you agree?

DavideD:
Ah, ok. I get it now.

I agree.

PR #631, opened 2021-03-01
Allow filtering of tests with testAll

DavideD:
Sometimes I want to run only selected tests on all databases without running the whole suite.

Example:

```
./gradlew testAll [-Pdocker] -PincludeTests=DefaultPortTest 
```

I know that a task of type `Test` supports the `--tests` parameter
but, for some reason, it doesn't work in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,15 @@ MySQL, MariaDB, or Db2, you must explicitly specify `-Pdb=mysql`,
 
     ./gradlew test -Pdb=db2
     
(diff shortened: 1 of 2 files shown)


DavideD:
I've also updated the readme to mention it

DavideD:
Merge https://github.com/hibernate/hibernate-reactive/pull/631
It shouldn't have any impact and we can revert it later if there are objections

PR #632, opened 2021-03-01
Add Stateless example to the workflow

DavideD:
Testing the build with the stateless examples


gavinking:
Thanks @DavideD, appreciated.

PR #633, opened 2021-03-02
rename selectLong() to selectIdentifier()

gavinking:
And add a comment about batching, to make its semantics clear.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/SequenceReactiveIdentifierGenerator.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/SequenceReactiveIdentifierGenerator.java
@@ -42,7 +42,7 @@ public class SequenceReactiveIdentifierGenerator
 
 	@Override
 	protected CompletionStage<Long> nextHiValue(ReactiveConnectionSupplier session) {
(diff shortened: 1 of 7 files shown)


DavideD:
Awesome!
Since you are changing this area, can we also change `updateReturning` to `insertSelectIdentity`,  or something similar?

gavinking:
> Since you are changing this area, can we also change `updateReturning` to `insertSelectIdentity`, or something similar?

I *guess*, but I feel like it's a bit more general than that. In this case this method is general-purpose, for *any* DML that returns a value.

`selectIdentifier()` OTOH now has rather special semantics.

DavideD:
I don't think it's that generic.

The `updateReturning` method has a condition that if the query doesn't return any value, it will call `getMySqlLastInsertedId`. Plus it returns a `Long`

DavideD:
We could also move the MySQL logic in the class generating the identity. It might make more sense

gavinking:
Ah, you have a good point. 

OK, fine, I will rename it.

gavinking:
> We could also move the MySQL logic in the class generating the identity. It might make more sense

Nah I'm hoping to keep that logic nicely contained in there because it will surely change at some stage when Vert.x figures out a better/more general way to do this.

DavideD (approved):

PR #636, opened 2021-03-02
Vert.x 4 updates

DavideD:
The updates we applied to `master`


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/636

PR #638, opened 2021-03-02
Fix workflow on GitHub

DavideD:
I've just realised I've forgotten to update the tasks in the two workflows for the latest ORM and Vert.x and so they were running the stateful example twice.

Apologies


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/638

PR #640, opened 2021-03-03
Modify parameters for CockroachDB

conorfarrell-coats:
CockroachDB uses a very similar dialect to Postgres, and requires the same modification of ?, ? -> $1, $2 for hibernate-reactive to work with it.

I was trying to follow the CockroachDB Hibernate tutorial using hibernate-reactive (as we use Vert.x 4) and came across this blocker for us. I don't believe there are any easy extension points that would allow us to change this behaviour at runtime, please correct me if I'm wrong!

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -54,6 +54,7 @@ dependencies {
     testImplementation "org.testcontainers:mysql:${testcontainersVersion}"
     testImplementation "org.testcontainers:mariadb:${testcontainersVersion}"
     testImplementation "org.testcontainers:db2:${testcontainersVersion}"
+    testImplementation "org.testcontainers:cockroachdb:${testcontainersVersion}"
 }
 
 // Print a summary of the results of the tests (number of failures, successes and skipped)
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
@@ -180,6 +180,8 @@ public class DefaultSqlClientPoolConfiguration implements SqlClientPoolConfigura
                 return 3306;
             case "db2":
                 return 50000;
+            case "cockroachdb":
+                return 26257;
             default:
                 throw new IllegalArgumentException( "Unknown default port for scheme: " + scheme );
         }
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/Parameters.java
@@ -5,6 +5,7 @@
  */
 package org.hibernate.reactive.pool.impl;
 
+import org.hibernate.dialect.CockroachDB192Dialect;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.dialect.PostgreSQL9Dialect;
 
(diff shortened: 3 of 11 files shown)


DavideD:
I see, so you are configuring the project to work with Postgres except for the dialect.
@gavinking WDYT?  I think we can merge it.

DavideD:
Thanks @conorfarrell-coats.

By any chance, did you already check if it actually works with CockrochDB?

DavideD:
I mean, by launching our test suite

conorfarrell-coats:
I ran the test suite locally (./gradlew test -Pdocker) and all passed, yes. I have been able to connect to CockroachDB using hibernate-reactive, albeit with some horrifying reflection-based workarounds due to Vert.x 4 removing the void close() method on various things. :disappointed: You can query for data, but not insert due to the parameters not being recognised.

DavideD:
What I meant was start CockroachDB and then run our test suite, including your change, using  `./gradlew test`. If you use the parameter for docker it will test with Postgres

gavinking:
Oh wow, it had not even occurred to me that it might work with Cockroach!

gavinking:
Question: is there some way to guess that the DB is Cockroach from the connection URI?

(I'm supposed to know the answer to this, and I *did* know, but now I forget.)

gavinking:
Oh and since test containers supports cockroach:

<https://www.testcontainers.org/modules/databases/cockroachdb/>

It should be added to the test suite.

DavideD:
It seems it relies on the Postgres connection URL: https://www.cockroachlabs.com/docs/stable/connection-parameters.html

DavideD:
> It seems it relies on the Postgres connection URL: https://www.cockroachlabs.com/docs/stable/connection-parameters.html

But I suppose we could add scheme for `cockroach`

gavinking:
> It seems it relies on the Postgres connection URL: https://www.cockroachlabs.com/docs/stable/connection-parameters.html

Yes that was what I had a hazy recollection of. That's annoying.

gavinking:
> But I suppose we could add scheme for `cockroach`

Hrm. But is that better than making them specify the `Dialect`. Not clear to me.

conorfarrell-coats:
I will give running the test suite with CockroachDB a go.

conorfarrell-coats:
I've added the ability to run CockroachDB tests in the same fashion as the other Dockerized databases (-Pdocker -Pdb=cockroachdb). A number of tests are failing, but I'm afraid I don't know enough about Hibernate's internals to be certain of the best way to fix them - any input welcome!

gavinking:
Well, *for now*, if they're caused by anything other than using a reserved word as a table or column name, I would say just disable them with a `Rule`.

DavideD:
> Hrm. But is that better than making them specify the `Dialect`. Not clear to me.

It's a bit better for them because they don't have to set an additional parameter, but, more important, it's better for us that we don't have to add conditions in the code like `if scheme is postgres but dialect is cochroach`. It's also consistent with the other databases.

DavideD:
> I've added the ability to run CockroachDB tests in the same fashion as the other Dockerized databases (-Pdocker -Pdb=cockroachdb). A number of tests are failing, but I'm afraid I don't know enough about Hibernate's internals to be certain of the best way to fix them - any input welcome!

Thanks a lot, this is great!
I will have a better look at it in a bit and get back to you.

gavinking:
> It's a bit better for them because they don't have to set an additional parameter, but, more important, it's better for us that we don't have to add conditions in the code like `if scheme is postgres but dialect is cochroach`. It's also consistent with the other databases.

OK, fine by me.

DavideD:
I've created an issue: #642

PR #643, opened 2021-03-03
initial support for @NaturalIds

gavinking:
I need to do some more thinking about this because `@SafeVarargs` doesn't work on interfaces so we get the stupid stupid useless fucking unchecked array creation warning. Dumbest thing in Java.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/NaturalIdTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/NaturalIdTest.java
@@ -1,0 +1,65 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import io.vertx.ext.unit.TestContext;
+import org.hibernate.annotations.NaturalId;
+import org.hibernate.cfg.Configuration;
+import org.junit.Test;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+
+import static org.hibernate.reactive.common.Identifier.value;
+
+public class NaturalIdTest extends BaseReactiveTest {
+
+    @Override
+    protected Configuration constructConfiguration() {
+        Configuration configuration = super.constructConfiguration();
+        configuration.addAnnotatedClass(Thing.class);
+        return configuration;
+    }
+
+    @Test
+    public void test(TestContext context) {
+        Thing thing1 = new Thing();
+        thing1.naturalKey = "abc123";
+        thing1.version = 1;
+        Thing thing2 = new Thing();
+        thing2.naturalKey = "abc123";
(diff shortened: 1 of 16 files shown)


gavinking:
@DavideD any idea why these CI runs are failing? Everything working fine locally.

DavideD:
Yes, it has to do with the latest ORM 5.5 SNAPSHOT: https://github.com/hibernate/hibernate-reactive/pull/643/checks?check_run_id=2024823063

DavideD:
You will see the error if you run this command (I think):
```
./gradlew build -Pdb=DB2 -Pdocker -PhibernateOrmVersion='[5.5,5.6)' -PskipOrmVersionParsing -PenableJBossSnapshotsRep
```

gavinking:
OK, so I guess this doesn't really affect us doing a release that targets 5.4, right?

DavideD:
I don't think so, but we will continue to see errors until it's compatible

DavideD:
By the way, the problem is due to this commit in ORM: https://github.com/hibernate/hibernate-orm/commit/ed3bbf15e476b064b6a55a80ac0c1570a4287bf0

DavideD:
I think i'm going to disable the build for 5.5 for now.
Unless somebody finds a solution that doesn't require a release of 5.4

gavinking:
Thanks @DavideD

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NaturalIdTest.java
```diff
@@ -0,0 +1,66 @@
...
+        thing2.naturalKey = "abc123";
+        thing2.version = 2;
+        test( context,
+                getSessionFactory()
+                        .withSession( session -> session.persist(thing1, thing2).thenCompose( v -> session.flush() ) )
+                        .thenCompose( v -> getSessionFactory().withSession(
+                                session -> session.find( Thing.class, composite(
+                                        id("naturalKey", "abc123"),
+                                        id("version", 1)
+                                ) )
```
gavinking, line 44: Alright, I've now got an API for this which isn't *terrible*, though it's also far from perfect. (It's better if you use the typesafe JPA metamodel.)

I propose including this in the release, and I've clearly marked it `@Incubating`.

gavinking: @DavideD? @Sanne?

DavideD: LGTM

DavideD: I guess an alternative would be a DSL:
```
session.find( Thing.class, id("naturalKey", "abc123")
                                 .and("version", 1) )
```

but I don't think I like this solution more

gavinking: I don't much like it either, even though it's perhaps a little more ergonomic.

gavinking:
I'm going to merge it. It's marked experimental, and I would rather have it there so people can experiment with it.

PR #644, opened 2021-03-04
Disable build for ORM 5.5 SNAPSHOTS

DavideD:
It causes too many errors at the moment because of this commit in ORM: https://github.com/hibernate/hibernate-orm/commit/ed3bbf15e476b064b6a55a80ac0c1570a4287bf0

I think the solution would be to add similar changes to 5.4 but that requires to wait for the another release.

It's easier to disable it for now


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/644

PR #645, opened 2021-03-04
Better way to disable 5.5 ORM Job

DavideD:
I've just realized that it is possible to disable the job on GitHub, so I've reverted the previous change and
added a comment to the original job.

We can decide later if we want to re-enable it. I will create an issue so that we don't forget.


DavideD:
Aargh... I've also disabled the build for 5.4 SNAPSHOT. Fixing it

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/645
Sorry for the noise

PR #647, opened 2021-03-04
[#642] CockroachDB

### LINKED ISSUE #642: Add CockroachDB to the list of of available databases
Thanks to @conorfarrell-coats, we learnt that we could support CockroachDB as well:
https://github.com/hibernate/hibernate-reactive/pull/640

We should add it to our test suite

DavideD:
There are still a couple of things I need to check but I want to check if it works with CI

Fixes #642


DavideD:
* Identifier generation from the database only works if we do an insert and select id  query. It seems that we can always do that at the moment, so I've removed the check: https://github.com/hibernate/hibernate-reactive/pull/647/commits/e0e3c136f3e1c31e5414e653bb8113dd7f10658c

* CockroachDB seems to have problems with temporary tables. It seems we use them when dealing with hierachies: I've disabled those tests

* Database identifier generation with CockroachDB uses timestamp and it seems to return negative values when the type of the id is not Long

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -397,14 +397,14 @@ void preInsertInMemoryValueGeneration(Object[] fields, Object object,
...
-			//use an extra round trip to fetch the id
-			generatedIdStage =  connection.update( sql, params )
-					.thenCompose( v -> connection.selectIdentifier( delegate().getIdentitySelectString(), EMPTY_OBJECT_ARRAY ) );
-		}
+//		}
+//		else {
+//			//use an extra round trip to fetch the id
+//			generatedIdStage =  connection.update( sql, params )
+//					.thenCompose( v -> connection.selectIdentifier( delegate().getIdentitySelectString(), EMPTY_OBJECT_ARRAY ) );
+//		}
```
gavinking, line 407: What's the reason for this change?

DavideD: When we use the extra trip the the identity generation test fails.

But why we do it? All the other databases support the single query insert select or similar

DavideD:
> [DavideD](/DavideD) requested a review from [gavinking](/gavinking) [13 seconds ago](#event-)

Sorry, didn't mean to ask for a review right now

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
gavinking: OK, I thought we needed that on MySQL, but if we don't that's fine.

gavinking: Oh now I just read your very clear commit message and now I understand. I had actually misinterpreted the change. So actually this was a good catch! 👍

gavinking:
> CockroachDB seems to have problems with temporary tables. It seems we use them when dealing with hierachies: I've disabled those tests

Yeah, that should be just the HQL build update queries. It looks like you've just disabled those tests, which is fine.

gavinking:
> Database identifier generation with CockroachDB uses timestamp and it seems to return negative values when the type of the id is not Long

That sounds a bit weird. I wonder if that's what it does in my dialect for H6, or is it just like that in H5?

gavinking:
> Database identifier generation with CockroachDB uses timestamp

Rm, according to `CockroachDB1920IdentityColumnSupport` in H5 it should be `serial4` or `serial8`. So I'm not sure I've understood your comment.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/NoJdbcEnvironmentInitiator.java
```diff
@@ -129,6 +130,9 @@ else if ( url.startsWith("postgresql:") || url.startsWith("postgres:") ) {
 		else if ( url.startsWith("db2:") ) {
 			return  DB297Dialect.class;
 		}
+		else if ( url.startsWith("cockroachdb:") ) {
+			return  CockroachDB192Dialect.class;
```
gavinking, line 134: @DavideD I think this should be `CockroachDB201Dialect`, the latest version.

gavinking (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/NoJdbcEnvironmentInitiator.java
DavideD: Thanks, I will fix it

DavideD:
> Rm, according to `CockroachDB1920IdentityColumnSupport` in H5 it should be `serial4` or `serial8`. So I'm not sure I've understood your comment.

Mmh.. yeah... it's one of those area that I need to look a bit better

DavideD:
I think this is now ready to be merged:

* The issue with temporary tables is solved by [enabling the feature in CockroachDB at startup](https://github.com/hibernate/hibernate-reactive/pull/647/files#diff-62ba851bfffa2cc562e718c4bf8b23d4edf742bd218d17e566eb670dbb948548R58);
* The problem with the id was on our side, when the id is too big for the desired type we do a narrow primitive type conversion. I've changed it so that [now we throw an exception](https://github.com/hibernate/hibernate-reactive/pull/647/commits/557a913e20841ce71689153018bcf6e45f3872ea). I suspect the right solution would be to reuse the casting`IdentifierGeneratorHelper` class in ORM, but it should be good enough for now;
* A user can use the scheme `cockroachdb` in the url but it should also work if they specify `postgres` and select the CockroachDB dialect;
* The credentials for CockroachDB we  use for testing are different than the ones for the other datastores because we are running it in insecure mode. I could replace the name of the dabatase to keep it consistent but, ultimately, it doesn't seem worth the effort for now;
* The first three commits are not really related to this PR, but I've noticed those changes while I was working on it and decided to include them anyway.

Ultimately, this PR is a bit bigger than I've anticipated but I've tried to keep the commits self contained hoping that it will make it easier to follow the logic of the changes.

gavinking (approved):
Looks great to me: merge it!

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/647

PR #662, opened 2021-03-11
Enable testEagerParentFetchQuery

DavideD:
It seems it works now, probably because issue #39 about
eager fetch select.


gavinking:
Excellent!

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/662

PR #664, opened 2021-03-12
Test against JDK17 EA

yrodiere:


yrodiere:
Maybe we should stop testing JDK15 now that 16 is GA, but personally I'd wait until Quarkus starts running their CI against JDK16.

yrodiere:
(JDK15 will EOL in about two weeks: https://endoflife.date/java)

DavideD:
> Maybe we should stop testing JDK15 now that 16 is GA, but personally I'd wait until Quarkus starts running their CI against JDK16.


I think we can wait.

yrodiere:
Looks like the Github action used to fetch the JDK does not provide JDK16 GA yet. Understandable, it's rather soon. I reverted part of my commit to only add JDK17 to the matrix.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/664

Thanks!

gavinking:
Why did this commit check in jpamodelgen-generated code?

yrodiere:
No idea... ? I just changed the Github config in my IDE.

yrodiere:
Some missing entries in .gitignore, maybe?

DavideD:
My bad, I didn't notice

DavideD:
I will update the .gitignore

DavideD:
> Some missing entries in .gitignore, maybe?

We updated the folders structure of the example and our rule wasn't generic enough.

I'm surprised they have been generated in the `src` folder and not in the `build` one though.
The fact that we had a rule implies it's not the first time it happens.

yrodiere:
> > Some missing entries in .gitignore, maybe?
> 
> We updated the folders structure of the examples and our rule wasn't generic enough.
> 
> I'm surprised they have been generated in the `src` folder and not in the `build` one though.
> The fact that we had a rule implies it's not the first time it happens.

Right, something must have gone wrong... maybe IDEA generating sources before it updated its metadata based on Gradle's? I hadn't worked on HR for a while.

Sorry about that, I'll be more careful next time.

Issue #661, opened 2021-03-11
JUnit 5


DavideD:
I was wondering if there is any particular reason for not moving to JUnit 5.

I'm asking because I suspect it will make it easier to work on these other issues:
- #648 Create a common schema for testing #648
- #487 Create the SessionFactory only once per test class

Because it uses a new Extension model.

DavideD:
@gavinking WDYT?

gavinking:
🤷 I have no opinion.

PR #665, opened 2021-03-12
Upgraded to JUnit 5

blafond:
Fixes [#661](https://github.com/hibernate/hibernate-reactive/issues/661)

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -26,6 +26,10 @@ dependencies {
     implementation "io.vertx:vertx-sql-client:${vertxVersion}"
 
     // Testing
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
+
     testImplementation 'org.assertj:assertj-core:3.13.2'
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/DefaultPortJUnit5Test.java
```diff
@@ -0,0 +1,64 @@
...
+import io.vertx.sqlclient.SqlConnectOptions;
+import org.assertj.core.api.Assertions;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.dbType;
+
+/**
+ * Test the default port is set correctly when using {@link DefaultSqlClientPoolConfiguration}
+ */
+public class DefaultPortJUnit5Test {
```
DavideD, line 29: We don't need this test, I've added it to check if we could test both version of JUnit, but we already have a `DefaultPortTest`

DavideD (changes requested):

hibernate-reactive-core/src/test/java/org/hibernate/reactive/DefaultPortJUnit5Test.java
blafond: k. np

DavideD (approved):

DavideD:
￼Looks good!
We can start merging this and then refactor gradually  the test suite as needed

PR #666, opened 2021-03-12
Update .gitignore and remove generated folders

DavideD:
I wonder why they have been generated in the src folder.
Anyway, now git will ignore any folder named `generated` and I've also updated the rule for `target`


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/666

Issue #667, opened 2021-03-12
hibernate.jdbc.time_zone=UTC


Ngosti2000:
When querying using Query i get the above error. My Entity class defines an OffsetDateTime field (code snippet before). Persisting the same works fine

```........
   @Column
    private String otp;

    @Column(name="mobile_number")
    private String mobileNumber;

    @Column(name="expiry_date")
    private OffsetDateTime expiryDate;

```
The hql query being executed

``` 
        Map<String,Object> values = new HashMap<>();
        values.put("otp",jsonObject.getString("otp"));
        values.put("msisdn",jsonObject.getString("msisdn"));
        values.put("expiryDate", OffsetDateTime.now());
        String hql="from Otp o where otp=:otp and mobileNumber=:msisdn and expiryDate > :expiryDate";
```

DavideD:
This issue seems to happen when the property `<property name="hibernate.jdbc.time_zone" value="UTC"/>` is set.

PR #668, opened 2021-03-12
#667 - Fix java.lang.UnsupportedOperationException org.hibernate.reac…

Ngosti2000:
Exception thrown when hibernate tries to map OffsetDateTime. #667 provides the details of the exception 
```
java.lang.UnsupportedOperationException
at org.hibernate.reactive.adaptor.impl.ResultSetAdaptor.getTimestamp(ResultSetAdaptor.java:698)
```

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -695,7 +695,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public Timestamp getTimestamp(String columnLabel, Calendar cal) {
-		throw new UnsupportedOperationException();
+		return getTimestamp(columnLabel);
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
@@ -11,11 +11,9 @@ import org.hibernate.annotations.UpdateTimestamp;
 import org.hibernate.cfg.Configuration;
 import org.junit.Test;
 
-import javax.persistence.Basic;
-import javax.persistence.Entity;
-import javax.persistence.GeneratedValue;
-import javax.persistence.Id;
+import javax.persistence.*;
 import java.time.Instant;
+import java.time.OffsetDateTime;
 
 public class TimestampTest extends BaseReactiveTest {
     @Override
@@ -29,6 +27,7 @@ public class TimestampTest extends BaseReactiveTest {
     public void test(TestContext context) {
         Record record = new Record();
         record.text = "initial text";
+        record.expiry=OffsetDateTime.now();
         test(context,
                 getMutinySessionFactory()
                         .withSession( session -> session.persist(record)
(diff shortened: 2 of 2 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -695,7 +695,7 @@ public Timestamp getTimestamp(int columnIndex, Calendar cal) {
 
 	@Override
 	public Timestamp getTimestamp(String columnLabel, Calendar cal) {
-		throw new UnsupportedOperationException();
+		return getTimestamp(columnLabel);
```
DavideD, line 698: Shouldn't you use the `cal` parameter to assign the expected timezone when it's not stored in the db?

DavideD (changes requested):
Thanks @Ngosti2000!
Could you also add a test case, please?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: > Shouldn't you use the `cal` parameter to assign the expected timezone when it's not stored in the db?

Probably, but if so `PreparedStatementAdaptor.setTimestamp()` is broken in the same way.

DavideD: I've just noticed that that in Postgres, if we save an OffsetDateTime we read a different value because of the inconsistency with the timezone. How come we store it as `timestamp` and not `timestampz`?

gavinking: Because postgres doesn't really support `timestamp with timezone`. You can *write* a timestamp with a zone attached, but it just converts it to the local timezone and that's what you get when you read it back.

Very few databases actually support `timestamp with timezone` correctly. It's pretty lame.

DavideD (changes requested):
Thanks for adding the test. 

The problem is that if I run the test on the `main` branch, it will pass even without your changes to `ResultSetAdaptor`.

A new test should solve the issue you are trying to solve when you send the PR.
In this case, if I run it on `main` without the fix, it should show me the `UnsupportedOperationException`.

Ngosti2000 (commented):
> Thanks for adding the test.
> 
> The problem is that if I run the test on the `main` branch, it will pass even without your changes to `ResultSetAdaptor`.
> 
> A new test should show the issue you are trying to solve when you send the PR.
> In this case, if I run it on `main` without the fix, it should show me the `UnsupportedOperationException`.

@DavideD @gavinking So i tried to reproduce the scenario on my fork but was unable to but still, the error occurred in my project. After drilling down I found my culprit on the persistence.xml ....bang!!!! This property 
```
<property name="hibernate.jdbc.time_zone" value="UTC"/>

```
After removing it, it works as expected

gavinking:
OK well at least we know how to reproduce it now.

But we still need to decide what to actually do about it. Fix it, I suppose.

DavideD:
> But we still need to decide what to actually do about it. Fix it, I suppose.

Yes, I think we should fix it.
In particular what bothers me is that a user can save an `OffsetDateTime` and then retrieve it without getting the same value

DavideD:
Thanks a lot @Ngosti2000 
I'm going to update the issue and add more details so that we can fix it properly.

DavideD:
> In particular what bothers me is that a user can save an `OffsetDateTime` and then retrieve it without getting the same value

Well, I suppose it will work if using UTC as a time zone. Still, I feel we should mention this somewhere in the documentation. We still need to investigate the use of the property

DavideD:
I'm going to close this PR for now but leave the issue open.

gavinking:
> In particular what bothers me is that a user can save an `OffsetDateTime` and then retrieve it without getting the same value

Well I think that's essentially unfixable, at least on the databases we support, because they don't properly implement `timestamp with time zone`.

gavinking:
> Well, I suppose it will work if using UTC as a time zone. 

Yeah in this _particular_ case, with that property set, we should be able to make it work, I think.

PR #670, opened 2021-03-14
Upgrade to vert.x 4.0.3

DavideD:


DavideD:
There is a problem with a test after the update to vert.x 4.0.3.

PR #674, opened 2021-03-18
Test for lazy OneToOne associations

DavideD:
From the conversation on Zulip


gavinking (approved):
Thanks!

DavideD:
It seems it doesn't work for all the dbs. I will check

DavideD:
It seems that MariaDB has some problems with UUID types used as query parameters. I've changed the test entities to a `Long id`.
It should work now.

gavinking:
Do we still have tests for UUID elsewhere?

gavinking:
Oh, of course, you only changed the new test. Sorry.

DavideD:
> Oh, of course, you only changed the new test. Sorry.

Yes, I will check this as well in a bit.
Is it me or every  time I work an on issue, I find 3 other different ones? :-)

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/674

gavinking:
Pretty sure that's normal.

PR #677, opened 2021-03-18
possibly more correct implementation of hibernate.jdbc.time_zone support

gavinking:
Hrm. I'm not positive, but after re-reading the Javadoc, I'm wondering if this is a more correct implementation of `ResultSet.getTimestamp()`?


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/677

Thanks

Issue #678, opened 2021-03-19
Id of type UUID cannot be retrieved with MariaDB and MySQL


DavideD:
The entity is in the db but a `find` will return null

DavideD:
I think this is a problem with the Vert.x SQL client but I need to check with a test

DavideD:
I've found the problem.
The table has a column defined as `binary(255)` but the UUID is created as a `binary(16)`. This work for the other dialect but not for MySQL/MariaDB with vert.x  SQL client

gavinking:
Hrm, yeah, looks this is a bug I fixed in H6, but fixing the broken default column length for `UUIDBinaryType`. I set it to 16 by default. In H5, the column length defaults to something stupid (255 apparently).

DavideD:
I've created an issue for the Vert.x SQL client: https://github.com/eclipse-vertx/vertx-sql-client/issues/922
This seems a bug considering that it works for all the other databases.

I guess a workaround at the moment is to define the type of the column:
```
		@Column(columnDefinition = "binary(16)")
		UUID id;
```

gavinking:
Honestly I'm not sure if it's a driver bug because I feel like I've run into the same issue with JDBC.

But I might be misremembering.

*Something* motivated me to go and fix the issue in H6.

gavinking:
> I guess a workaround at the moment is to define the type of the column

Yes, of course, exactly.

DavideD:
Maybe I can change the UUIDType when using reactive

gavinking:
No because in H5 a type can't determine the column length.

I had to build a whole infrastructure for that in H6. And IIRC it wasn't that trivial.

DavideD:
> Honestly I'm not sure if it's a driver bug because I feel like I've run into the same issue with JDBC.

I've tested it and it seems to work when I use Hibernate ORM and JDBC

DavideD:
> I had to build a whole infrastructure for that in H6. And IIRC it wasn't that trivial.

It seems there is not much more I can do about this at the moment.
I will wait to see what they think on the Vert.x SQL client side and, in the meanwhile, send a PR with a test case for the other dbs.

gavinking:
> > Honestly I'm not sure if it's a driver bug because I feel like I've run into the same issue with JDBC.
> 
> 
> 
> I've tested it and it seems to work when I use Hibernate ORM and JDBC

OK, fine, so I must have misremembered.

[...]

PR #680, opened 2021-03-23
[#678] Test for UUID as field type 

DavideD:
Doesn't fix the issue for MariaDB or MySQL but it checks that everything works for the other dbs


gavinking:
I thought you were going to set the column length using @Column

DavideD:
I've added the test for MariaDB and MySQL with the column defition

DavideD:
CI is a bit weird right now, I will re-launch the checks in a bit

DavideD:
Weird, GitHub wasn' t happy about the version of the gradle wrapper we were using.
I've updated it to he latest and it works now.

PR #683, opened 2021-03-29
[#487] One SessionFactory per test class

### LINKED ISSUE #487: Create one SessionFactory per class in tests
Not critical or urgent.

At the moment we open/close a session factory after each single test and therefore running the schema generation queries each time.

We could save some time creating the factory only once.
This requires:
* Add appropriate populateDb/cleanUpDb methods before and after each tests (if they are missing)
* Create the factory statically in the base class

Our build is already quite fast with some db but it might help with the slower ones.

DavideD:
Fixes #487

Work on this with @blafond.

Now the test will create the factory once per test class. This saves some time with dbs like Db2 or CockroachDB.
The only issues I had is that some times tests get stuck but I couldn't figure out why. Refactoring some tests so that they now
use transaction solved the issues most of time.

> If you do this, make sure that you recreate the factory after any test failure. Otherwise, the failure of one test method would have a cascading effect on all the other tests in that class (because its data won't be cleaned up).

@gavinking. do we really need to do this? The `cleanDb` method is called even when there is a failure. I suppose we could reset the factory if there is a problem deleting the entities but we don't have that many tests in a single class so I don't think it would be a problem. The reason I would prefer not to do it is because the only place I could do something like this is after the `deleteEntities` method has been called. I might thinkg of a better solution later.

For some reasons, this test doesn't work with Db2: https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:487-sf-per-class?expand=1#diff-659762d6654e6fae801ebfcf7577499a86f6799bd9ce7948922726acc994fbab
Unless I recreate the factory each time.

I've also removed JUnit 5, in the end we don't need it and having both dependencies on the classpath might only causes issues.
I might add it back in the future if they add an extension that's now missing in vertx-junit-5

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
@@ -32,6 +32,7 @@ import org.hibernate.reactive.vertx.VertxInstance;
 import org.hibernate.tool.schema.spi.SchemaManagementTool;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.runner.RunWith;
 
@@ -59,8 +60,8 @@ public abstract class BaseReactiveTest {
 	@Rule
 	public Timeout rule = Timeout.seconds( 5 * 60 );
 
-	@Rule
-	public RunTestOnContext vertxContextRule = new RunTestOnContext( () -> {
+	@ClassRule
+	public static RunTestOnContext vertxContextRule = new RunTestOnContext( () -> {
 		VertxOptions options = new VertxOptions();
 		options.setBlockedThreadCheckInterval( 5 );
 		options.setBlockedThreadCheckIntervalUnit( TimeUnit.MINUTES );
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
@@ -6,6 +6,8 @@
 package org.hibernate.reactive;
 
 import io.smallrye.mutiny.Uni;
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.RunTestOnContext;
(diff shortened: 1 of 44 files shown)


gavinking:
> do we really need to do this?

Well I think so, yes. I mean, I don't trust us to always write a `cleanDb()` method. Indeed, that is a total PITA to have to do.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -127,6 +130,26 @@ protected Configuration constructConfiguration() {
...
+		return deleteEntities( Arrays.stream( entities )
+								  .map( BaseReactiveTest::defaultEntityName )
+								  .collect( Collectors.toList() )
+								  .toArray( new String[entities.length] ) );
+	}
+
+	private static String defaultEntityName(Class<?> aClass) {
+		int index = aClass.getName().lastIndexOf( '.' );
+		index = index > -1 ? index : 0;
+		return aClass.getName().substring( index );
```
blafond, line 146: Complex Entity names that include a `.` will error due to the initial index for `.` is used as the start of the substring 
 and ends up with `.SomeName$Othername`

DavideD:
> Well I think so, yes. I mean, I don't trust us to always write a `cleanDb()` method. Indeed, that is a total PITA to have to do.

@Barry is having a look to see if we can find the name of the entities and create  a default `deleteEntities` that will work for most cases

DavideD:
Something doesn't work as expected on the latest JDKs.
I've tried a quick experiment and noticed that the error message is different. Instead of containing the expected
`Detected use of the reactive Session from a different Thread` it says `Session/EntityManager is closed` (
https://github.com/hibernate/hibernate-reactive/pull/683/checks?check_run_id=2221537360#step:7:988)

Note that, because I wanted to see the error message, I've changed the assertion for that specific run. The assertion failing is actually: https://github.com/hibernate/hibernate-reactive/blob/29733eb91c01d0b435f73a1630640410cf705ef4/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultipleContextTest.java#L57

It would be interesting to know why  there is a different error.
I will be on holiday until next week, if somebodys want to have a look at it is more than welcome, otherwise I will get back to it later.

DavideD:
I think this is ready to be merged now

blafond:
Looks good.  Clean implementation and provides test classes option to manage DB entities life cycle.

blafond (approved):

DavideD:
The build is still failing on CI sometimes. I'm having a look

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/683

We can tweak it later when problems occur.

Issue #707, opened 2021-04-15
BlockingIdentifierGenerator is broken


gavinking:
See https://github.com/quarkusio/quarkus/issues/16255.

DavideD:
@gavinking are you going to work on this or should I do it?

gavinking:
Yes, I'm on it.

DavideD:
Thanks

PR #708, opened 2021-04-15
proposed fix for #707

gavinking:
Make sure that queued requests use the right `session` and `entity` args.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
@@ -54,6 +54,11 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
 
     @Override
     public CompletionStage<Long> generate(ReactiveConnectionSupplier session, Object entity) {
+        if ( getBlockSize()<=1 ) {
+            //special case where we're not using blocking at all
+            return nextHiValue(session);
+        }
+
         long local = next();
         if ( local >= 0 ) {
             // We don't need to update or initialize the hi
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
@@ -38,7 +38,7 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
     private int loValue;
     private long hiValue;
 
-    private volatile List<CompletableFuture<Long>> queue = null;
+    private volatile List<Runnable> queue = null;
 
     protected synchronized long next() {
         return loValue>0 && loValue<getBlockSize()
@@ -70,7 +70,7 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
                     // go off and fetch the next hi value from db
                     nextHiValue(session).thenAccept( id -> {
 //						Vertx.currentContext().runOnContext(v -> {
-                        List<CompletableFuture<Long>> list;
+                        List<Runnable> list;
                         synchronized (this) {
                             // clone ref to the queue
                             list = queue;
@@ -79,14 +79,14 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
                             result.complete( next(id) );
                         }
                         // send waiting streams back to try again
-                        list.forEach( completion -> generate(session, entity)
-                                .thenAccept(completion::complete) );
+                        list.forEach(Runnable::run);
 //						} );
                     } );
                 }
                 else {
                     // wait for the concurrent fetch to complete
-                    queue.add(result);
+                    // note that we carefully capture the right session,entity here!
+                    queue.add( () -> generate(session, entity).thenAccept(result::complete) );
                 }
                 return result;
             }


gavinking:
OK, @Sanne, @DavideD, @stuartwdouglas, WDYT of this?

Obviously I have not tested it.

Sanne:
I think we need good tests for this area, not least to check how this behaves when the consumer rate is higher than our ability to produce new ids.

gavinking:
> I think we need good tests for this area

Well, yeah, clearly. The problem is that we have to precisely no tests for concurrent operation in the HR project, and I'm not clear even on the right way to write them.

gavinking:
(I force-pushed to remove the useless `Queued` class and just use a `Runnable` instead.)

stuartwdouglas:
Looks good, but if we have n consumers waiting maybe we should make sure we request at least n IDs? At the moment by default it is one at a time, so this is basically a global lock in terms of persist performance.

gavinking:
> Looks good, but if we have n consumers waiting maybe we should make sure we request at least n IDs? At the moment by default it is one at a time, so this is basically a global lock in terms of persist performance.

Well the assumption I made was that you always choose a big enough block size that you "rarely" need to request a new block. If that's true, then I think having more requests waiting than the size of the block is an extremely low-probability event.

On the other hand, what I realized earlier was that, ooops, what about people who *don't* want blocks, and set the block size to 1? We should actually be bypassing all this logic entirely in that case, and we don't. Not quite sure how I overlooked that.

gavinking:
> if we have n consumers waiting maybe we should make sure we request at least n IDs?

Also, at the time we request an hi value, we have zero consumers waiting.

gavinking:
> Not quite sure how I overlooked that.

Well, hrm, actually I supposed it's *not* entirely clear what the best behavior here is.

For *sequences*, yes, I think the right thing to do is not *not* wait for concurrent requests to complete.

For *table* generation it's not so clear, however.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
```diff
@@ -54,6 +54,11 @@ protected synchronized long next(long hi) {
 
     @Override
     public CompletionStage<Long> generate(ReactiveConnectionSupplier session, Object entity) {
+        if ( getBlockSize()<=1 ) {
+            //special case where we're not using blocking at all
+            return nextHiValue(session);
+        }
+
```
gavinking, line 61: @stuartwdouglas WDYT?

Sanne:
FWIW I had written a POC about a year ago to have such an optimizer, which would tune the page sized in an "adaptative" way.

I abandoned it as I needed to learn how to test such things with vertx and never had much time - but it sounds like conceptually you might want to revive the idea.

gavinking:
I mean in principle it would not be at all hard to increase the block size if we find ourselves requesting too many new blocks. Not hard at all.

Sanne:
Right - and there's several other smart things one can easily do in the reactive model, such as we could already trigger a block allocation before the last available ID has been handed out to clients.

So by observing the rate at which IDs are being consumed and the time it takes for block allocation tasks to complete it should be possible to:

- prefetch ID blocks in advance
- adapt the block sizes

One caveat of using dynamically allocated block sizes might be fragmentation; if there's multiple application nodes they still need to use a consistent strategy - should not be a problem when it's based on sequences (atomic) but if it's backed by some of the other models (table?) this is probably not a good idea.

gavinking:
@stuartwdouglas can you do us a favor and test this patch in the context of the Quarkus issue?

DavideD:
> > I think we need good tests for this area
> 
> Well, yeah, clearly. The problem is that we have to precisely no tests for concurrent operation in the HR project, and I'm not clear even on the right way to write them.

I think I have an idea, I will start prototyping something

stuartwdouglas:
This works now, instead of doing a couple of requests then hanging it can now handle 3.5k inserts per second (this is not tuned in any way, it could probably do a lot more if I messed with some settings).

gavinking:
Thanks 🙏 

So perhaps it's worth doing a release just to fix this one thing. Thoughts?

stuartwdouglas:
I think so, its a pretty bad bug.

stuartwdouglas:
Actually I just realized that I am only testing the  block size == 1 path, how do I configure a larger block size to test the other code path?

gavinking:
With the `@SequenceGenerator` or `@TableGenerator` annotation just like in regular JPA.

DavideD:
> So perhaps it's worth doing a release just to fix this one thing. Thoughts?

Yes, we should release it. I will work on adding some more tests for the id generation but considering that @stuartwdouglas already tested on quarkus and that our current test suite works, releasing this version seems an important improvement for a release.

stuartwdouglas:
It has been more than 10 years since I last actually 'used' JPA rather than just dealing with integration stuff. So it looks like the quickstart sets allocationSize = 1 for some reason, if you remove it then performance increases, but we get the occasional PK violation:

```
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.86ms   17.38ms 219.92ms   98.11%
    Req/Sec     8.11k     1.49k    9.41k    90.86%
  159144 requests in 10.10s, 15.69MB read
  Non-2xx or 3xx responses: 31

```

Looking at the logs some entities are being given the same PK.

gavinking:
> we get the occasional PK violation

is this with a sequence generator or a table generator?

stuartwdouglas:
```
    @SequenceGenerator(name = "fruitsSequence", sequenceName = "known_fruits_id_seq", initialValue = 10)
```

gavinking:
Well that's very strange. Can you tell if it's because they are assigned dupe hi values or dupe lo values?

gavinking:
I mean: 

1. a dupe hi value is impossible because we're getting them from a database sequence, but
2. a dupe lo value also looks impossible because lo values are assigned within a `synchronized` block.

So ... another silly logical error in my code??

stuartwdouglas:
The id's in the errors are a long way apart, e.g:
5161
6361
8561
13211

So it's not allocating duplicate blocks.

stuartwdouglas:
They all seem to end in '1'

gavinking:
> another silly logical error in my code

Yup it's that, actually. Hold tight I will push a fix.

gavinking:
wait, *all* the ids, or just all the *dupe* ids?

gavinking:
> Yup it's that, actually. Hold tight I will push a fix.

No, excuse me. What I thought was wrong is actually fine.

gavinking:
> wait, all the ids, or just all the dupe ids?

No, you mean the ones in the errors. So by the looks, they're all the "second id generated after fetching a new hi value".

gavinking:
> they're all the "second id generated after fetching a new hi value".

No, wait I guess you have `initialValue=1`, the default, so they're the *first* id's in their respective blocks.

stuartwdouglas:
I had to step away and have dinner, but my guess is something to do with
reentrant calls, so the sync block doesn't help. I can't see how though.

gavinking:
Do you think it could be an out-of-order execution thing??

Like, `list.forEach(Runnable::run)` gets executed before ` result.complete( next(id) )`.

Or am I trying to make this too complicated again?

gavinking:
I mean, would moving the line `list.forEach(Runnable::run)` inside the `synchronized` block fix it?

stuartwdouglas:
Bugger, turns out this was user error. I forgot to clean before running quarkus:dev so it had cached the old maven dependencies, so it was not picking up the new version. It looks like this was a problem with the old version when the block size was larger than 1.

Sorry for the noise.

gavinking:
No problem. So how does it perform?

stuartwdouglas:
Comparatively way better, 13,429 inserts per sec. This is no no way a proper benchmark, it's only using 2 threads and 10 connections.

gavinking:
Alright. So that sounds good.

Shall I merge this?

DavideD:
Sounds good to me

DavideD:
I've released it and sent a PR for quarkus: https://github.com/quarkusio/quarkus/pull/16588

DavideD:
Thanks a lot @stuartwdouglas

PR #715, opened 2021-04-20
Improve ProxyConnection error handling

stuartwdouglas:
- Add better error message if it has been closed
- Don't explicity throw the exception, return a failed CS instead

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
@@ -7,6 +7,7 @@ package org.hibernate.reactive.pool.impl;
 
 import java.sql.ResultSet;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
 
@@ -24,6 +25,7 @@ final class ProxyConnection implements ReactiveConnection {
 	private final ReactiveConnectionPool sqlClientPool;
 	private ReactiveConnection connection;
 	private boolean connected;
+	private boolean closed;
 	private final String tenantId;
 
 	public ProxyConnection(ReactiveConnectionPool sqlClientPool) {
@@ -38,6 +40,11 @@ final class ProxyConnection implements ReactiveConnection {
 
 	private <T> CompletionStage<T> withConnection(Function<ReactiveConnection, CompletionStage<T>> operation) {
 		assertUseOnEventLoop();
+		if ( closed ) {
+			CompletableFuture<T> ret = new CompletableFuture<>();
+			ret.completeExceptionally( new IllegalStateException( "session is closed" ) );
+			return ret;
+		}
 		if ( !connected ) {
 			connected = true; // we're not allowed to fetch two connections!
 			CompletionStage<ReactiveConnection> connection =
@@ -49,7 +56,9 @@ final class ProxyConnection implements ReactiveConnection {
 			if ( connection == null ) {
 				// we're already in the process of fetching a connection,
 				// so this must be an illegal concurrent call
-				throw new IllegalStateException( "session is currently connecting to database" );
+				CompletableFuture<T> ret = new CompletableFuture<>();
+				ret.completeExceptionally( new IllegalStateException( "session is currently connecting to database" ) );
+				return ret;
 			}
 			return operation.apply( connection );
 		}
@@ -140,6 +149,7 @@ final class ProxyConnection implements ReactiveConnection {
 			connection.close();
 			connection = null;
 		}
+		closed = true;
 	}
 
 }


DavideD:
@gavinking How come [`ReactiveConnection#close()` ](https://github.com/hibernate/hibernate-reactive/blob/65ff89ad5594e36cb2152bf08a2ef0c6807b2bdc/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java#L64) doesn't return a `CompletionStage<Void>`?

DavideD:
I'm asking because `ProxyConnection` delegates to `SqlClientConnection` and it seems to me that we should add the same logic in that class as well.

gavinking:
> @gavinking How come [`ReactiveConnection#close()` ](https://github.com/hibernate/hibernate-reactive/blob/65ff89ad5594e36cb2152bf08a2ef0c6807b2bdc/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java#L64) doesn't return a `CompletionStage<Void>`?

Well I guess because we wanted `Session.close()` to be synchronous, and because in Vert.x 3, `SqlClient.close()` was synchronous.

Looks like it returns a `Future` in Vert.x 4, and so we have a problem :-(

DavideD:
I suppose we can change it for the next release?

gavinking:
> I suppose we can change it for the next release?

Well, we'll *have to*. But I don't see how we can fix this without changing APIs.

`Session` can't even be `AutoCloseable` now.

Ugh. Wish we had caught this sooner.

gavinking:
Unless there's some way to do it fire-and-forget???

gavinking:
> Unless there's some way to do it fire-and-forget???

Well, perhaps it is implicitly fire-and-forget. If I just ignore the returned `Future`, it will still get closed, right?

gavinking:
> If I just ignore the returned `Future`, it will still get closed, right?

But then if we do that maybe the rest of our stream never gets completed?

@cescoffier @jponge need your help on understanding this one.

DavideD:
> Well, perhaps it is implicitly fire-and-forget. If I just ignore the returned `Future`, it will still get closed, right?

It probably doesn't happen often, but a user might want to wait for the session or connection to be really closed before doing something else.

DavideD:
> But then if we do that maybe the rest of our stream never gets completed?

What do you mean?

gavinking:
> What do you mean?

Well TBH I'm just not clear enough on Vert.x semantics to know what happens in the case where we throw away a `Future`.

gavinking:
> A user might want to wait for the session or connection to be really closed before doing something else.

I can't *imagine* a case where the user would care. But I admit I can't imagine everything.

DavideD:
> Well TBH I'm just not clear enough on Vert.x semantics to know what happens in the case where we throw away a `Future`.

Isn't this more of a conceptual issue? If we don't wait for the `Future` to be completed, the connection might not have been closed when we run the next operation and we loose exceptions if something went wrong.

I wonder if this happens also when the session is closed right now.

gavinking:
> Isn't this more of a conceptual issue? 

Absolutely it's a conceptual issue. 

That's what I'm saying, this is a potentially serious problem.

> If we don't wait for the `Future` to be completed, the connection might not have been closed when we run the next operation and we loose exceptions if something went wrong.

Well I agree we should log them at least. But I would say that the failure to return a connection to the pool when the transaction has already completed successfully is something the users of my program don't care about.

The *administrators* certainly might care, and so we should certainly log such a thing.

DavideD:
> `Session` can't even be `AutoCloseable` now.

Is this that bad? We already have `.withSession` that does the same thing. Plus the `Autoclosable` is not for reactive things.

gavinking:
Well, it's not a *disaster*.

gavinking:
I mean, I don't think you can use try with resources to manage a reactive session, so if you interpret AutoClosesble as being all about that one thing, then Session shouldn't implement it anyway. (That's not clear to me.)

DavideD:
One alternative I can think of is that we could overload `close` to something like:
```
void close(CompletableFuture<Void> closed);
```

So that a user could pass a `CompletableFuture` that we can complete when everything has been closed.
But I don't like this solution, I think it would teach bad habits and also confuse people quite a bit.

I think we should update the API.

DavideD:
I mean, add the `CompletionStage<Void>` to the `close()` method.

DavideD:
> I mean, I don't think you can ever use try with resources to manage a reactive session, so if you interpret AutoClosesble as being all about that one thing, then Session shouldn't implement it anyway.

+1

gavinking:
Sure, that's probably best, but I can't help wondering why the operation of "return a connection to the pool" needs to be async in the first place. There's probably some really good reason, but I don't know what it might be.

DavideD:
I've created https://github.com/hibernate/hibernate-reactive/issues/716

DavideD:
Merged  https://github.com/hibernate/hibernate-reactive/pull/715, Thanks @stuartwdouglas !

gavinking:
> Sure, that's probably best, but I can't help wondering why the operation of "return a connection to the pool" needs to be async in the first place.

Well, it looks like one reason is that if you return a connection to the pool with an open transaction, it rolls it back. I guess that's a good enough reason.

PR #717, opened 2021-04-20
[#709]  Fix orphan removal when setting association to null

### LINKED ISSUE #709: Cascade calls non reactive delete
[We have some logic in our code that could end in a call to a non reactive method](https://github.com/hibernate/hibernate-reactive/blob/d17862792b266d29ace9692bd305c29b48f034a4/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java#L363).
In `Cascade`:
```
if ( type.isAssociationType()
		&& ( (AssociationType) type ).getForeignKeyDirection()
				.equals(ForeignKeyDirection.TO_PARENT) ) {
       // Related to issue #702
      ...
else {
	// Else, we must delete after the updates.
	eventSource.delete( entityName, loadedValue, isCascadeDeleteEnabled, new HashSet<>() );
}
```

We need a test to cover this part and make sure that the code calls the equivalent reactive method (it needs to be created).
I think the result will be something similar to:
```
else {
	// Else, we must delete after the updates.
	stage = stage.thenCompose(v -> eventSource.reactiveDelete( entityName, loadedValue, isCascadeDeleteEnabled, new HashSet<>() ));
}
```

DavideD:
Fixes #709

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
@@ -34,7 +34,6 @@ import org.hibernate.type.Type;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.concurrent.CompletionStage;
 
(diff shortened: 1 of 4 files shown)


blafond (approved):

gavinking:
Looks fine.

DavideD:
I'm merging this one, it seems the build fails because it has trouble downloading the latest SNAPSHOTS. I will check that later

PR #720, opened 2021-04-22
Test against JDK16 GA instead of EA

yrodiere:


DavideD:
Thanks https://github.com/hibernate/hibernate-reactive/pull/720!

yrodiere:
No problem #1!

PR #724, opened 2021-04-26
Upgrade to 5.4.31.Final and fix compatibility with ORM 5.5

### LINKED ISSUE #646: Re-enable workflow for 5.5 ORM SNAPSHOT
I've currently disabled the workflow for 5.5 as it was causing too many errors and I don't think it's fixable without
reverting a change in ORM 5.5 or a new release of ORM 5.4.

After talking to @Sanne about this, I'm not even sure if we need to continue tracking it.
I've created this issue as a reminder and in case somebody has a simple solution for making it works.

DavideD:
This is a draft because we need to upgrade to the next ORM 5.4 release before we can merge it.

Fixes #646

### DIFF
--- a/.github/workflows/tracking-orm-5.build.yml
+++ b/.github/workflows/tracking-orm-5.build.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         example: [ 'session-example', 'native-sql-example' ]
-        orm-version: ['[5.4,5.5)']
+        orm-version: [ '[5.4,5.5)','[5.5,5.6)' ]
         db: ['MySQL', 'PostgreSQL']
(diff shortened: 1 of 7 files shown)


gavinking:
This is cleaner anyway.

Sanne:
force-pushed to trigger the CI jobs again

Sanne:
Looks like we have some timeouts on DB2

DavideD:
Yeah, I'm checking if I can find some workaround :-/

DavideD:
@Sanne it seems to work now

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/724

Issue #716, opened 2021-04-20
ReactiveConnection#close should return a `CompletionStage` 


DavideD:
See the discussion on https://github.com/hibernate/hibernate-reactive/pull/715

DavideD:
I think the main issue is that now `ReactiveSessionImpl` cannot extend `SessionImpl` if we change the method for the session as well

gavinking:
> I think the main issue is that now `ReactiveSessionImpl` cannot extend `SessionImpl` if we change the method for the session as well

I don't think that's an issue at all. `ReactiveSessionImpl` would have a `reactiveClose()` method that would be called by `StageSessionImpl.close()` and `MutinySessionImpl.close()`.

DavideD:
Merged

PR #725, opened 2021-04-26
Convert ReactiveConnection#close to a reactive method

DavideD:
Fixes #712 

This should fix the issue about the close method not being reactive, I had to adapt our tests as well so there are a lot of changes to the test suite.

I'm not sure about the impact on Quarkus, considering that they are close to a release, it would be safer to merge this after their next release.


gavinking (approved):

gavinking:
@DavideD why's this still a draft?

DavideD:
No reason, I just need to update the tests we merged yesterday and then I can switch it a regular PR

DavideD:
@Sanne What do you think? In relation to Quarkus, is it safer to wait and merge it after the next release or we can merge it now?

Sanne:
I think we need to sort these aspects quickly, so I'm inclined to suggest mering it now - but I wonder if t hat implies that we won't be able to upgrade Hibernate Reactive in Quarkus for a while as I understand it will require some changes there too?

Did you have a look at how significant the changes would need to be on that side?

As you know there's an upgrade of ORM which needs to get in Quarkus soon, which will break Hibernate Reactive so we'll need to be able to relase an HR build which keeps the Quarkus build afloat while this other work is settled.

gavinking:
I think it's very important that we get breaking API changes released as soon as possible, to avoid creating more problems for users. If that includes fixes to the Quarkus extension and to Panache Reactive, then they also need to be prioritized.

Sanne:
agreed @gavinking , but when we have both ORM core and HR bringing breaking changes to the table, we might need to address one at a time - which in this case would imply having an HR release which adapts to the changes in ORM but doesn't break other things.

The alternative is to send both upgrades combined to Quarkus, which is entirely an option but I'm just warning that for it to be feasible we need to have an idea of the complexity: I can't delay the ORM upgrade in Quarkus further as there's many other works depending on it.

gavinking:
Right, but I mean: it's _pretty bad_ that we're not releasing connections!

Sanne:
ah yes it's very bad. Don't get me wrong, I want both breaking changes integrated ASAP - just figuring out what's the fastest way to have it all at least resistance.

Sanne:
Interestingly we can merge this as it doesn't need any changes to compile in Quarkus.

Obviously we'll need to make changes to actually close the resources, but we don't need to wait with merging this particular patch.

DavideD:
Merged!

DavideD:
Damn! Wrong issue number in the commit messages. It's actually #716

PR #732, opened 2021-04-28
fix #726 npe when logging enabled

### LINKED ISSUE #726: Setting hibernate logging to debug causes NullPointerException
To quickly see it, it's enough to set this in our test suite:
```
logger.hibernate.name =	org.hibernate
logger.hibernate.level = debug
```
The Exception:
```
Caused by: java.lang.NullPointerException
	at org.hibernate.reactive.persister.collection.impl.ReactiveBasicCollectionPersister.getSQLInsertRowString(ReactiveBasicCollectionPersister.java:141)
	at org.hibernate.persister.collection.AbstractCollectionPersister.logStaticSQL(AbstractCollectionPersister.java:693)
	at org.hibernate.persister.collection.AbstractCollectionPersister.<init>(AbstractCollectionPersister.java:529)
	at org.hibernate.persister.collection.BasicCollectionPersister.<init>(BasicCollectionPersister.java:59)
	at org.hibernate.reactive.persister.collection.impl.ReactiveBasicCollectionPersister.<init>(ReactiveBasicCollectionPersister.java:48)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.hibernate.persister.internal.PersisterFactoryImpl.createCollectionPersister(PersisterFactoryImpl.java:152)
```

It happens because Hibernate ORM log static queries in the `AbstractCollectionPersister` in the constructor.
The log operation calls `getSQLInsertRowString` that it is overriden by our subclasses because we need to parse the generated queries. Because the log is in the constructor, our parser has not been created yet.
See for example: https://github.com/hibernate/hibernate-reactive/blob/f283b6e8e4e0cb59add668ea70620af860924962/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java#L144

gavinking:
Follow the pattern used in the entity persisters, which I guess also slightly improves performance.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveBasicCollectionPersister.java
@@ -39,14 +39,16 @@ import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
  */
 public class ReactiveBasicCollectionPersister extends BasicCollectionPersister
 		implements ReactiveAbstractCollectionPersister {
-	private final Parameters parameters;
+
+	private Parameters parameters() {
+		return Parameters.instance( getFactory().getJdbcServices().getDialect() );
+	}
 
 	public ReactiveBasicCollectionPersister(Collection collectionBinding,
 											CollectionDataAccess cacheAccessStrategy,
 											PersisterCreationContext creationContext)
 			throws MappingException, CacheException {
 		super( collectionBinding, cacheAccessStrategy, creationContext );
-		this.parameters = Parameters.instance( getFactory().getJdbcServices().getDialect() );
 	}
 
 	public CompletionStage<Void> reactiveInitialize(Serializable key,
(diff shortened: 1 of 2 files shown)


gavinking:
@DavideD know why it's failing? Tests run fine locally.

DavideD (commented):
Would it be possible to change the tests log level to debug only for `AbstractCollectionPersister` ?
It would be annoying if in the future we decide to refactor the code and forget completely about this.

DavideD:
> @DavideD know why it's failing? Tests run fine locally.

Yeah, it's the build with the latest ORM snapshots.  The latest ORM snapshots require those changes that I cannot include in master until the release ORM 5.4.31.Final.

Anyway, the important build on CI with the current stable ORM, works: https://github.com/hibernate/hibernate-reactive/actions/runs/793162539

blafond:
I tested locally and am getting  similar NPE in:  https://github.com/gavinking/hibernate-reactive/blob/726/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/collection/impl/ReactiveOneToManyPersister.java

Same problem with `parameters` since `ReactiveOneToManyPersister` doesn't extend `ReactiveBasicCollectionPersister`

gavinking:
> Same problem with `parameters` since `ReactiveOneToManyPersister` doesn't extend `ReactiveBasicCollectionPersister`

Ah, OK, thanks, let me fix that one too.

gavinking:
OK, I replicated the same fix to `ReactiveOneToManyPersister`.

DavideD:
@gavinking what about this?

> Would it be possible to change the tests log level to debug only for AbstractCollectionPersister ?
It would be annoying if in the future we decide to refactor the code and forget completely about this.

gavinking:
@DavideD we don't in general test things with debug logging turned on. So i don't think it makes sense to do this for this special case.

Now, if you wanted to add a separate build job that runs the test suite with *all* the debug logging enabled, I guess in principle that might be vaguely useful, but it will surely be slow, and I don't want to do anything to make our test suite slower in normal operation/

DavideD:
> @DavideD we don't in general test things with debug logging turned on. So i don't think it makes sense to do this for this special case.

I think it make sense to make an exception for this case because it is pretty easy to miss (as you have just noticed).

gavinking:
@DavideD just trust me, we really don't want to do this. You really don't want a test suite which spews out a bunch of useless log messages making it harder to interpret the output.

I just wasted like half an hour of my life, that I will never get back, trying to make this work for you by setting up a separate log4j appender that throws away the log messages it receives, but that is apparently impossible to make work because of the structure of our log namespaces ad because `log4j.properties` just isn't really flexible enough.

That's more than enough time to waste on this.

DavideD:
Thank you for having a look to see if it was doable.

gavinking:
> Please, don't tell me what I want or don't. Thank you very much.

Davide it's a figure of speech. "You" in ordinary English conversation doesn't always mean you Davide personally. I can mean "you" in general. Most especially in the phrase "you don't want to...".

DavideD:
It sounded more ironic than anything in my  head when I wrote it :-)

Merged, thanks

gavinking:
Oops. Given that I'm the absolute king of easily-misinterpreted ironic overstatement, I should have been more sensitive to this possibility. 🤷‍♂️

PR #735, opened 2021-04-28
[#710] Rename tables used for tests

### LINKED ISSUE #710: Support for `@MapsId` and `@PrimaryKeyJoinColumn`
We don't have any tests at the moment.

[See examples ](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#identifiers-derived) in the Hibernate ORM documentation.

DavideD:
@blafond  Renaming the tables seems to fix the build. I don't know what's the problem with those names

Fixes #710


DavideD:
I'm going to merge it, it fixes the build and I the problem is because our tests don't drop the tables when they have finished and it creates conflicts with an already existing table person

PR #737, opened 2021-04-29
[#736] Closing of Vert.x instance should be waited on

### LINKED ISSUE #736: Closing of Vert.x instance should be waited on
We also need to patch `DefaultVertxInstance#stop()` to actually wait for the shutdown of the instance.

I'm assume we could "just" block in this case? But it would be nice to have a way to do it properly; maybe introducing a reactive flavour of `Startable` and `Stoppable` would make sense even for ORM blocking so to allow parallelising some setup/teardown operations.

Sanne:
Fixes #736


gavinking:
lgtm

Issue #721, opened 2021-04-23
Investigate how to create a template for JBang to help with bug reports


DavideD:
See https://github.com/jbangdev/jbang#init-templates

It would be nice if users could generate a basic test project using [JBang](https://www.jbang.dev/) so that they have a quick way to start creating a project for reproducing the issue they are having.

I haven't checked if this is feasible or helpful yet. In my head it should make it even easier than [cloning a template from our repository](https://github.com/hibernate/hibernate-test-case-templates).

gavinking:
@maxandersen already played with this. What he didn't like, and what might also be a problem here, is that he couldn't easily use a testcontainers database.

maxandersen:
It works but yeah users would need to manually setup the database. 

But so should users using more complex setups :)

What you can do is to have the jbang script setup via test containers manually rather than using the jdbc url magic which does not work (yet) with hibernate reactive.

gavinking:
I did look briefly at making the testcontainers bootstrap code in our test suite reusable, but it wasn't going to be that trivial.

DavideD:
I mean... testcontainers is a great tool... but I'm going out on a limb here and say that we can expect people using Hibernate Reactive (or ORM), to be able to start a db without it :-)

DavideD:
It certainly not a deal breaker to me, most of the boring part is really the configuration of the project. If I have to start a db  launching a script manually before being able to launch the tests or the program. it would already be a great improvement.

gavinking:
Sure, that's fine. But it's a nice-to-have.

DavideD:
Thinking about it, we probably don't even need a template to begin with, we could provide a java file with all the dependencies needed for JBang to run it.

gavinking:
I figured by "template" you were hoping to parameterize by database type.

DavideD:
I was talking about a JBang template, something that would allow the user to write:
```
jbang init --template=hibernate-reactive Issue456.java
```
but I guess this would also be possible:
```
jbang init --template=hibernate-reactive-db2 Issue456.java
```

But I think we could achieve the same result providing a single file (or one per database) so that a user can run it with:
```
jbang HReactiveWithDb2.java
```
With the caveat that a Db2 instance with the right credentials must be already running.

Then they could start editing it with:
```
 jbang edit --open=idea  HReactiveWithDb2.java
```
We can think of ways to add all the bells and whistles we want, but this would be a good starting point for me.

DavideD:
[...]

PR #749, opened 2021-05-10
draft - generate junit test from jbang template

blafond:
targets issue [#721](https://github.com/hibernate/hibernate-reactive/issues/721)


DavideD (commented):
I will have a better look at this PR tomorrow but at first glance, I don't think we need all the utility methods in `CoreReactiveTest`. We've added those methods because we run a test suite with hundreds of different tests on multiple databases. This issue is supposed to be for a user wanting to reproduce a single problem for us.

So, for example, we don't need the `SessionFactoryManager` class and maybe the `deleteAll` methods.
We should also avoid the `openSession` and `openStateless` approach and use instead the `getSessionFactory().withSession` approach (the recommended one). Another reason is that in the tests the session is closed automatically, but people looking at the code might not notice it and get the wrong idea.

I'm also not sure there is any real benefit of using templates unless they are already included in JBang. The user need an extra-step to use it if it's not already included. Maybe we should propose the template as a second step when we agree what the result should look like.

In general, I'm not sure if we should base this on our tests, even if it would be easier for us to include the result to our suite. I suspect that for a user it would be easier if we'd use the public API (that they should be already familiar with and there is a lot of documentation). I would base it more to the examples we already have.

examples/usertests/resources/CoreReactiveTest.java
```diff
@@ -0,0 +1,316 @@
...
+	private static boolean doneTablespace;
+
+	protected Configuration constructConfiguration() {
+		Configuration configuration = new Configuration();
+		String jdbcURL = "jdbc:postgresql://localhost:5432/hreact?loggerLevel=OFF&user=hreact&password=hreact";
+		String dbType = "PostgreSQL";
+		configuration.setProperty( Settings.HBM2DDL_AUTO, "create" );
+		configuration.setProperty( "db", dbType );
+		configuration.setProperty( Settings.URL, jdbcURL );
+		if ( configuration.getProperty( "db" ).equals( "DB2" ) && !doneTablespace ) {
```
DavideD, line 121: I don't think this configuration for DB2 is needed most of the time, we could omit it.
It makes sense only for specific use cases

examples/usertests/resources/reactiveTestTemplate.java.qute
```diff
@@ -0,0 +1,87 @@
...
+
+@RunWith(VertxUnitRunner.class)
+public class {baseName} extends CoreReactiveTest {
+
+	@Override
+	protected Configuration constructConfiguration() {
+		Configuration configuration = super.constructConfiguration();
+		// SET JDBC LOGIN INFORMATION
+		configuration.setProperty( Settings.URL, "{hibernate connection url}";
+		configuration.setProperty( Settings.USER, "{hibernate connection username}"; // OPTIONAL if included in URL
```
DavideD, line 37: [Username is not optional](https://github.com/hibernate/hibernate-reactive/blob/134fca9565f9b14fc39cc4e7f2c9ec27518aa851/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L134)

examples/usertests/resources/reactiveTestTemplate.java.qute
```diff
@@ -0,0 +1,87 @@
...
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+
+import org.assertj.core.api.Assertions;
+
+@RunWith(VertxUnitRunner.class)
+public class {baseName} extends CoreReactiveTest {
+
+	@Override
+	protected Configuration constructConfiguration() {
+		Configuration configuration = super.constructConfiguration();
```
DavideD, line 34: If we keep all the configuration in one place, it's easier for the user to see how we set all the different properties.
I would remove it from the super class

examples/usertests/resources/reactiveTestTemplate.java.qute
```diff
@@ -0,0 +1,87 @@
...
+		// SET JDBC LOGIN INFORMATION
+		configuration.setProperty( Settings.URL, "{hibernate connection url}";
+		configuration.setProperty( Settings.USER, "{hibernate connection username}"; // OPTIONAL if included in URL
+		configuration.setProperty( Settings.PASS, "{hibernate connection password}"; // OPTIONAL if included in URL
+		configuration.addPackage(this.getClass().getPackage().getName());
```
DavideD, line 39: I don't think `addPackage` is necessary for most cases (I might be wrong). We could remove this configuration.

examples/usertests/resources/reactiveTestTemplate.java.qute
DavideD: Are these placeholder for something that JBang will replace?

examples/usertests/resources/reactiveTestTemplate.java.qute
```diff
@@ -0,0 +1,87 @@
...
+//DEPS org.hibernate.reactive:hibernate-reactive-core:1.0.0.CR4
+//DEPS org.assertj:assertj-core:3.13.2
+//DEPS io.vertx:vertx-core:4.0.3
+//DEPS io.vertx:vertx-unit:4.0.3
+//DEPS io.vertx:vertx-pg-client:4.0.3
+//FILES CoreReactiveTest.java
+//
+//
+
+import javax.persistence.*;
```
DavideD, line 14: Not critical, but let's avoid `*` imports.

examples/usertests/resources/reactiveTestTemplate.java.qute
DavideD: If not we should initialize everything with default values for PostgreSQL (to begin with)

examples/usertests/resources/Instructions.md
```diff
@@ -0,0 +1,34 @@
...
+   [jbang] No explicit target pattern was set, using first file: {basename}.java=./resources/reactiveTestTemplate.java.qute
+   [jbang] Template 'reactiveTestTemplate' added to '/home/blafond/.jbang/jbang-catalog.json'
+   ```
+3. Generate your unit test in the `.../usertests/tests directory` (see [Init templates](https://github.com/jbangdev/jbang#init-templates))
+    - `jbang init --template=reactiveTestTemplate ./tests/MyUnitTest`
+
+   ```
+   [jbang] File initialized. You can now run it with 'jbang ./tests/MyUnitTest' or edit it using 'jbang edit --open=[editor] ./tests/MyUnitTest' where [editor] is your editor or IDE, e.g. 'netbeans'
+   ```
+1. Open your unit test in the IDEA for [editing](https://github.com/jbangdev/jbang#editing)
```
DavideD, line 30: It doesn't matter with markdown because it will replace the numbered list with the right numbers, but this is the 4th element of the list and the next one is the 5th.

DavideD (changes requested):
Thanks, now that I gave a closer look I think this can be simplified a lot and we probably don't need the super class.
I have another couple of comments and questions:

* I don't think we should add these files to the `examples` folder. They don't represent the typical usage of Hibernate Reactive. We can create new folders in the root of the project for tooling:
  ```
  /tooling/jbang/
  ```
* Is it possible to run this with JBang without having to open it in an IDE? I would like to be able to do this: `jbang  MyUnitTest.java`. If not I think we should use the main approach instead (like in the examples). This would allow us to also read parameters. Unless there is some other way to do that. Ideally, if I've already started a PostgreSQL instance using the instructions in our [podman.md](https://github.com/hibernate/hibernate-reactive/blob/main/podman.md#postgresql) file, I would like to be able to see that everything works with `jbang MyUnitTest.java` without having to touch anything else.

DavideD:
This might not be super helpful, but if we keep a JBang file on our repository, a user could run it directly and see that it works by copying and pasting a line like this (assuming that they've already started the default database):
```
jbang https://raw.githubusercontent.com/hibernate/hibernate-reactive/49915a5fb83f707e477e2e449d4130efa9334121/tooling/jbang/Issue.java
```
Although, we would recommend to download it first or generate it instead if our template gets included in the JBang catalog.

examples/usertests/resources/reactiveTestTemplate.java.qute
blafond: > Are these placeholders for something that JBang will replace?

just placeholders. I was thinking jbang-script using *.jsh might provide arg injection/replacement but looks complicated and probably not applicable.

examples/usertests/resources/reactiveTestTemplate.java.qute
blafond: got it

examples/usertests/resources/reactiveTestTemplate.java.qute
blafond: :+1:

blafond:
> * Is it possible to run this with JBang without having to open it in an IDE? 

yes:  https://github.com/jbangdev/jbang#usage

DavideD:
> > * Is it possible to run this with JBang without having to open it in an IDE?
> 
> yes: https://github.com/jbangdev/jbang#usage

I guess this is on me for not being clear enough.
I mentioned that I would like to be able to run it like that in pretty much all places so I'm aware of the fact that you can run a `main()`.

What I mean in my previous comment is if I can run a class generated with the template in this pull request  in the same way because it doesn't have a main. Basically, can I run the test directly with JBang?

PR #755, opened 2021-05-13
[POC] New Vert.x pool integration

Sanne:
(pair programmed with @DavideD )

The first and last commits probably don't belong in the final PR, but are useful for testing and sake of discussion.

Integrating the new Pool is fundamentally trivial: we only had to remove the `ThreadLocalPoolManager` and revert to using the vert.x pool as-is.

A slight complexity arises in the need of actually closing the vert.x pool in an asynchronous safe way: the `org.hibernate.service.spi.Stoppable` services in Hibernate ORM declare a `void stop()` method; but to ensure the pool was actually closed we should wait for it.
We can't really wait for it via a `join` as that blocks the eventloop; so we introduced a way to extract the promise later.

This highlights the need that perhaps Hibernate Reactive could use a properly reactive `SessionFactory`: in this case and primarily to stop it in a reactive way, but we can see some benefits in having a reactive API to boot as well.

I'm not sure of the usefulness of this though, as all integrating platforms I know will be blocking for the component lifecycle (e.g. components in Quarkus have their lifecycle managed in old style, blocking way). So an alternative could be to introduce a simple helper like in 8c4188f651e38bd8cd61315d231f9b14175d9bbd .

I'll focus next on seeing how this pool behaves.


Sanne:
modified to use vertx `4.1.0.Beta1` - obviously CI can't deal with the snapshot

DavideD:
> modified to use vertx `4.1.0.Beta1` - obviously CI can't deal with the snapshot

[We already test on CI with the latest snapshot](https://github.com/hibernate/hibernate-reactive/pull/755/checks?check_run_id=2577534506), so there's no need to update the default when you send the PR.
It's actually better if you don't, this way we can check that it works with both the stable version and the snapshots.

Sanne:
Good point. BTW while I'm not fully done with performance tests, I'm seeing no issues with the pool.

I guess I'll convert this POC to a real PR...

Sanne:
Removed the first and last commits.

It will need a rebase, could you do that @DavideD ? There's a little conflict with your multitenancy fix.

DavideD:
Yes, no worries

DavideD:
I've rebased this here: https://github.com/hibernate/hibernate-reactive/pull/771

Issue #70, opened 2020-02-28
Add CONTRIBUTING.md to the project


DavideD:

DavideD:
We also need to add a contribute section to the website. Similar to http://hibernate.org/validator/contribute/

DavideD:
Superseeded by #2064

PR #756, opened 2021-05-13
single executable jbang-based junit test class

blafond:
targets issue  [#721](https://github.com/hibernate/hibernate-reactive/issues/721)

FIrst pass at a single junit test class that's executable by jbang CLI

### DIFF
--- a/tooling/jbang/SampleIssueTest.java
+++ b/tooling/jbang/SampleIssueTest.java
@@ -1,0 +1,177 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//DEPS org.hibernate:hibernate-core:5.4.31.Final
+//DEPS junit:junit:4.12
+//DEPS javax.persistence:javax.persistence-api:2.2
+//DEPS org.hibernate.reactive:hibernate-reactive-core:1.0.0.CR4
+//DEPS org.assertj:assertj-core:3.13.2
+//DEPS io.vertx:vertx-pg-client:4.0.3
+//DEPS io.vertx:vertx-db2-client:4.0.3
+//DEPS io.vertx:vertx-mysql-client:4.0.3
+//DEPS io.vertx:vertx-sql-client:4.0.3
+//DEPS io.vertx:vertx-unit:4.0.3
+//
+//
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.boot.registry.StandardServiceRegistry;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.provider.ReactiveServiceRegistryBuilder;
+import org.hibernate.reactive.provider.Settings;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.JUnitCore;
+import org.junit.runner.Result;
+import org.junit.runner.RunWith;
+import org.junit.runner.notification.Failure;
+
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * This test class provides a working example of setting up a simple hibernate-reactive test compatible with
+ * hibernate-reactive junit tests.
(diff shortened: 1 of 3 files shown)


tooling/jbang/SampleIssueTest.java
```diff
@@ -0,0 +1,179 @@
...
+//DEPS org.assertj:assertj-core:3.13.2
+//DEPS io.vertx:vertx-pg-client:4.0.3
+//DEPS io.vertx:vertx-db2-client:4.0.3
+//DEPS io.vertx:vertx-mysql-client:4.0.3
+//DEPS io.vertx:vertx-sql-client:4.0.3
+//DEPS io.vertx:vertx-unit:4.0.3
+//
+//
+
+package org.hibernate.reactive;
```
DavideD, line 21: If you add this package you will have a compilation error when running:
```
jbang edit --open=idea  https://raw.githubusercontent.com/hibernate/hibernate-reactive/61ae4c16e65eaaea18dfe30b3e8a0763a189f128/tooling/jbang/SampleIssueTest.java
```

DavideD: It's unecessary with JBang

DavideD:
Thanks @blafond! 

It needs some minor improvements to the way it's documented but overall it looks good to me.
An alternative would be to create something  closer to what we do for the examples.

I've created a quick POC here: https://github.com/DavideD/hibernate-reactive/blob/721-jbang-executable-test/tooling/jbang/Example.java

If you want to see them in action, here's how it work:

1. [Download JBang](https://www.jbang.dev/download)
2. Start the default Postgres we use for tests (you can replace podman with docker):
    ```
    podman run --rm --name HibernateTestingPGSQL \
    -e POSTGRES_USER=hreact -e POSTGRES_PASSWORD=hreact -e POSTGRES_DB=hreact \
    -p 5432:5432 postgres:13.2
    ```
3. JBang it:
    ```
    jbang https://github.com/DavideD/hibernate-reactive/blob/721-jbang-executable-test/tooling/jbang/Example.java
    ```
    or  
    ```
    jbang https://github.com/blafond/hibernate-reactive/blob/721-jbang-executable-test/tooling/jbang/SampleIssueTest.java
    ```
4. (Optional) If you want to edit the class with Idea
    ```
    jbang edit --open=idea https://github.com/blafond/hibernate-reactive/blob/721-jbang-executable-test/tooling/jbang/SampleIssueTest.java
    ```

Not sure which one I prefer. Maybe we can keep both? The JUnit one could be helpful to show how one can test Hibernate Reactive (maybe we can make it make it slightly more complex and add a couple of tests instead of only keep one) while the example one might be more familiar to the user and we can keep it simpler.

Note that I've asked @blafond  to not create a template for now (because it makes it slightly harder to prototype with them). I think it makes sense to convert these classes to JBang templates ones we are happy with them and we can send a pr to JBang to add them to the catalogue.

@gavinking What do you think?
@gbadner, @Sanne In case you are interested and want to see an example of JBang in action

Cheers

tooling/jbang/SampleIssueTest.java
```diff
@@ -0,0 +1,177 @@
...
+	private Mutiny.SessionFactory sessionFactory;
+
+	/**
+	 * Define the configuration parameter values for your use-case
+	 */
+	private Configuration createConfiguration() {
+		Configuration configuration = new Configuration();
+
+		// Use the correct JDBC url
+		configuration.setProperty( Settings.URL, "jdbc:postgresql://localhost:5432/hreact" );
```
DavideD, line 79: It might be nice to have the default JDBC strings for the other databases as well (commented for now ).
So that a user doesn't have to look for them if they want to test with a different db.

tooling/jbang/SampleIssueTest.java
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	public static void main(String[] args) {
+		Result result = JUnitCore.runClasses( SampleIssueTest.class);
+
+		for ( Failure failure : result.getFailures()) {
+			System.out.println(failure.toString());
+		}
+
+		System.out.println(" All unit tests succeeded:  " + result.wasSuccessful());
```
DavideD, line 175: This will always print "All unit tests succeeded: ", even if they didn't. Also there is a space at the beginning.

DavideD (changes requested):

tooling/jbang/SampleIssueTest.java
blafond: I agree... I even played with adding a DB arg for main that users could `jbang myTest.java mysql`  So we'd have a static map of URL's.  But that might open up having to add a bunch more CLI properties (pwd/user etc...) and seemed out of scope.

tooling/jbang/SampleIssueTest.java
blafond: fixed

DavideD (approved):
Thanks,
looks pretty good. 

We can play a bit more with this in the future, but I think this is a good starting point.
I will probably change a couple of minor things when working on the [contribution guide](https://github.com/hibernate/hibernate-reactive/issues/70).

In particular, I wouldn't keep a separate README for this. But we can change it later.

PR #757, opened 2021-05-14
Multitenancy without accessing ThreadPoolManager

### LINKED ISSUE #754: Multitenancy support
Hi,

according to the documentation, to add support for multitenancy, the following must be done:
1. define your own implementation of ReactiveConnectionPool or extend DefaultSqlClientPool and override getTenantPool(String tenantId).
2. a class which implements CurrentTenantIdentifierResolver
3. multitenancy strategy: database or schema

So, regarding (1), as you recommend, I tried to extend the DefaultSqlClientPool and create a pool for each tenant.
It seems that I have a big misunderstanding. I need a pool per URI, and keep them into a map with the tenant-id as the key, so the getTenantPool will return the correct pool for the tenant.
But I cannot use the createPools(URI) to create the pool per tenant because the ThreadLocalPoolManager is not public, and so I have no access to this class.

I would very appreciate some lights how to do it.

Thanks.

DavideD:
Fixes #754 

Does this make sense?
It should also be backward compatible to when we are going to remove `ThreadPoolManager`.


gavinking (approved):
LGTM

DavideD:
Added a commit to remove the` Handlers` class

DavideD:
Ci seems to have some trouble downloading dependencies. I will wait until CI pass before merging this

DavideD:
@yaakov-berkovitch Is this going to solve your issue?

yaakov-berkovitch:
@DavideD Looks good even more. Thanks. Can I use your branch or better to wait for the merge to main ?

DavideD:
I'm going to merge it tomorrow. I don't think there are going to be differences between main and my branch

DavideD:
Merged!
Thanks

gavinking:
@DavideD after pulling this I get a test failure when running the tests w/o Docker.

(They work fine when running on Docker.)

DavideD:
What kind of error? I doubt it's related

gavinking:
```
Gradle Test Executor 3 > org.hibernate.reactive.ReactiveMultitenantTest > testTenantSelection FAILED
    java.util.concurrent.CompletionException: io.vertx.pgclient.PgException: { "message": "database \"dbtenant1\" does not exist", "severity": "FATAL", "code": "3D000", "file": "postinit.c", "line": "877", "routine": "InitPostgres" }
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
        at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:607)
```

Pretty sure it's related :-P

DavideD:
Maybe something goes wrong during the creation of the additional databases?
If you set SHOW_SQL to true in the base class, you should see this:
```
INFO SchemaCreatorImpl [vert.x-worker-thread-0] HHH000476: Executing import script 'file:/home/ddalto/workspace/hibernate-reactive/hibernate-reactive-core/build/resources/test/multitenancy-test.sql'
[Hibernate] drop table if exists Pig cascade
[2021-05-18 10:10:03] WARNING io.vertx.sqlclient.impl.SocketConnectionBase            Backend notice: severity='NOTICE', code='00000', message='table "pig" does not exist, skipping', detail='null', hint='null', position='null', internalPosition='null', internalQuery='null', where='null', file='tablecmds.c', line='1217', routine='DropErrorMsgNonExistent', schema='null', table='null', column='null', dataType='null', constraint='null'
[Hibernate] create table Pig (id int4 not null, name varchar(255), version int4 not null, primary key (id))
[Hibernate] CREATE DATABASE dbtenant1
[Hibernate] CREATE DATABASE dbtenant2
[Hibernate] select current_database()
[Hibernate] select current_database()
[Hibernate] insert into Pig (name, version, id) values ($1, $2, $3)
[Hibernate] select reactivemu0_.id as id1_0_0_, reactivemu0_.name as name2_0_0_, reactivemu0_.version as version3_0_0_ from Pig reactivemu0_ where reactivemu0_.id=$1
[Hibernate] select reactivemu0_.id as id1_0_0_, reactivemu0_.name as name2_0_0_, reactivemu0_.version as version3_0_0_ from Pig reactivemu0_ where reactivemu0_.id=$1
[Hibernate] delete from Pig where id=$1 and version=$2
```

I don't see why using docker should make a difference. When you say that you don't use docker, how do you start posgres?

gavinking:
Well the user `hreact` doesn't have permissions to create a database. Because we don't grant it that permission in the README instructions.

DavideD:
Could you try this additional step?
```
ALTER USER hreact CREATEDB
```

If it works I will update the instructions in the README

gavinking:
Yes, that works. I was trying (and failing) to figure out how to do that with `grant`.

gavinking:
Also, it would probably be more elegant if the test cleaned up the database it creates when it finishes.

DavideD:
I thought about it, but in the end I didn't see any actual benefit  in adding the drop queries.

DavideD:
Anyway, I will send a PR shortly

yaakov-berkovitch:
Hi @DavideD, I started integrating with multi-tenancy, but I faced an issue.
According to the doc, the hibernate.tenant_identifier_resolver is optional, if I specify the tenant-id explicitly when opening a session. So I didn't define a tenant identifier resolver class.
My test failed anyway with the error "SessionFactory configured for multi-tenancy, but no tenant identifier specified".
So, something wrong with the doc or the code  ?

DavideD:
Mmh... no, the documentation should be correct. Maybe the error happens during schema generation? If you are creating the database with Hibernate Reactive, I don't think you can specify a schema, so without a schema resolver it doesn't know which tenant to use.

yaakov-berkovitch:
it doesn't occur during schema creation, but during tests. Schema creation is currently not part of my tests.

DavideD:
OK, I will have a look. If you could share your test, that would help

DavideD:
Nevermind. I've found the issue

yaakov-berkovitch:
Good, I just stand to send you my test :-)

yaakov-berkovitch:
Hi @DavideD, did you already push a fix for the tenant identifier resolver ?

PR #760, opened 2021-05-17
Avoid lookup of Event Listener Group(s)

Sanne:


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/760

Thanks!

PR #761, opened 2021-05-18
Drop extra Databases and update README 

DavideD:
A couple of missing bits for #754

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ following commands:
     create database hreact;
     create user hreact with password 'hreact';
     grant all privileges on database hreact to hreact;
+    alter user hreact createdb;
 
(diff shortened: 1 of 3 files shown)


DavideD:
I don't think one needs additional permission to drop the database if they've created it.

gavinking:
Great, thanks :-)

> I don't think one needs additional permission to drop the database if they've created it.

Nope, it seems not.

gavinking:
Oops, what did I do?!

It didn't pass the CI build, and I didn't notice.

I *did* test it locally.

gavinking:
Ugh, it broke the tests on Cockroach. I should definitely not have merged it. Sorry.

DavideD:
Yeah, I noticed. No problem

PR #765, opened 2021-05-18
Multitenancy without CurrentTenantIdentifierResolver

### LINKED ISSUE #764: Multitenancy fails without CurrentTenantIdentifierResolver

DavideD:
Fixes #764 

@gavinking shouldn't also be possible for a user to set the tenant for the stateless session?


gavinking:
Huh, I don't get it: why to `SessionFactory`s have tenant ids? That doesn't seem to make sense.

> shouldn't also be possible for a user to set the tenant for the stateless session?

Sure, it's easy to add.

DavideD:
I think it's because ORM uses the following approach for the creation of the session with a tenant:
```
ormFactory.withOptions().tenantIdentifier( "tenantId" ).openSession()
```

gavinking:
Oh, I see, Ok, fine.

PR #767, opened 2021-05-18
[POC] New strategy to fire events

Sanne:
also requires changes in ORM core: https://github.com/hibernate/hibernate-orm/commit/17ef565fa6db5a5045bfdbd64598ba47a314c521

This gets Hibernate Reactive to use a similar optimisation as we recently did for ORM core, which is to delegate the details of iteration to the implementation of the `EventListenerGroup`.

This reduces the runtime overhead significantly; as it:
- needs no lookup of the right ListenerGroups by type
- doesn't need enumerating / iterating on each listener list
- doesn't need defensive copies of such iterators, yet encapsulates safely
- allows to skip creating the Event object for empty listener groups (useful for some groups only)
- less code :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ version = projectVersion
 // ./gradlew clean build -PhibernateOrmVersion=5.4.32-SNAPSHOT
 ext {
     if ( !project.hasProperty('hibernateOrmVersion') ) {
-        hibernateOrmVersion = '5.4.31.Final'
+        hibernateOrmVersion = '5.5.0-SNAPSHOT'
     }
     // For ORM, we need a parsed version (to get the family, ...)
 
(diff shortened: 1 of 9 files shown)


gavinking:
looks good

Sanne:
thanks :)

Just spotted by profiling this.. there's another expensive iteration in `ReactivePlanEntityLoader#reactivePerformTwoPhaseLoad` .. I can't work on that now but let's patch it somehow as well.

DavideD:
LGTM

Are these changes going to be backported to ORM 5.4? Mostly, I'm wondering what's our plans in term of Hibernate Reactive dependencies

Sanne:
Yes we can backport to 5.4 - totaly harmless to ORM users.

DavideD:
Changes here have been applied via:
* https://github.com/hibernate/hibernate-reactive/pull/793
* https://github.com/hibernate/hibernate-reactive/pull/786

PR #770, opened 2021-05-19
call prepareCollectionFlushes() from thenAccept()

gavinking:
This change might not be necessary but it at least *looks* cleaner.

WDYT, @DavideD?


DavideD:
Merged, thanks!

DavideD:
Good catch

PR #778, opened 2021-05-21
Infinite recursion calls

### LINKED ISSUE #775: Infinite recursion calls in ReactiveBatchingEntityLoader#load
See https://github.com/hibernate/hibernate-reactive/blob/c91dd01e7a1d779c270b08f411606c2d5f1dd3f6/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/impl/ReactiveBatchingEntityLoader.java#L57

DavideD:
Fixes #775

Any idea how to test this?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/impl/ReactiveBatchingEntityLoader.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/impl/ReactiveBatchingEntityLoader.java
@@ -54,16 +54,6 @@ public abstract class ReactiveBatchingEntityLoader implements ReactiveUniqueEnti
 	}
 
 	@Override
-	public CompletionStage<Object> load(
-			Serializable id,
-			Object optionalObject,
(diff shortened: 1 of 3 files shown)


gavinking:
Well, where precisely is the infinte recursion? I don't see it.

There was none at `ReactiveBatchingEntityLoader.load()` because that method is overridden on both concrete subclasses. (Nevertheless it's better to delete it.)

The motivation for the other change I'm not seeing, so that must be causing the recursion, though it's not obvious how.

DavideD:
> Well, where precisely is the infinte recursion? I don't see it.

I should have said **potential** Infinite recursion.

As you said, all the classes override the method so we never experience it.

The other change is unrelated, it removes a deprecated method call.

DavideD:
> I should have said potential Infinite recursion.

Well... unless you consider causing a stackoverflow exception a good default implementation :-)

DavideD:
Merged, thanks

gavinking:
> Well... unless you consider causing a stackoverflow exception a good default implementation :-)

It's not ideal :-)

PR #779, opened 2021-05-22
bind the Session object to the current stream

gavinking:
Proof of concept. cc @Sanne @FroMage

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
@@ -125,41 +125,82 @@ public class StageSessionFactoryImpl implements Stage.SessionFactory {
 	}
 	@Override
 	public <T> CompletionStage<T> withSession(Function<Stage.Session, CompletionStage<T>> work) {
-		Stage.Session current = Vertx.currentContext().getLocal( Stage.Session.class.getName() );
-		if ( current!=null ) {
+		String id = sessionId();
+		Stage.Session current = Vertx.currentContext().getLocal(id);
+		if ( current!=null && current.isOpen() ) {
 			return work.apply( current );
 		}
 		return newSession().thenCompose(
-				session -> work.apply(session).whenComplete( (r, e) -> session.close() )
+				session -> {
+					Vertx.currentContext().putLocal(id, session);
+					return work.apply(session).whenComplete( (r, e) -> {
+						Vertx.currentContext().removeLocal(id);
+						session.close();
+					} );
+				}
 		);
 	}
 
 	@Override
 	public <T> CompletionStage<T> withSession(String tenantId, Function<Stage.Session, CompletionStage<T>> work) {
+		String id = sessionId(tenantId);
+		Stage.Session current = Vertx.currentContext().getLocal(id);
+		if ( current!=null && current.isOpen() ) {
+			return work.apply( current );
+		}
 		return newSession( tenantId ).thenCompose(
-				session -> work.apply(session).whenComplete( (r, e) -> session.close() )
+				session -> {
+					Vertx.currentContext().putLocal(id, session);
+					return work.apply(session).whenComplete( (r, e) -> {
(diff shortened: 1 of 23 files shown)


gavinking:
@vietj what is the performance implication of `Context.putLocal()`?

gavinking:
Upon reflection, it might be better to expose this as `sessionFactory.currentSession().currentTransaction()`.

gavinking:
> Upon reflection, it might be better to expose this as `sessionFactory.currentSession().currentTransaction()`.

I've made it `Mutiny.currentSession().currentTramsaction()` now.

That isn't quite right if you have multiple session factories, but on the other hand it's *much* more convenient in the common case where you have just one.

gavinking:
I still need to make this tolerate the case where there is no Vert.x context.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -133,10 +134,12 @@ private ReactiveConnection proxyConnection(String tenantId) {
...
-					return work.apply(session).whenComplete( (r, e) -> {
-						Vertx.currentContext().removeLocal(id);
-						session.close();
-					} );
+					return work.apply(session)
+							.handle(this::handler)
+							.thenCompose( handler -> {
+						        Vertx.currentContext().removeLocal(id);
+								return session.close().thenApply(handler);
+							} );
```
gavinking, line 142: @DavideD @Sanne note that this fixes a really bad bug on main where `Stage` sessions opened using `withSession()` _still_ weren't getting closed after the migration to Vert.x 4.

gavinking:
> I still need to make this tolerate the case where there is no Vert.x context.

I'm trying to decide what to do here: I think the best thing might be to introduce an extension point that lets you plug in a `Context` object.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
Sanne: sorry unable to review it all right now, but if there's some urgent bugfix in here could you extract it for immediate integration?

We're very close to deadlines for Quarkus 2 : we'll need another tag of HR tomorrow - as soon as I manage to get an ORM tag released to match it.

gavinking: Well, TBH I'm pretty happy with this work now, so if I add a couple more tests and update the docs, I think it would be fine to merge it.

Sanne: Ok!

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -484,6 +484,18 @@ public void testMetamodel(TestContext context) {
 		context.assertEquals( "GuineaPig", pig.getName() );
 	}
 
+	@Test
+	public void reactiveThreadBoundTransaction(TestContext context) {
+		test( context, getMutinySessionFactory().withTransaction(
+				(session, transaction) -> session.createQuery("from GuineaPig").getResultList().invoke( list -> {
+					context.assertNotNull( session.currentTransaction() );
```
DavideD, line 491: Shouldn't be `assertTrue( session.currentTransaction() == transaction)`?

DavideD (approved):

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
gavinking: Well, I don't think we care if the actual object is identically equal, only that it represents the same _state_.

DavideD: Ah right, makes sense. Thanks

gavinking:
Now #787.

PR #781, opened 2021-05-22
[#780] Fix driver discovery issue

### LINKED ISSUE #780: Wrong driver selected when there is more than one dependency on the class path
With both `io.vertx:vertx-mysql-client` and `io.vertx:vertx-db2-client` on the classpath and set url: `jdbc:mysql://localhost:3306/hreact?user=hreact&password=hreact&serverTimezone=UTC`

Hibernate Reactive will select the DB2 driver instead of the MYSQL one.

The error might not occurs if the driver are discovered in the right order

DavideD:
Fixes #780 

If there are multiple drivers in the classpath,
we sometimes select the wrong one.

We didn't notice before because it might not happen
if the drivers are discovered in the right order.


DavideD:
I prefer to make some changes

PR #788, opened 2021-05-24
Avoid conversion of Arrays to streams when looping

DavideD:
@Sanne are these changes going to be helpful?

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
@@ -14,7 +14,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.IntStream;
 import javax.persistence.metamodel.Attribute;
 
 import org.hibernate.AssertionFailure;
@@ -501,16 +500,15 @@ public interface ReactiveAbstractEntityPersister extends ReactiveEntityPersister
 		}
 
 		Object[] state = loadedState;
-		return loop(
-				IntStream.iterate(span-1, (i) -> i-1 ).limit( span ),
-				table -> deleteReactive(
-						id,
-						version,
-						table,
-						deleteStrings[table],
-						session,
-						state
-				)
+		return loop( 0, span,
+					 table -> deleteReactive(
+							 id,
+							 version,
+							 span - table - 1,
+							 deleteStrings[span - table - 1],
+							 session,
+							 state
+					 )
 		);
 	}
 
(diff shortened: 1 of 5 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -501,16 +500,15 @@ else if ( isAllOrDirtyOptimisticLocking() && loadedState != null ) {
...
-				)
+		return loop( 0, span,
+					 table -> deleteReactive(
+							 id,
+							 version,
+							 span - table - 1,
+							 deleteStrings[span - table - 1],
+							 session,
+							 state
+					 )
```
gavinking, line 511: But doesn't this reverse the order of the iteration?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -728,19 +726,20 @@ else if ( !delegate().isModifiableEntity( entry ) ) {
...
+								fields,
+								oldFields,
+								table == 0 ? rowId : null,
+								propsToUpdate,
+								table,
+								oldVersion,
+								updateStrings[table],
+								session
+						)
+						: voidFuture()
```
gavinking, line 742: I'm not convinced by this one. I'm not sure what implication passing all those extra `voidFuture()`s has. Perhaps it's fine, but I would need to dig a lot deeper into the code.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
```diff
@@ -143,8 +142,12 @@ public static void logSqlException(Throwable t, Supplier<String> message, String
...
 	 * </pre>
 	 */
-	public static <T> CompletionStage<Void> loop(T[] array, Function<T,CompletionStage<?>> consumer) {
-		return loop( Arrays.stream(array), consumer );
+	public static <T> CompletionStage<Void> loop(T[] array, Function<T, CompletionStage<?>> consumer) {
+		AtomicInteger index = new AtomicInteger();
+		return AsyncTrampoline.asyncWhile( () -> {
+			int i = index.getAndIncrement();
+			return consumer.apply( array[i] ).thenApply( v -> i < array.length - 1 );
+		} );
```
gavinking, line 150: Are we really certain that creating an `AtomicInteger` is less expensive than creating a stream? I mean, it certainly sounds _possible_, but it's not _obviously_ true.

gavinking (commented):
I mean, I'm not convinced on first sight, but I guess if you really dug into this??

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
gavinking: Remember: "creating a stream" in this case just means making a little wrapper for an array.

DavideD:
We can wait about merging this one and do proper performance testing.
There are other things that I want to try. 

I'm going to close the pr for now

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
DavideD: That was what the original code was doing

gavinking: Oh, apologies, I missed that you had changed it to use `span - table - 1`. That's better, actually.

PR #794, opened 2021-05-24
Add comment about updating services in Quarkus

DavideD:
A small reminder to check Quarkus when we add new services


gavinking:
Good idea.

Issue #766, opened 2021-05-18
Support multitenancy for the stateless session


DavideD:
Currently, it's not possible to specify the tenant for the stateless session (not sure about using a `CurrentTenantIdentifierResolver`).

DavideD:
@blafond Would you take this one, please?

PR #797, opened 2021-05-24
draft - support multitenancy for stateless session

blafond:
Fixes [#766](https://github.com/hibernate/hibernate-reactive/issues/766)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -171,6 +185,23 @@ private ReactiveConnection proxyConnection(String tenantId) {
 				} );
 	}
 
+	@Override
+	public <T> Uni<T> withStatelessSession(String tenantId, Function<Mutiny.StatelessSession, Uni<T>> work) {
+		String id = statelessSessionId();
```
gavinking, line 190: @blafond this would need to be an id that includes the tenant id. At least that's how I did it with regular sessions.

gavinking (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
blafond: ah.. I missed that. thx

blafond: @DavideD Note locally this [test](https://github.com/hibernate/hibernate-reactive/blob/da7c79cf87ca88942b53bcdd15040c9253eb89c3/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveMultitenantNoResolverTest.java#L304) is failing

Though a HibernateException is getting thrown as expected.  The Matcher isn't matching because the expected exception isn't wrapped in a CompletionException like in the previous test in that class.  Can't figure out why though, but guessing it's in my `withStatelessSession(String tenantId, ....)` logic which I mimic'd from the non-tenantId version.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveMultitenantNoResolverTest.java
```diff
@@ -227,6 +292,22 @@ public void testWithTransactionThrowsExceptionWithoutTenantWithMutiny(TestContex
...
+	public void testWithStatelessSessionThrowsExceptionWithoutTenant(TestContext context) {
+		thrown.expectCause( isA( HibernateException.class ) );
+		thrown.expectMessage( "no tenant identifier" );
+
+		test( context, getSessionFactory().withStatelessSession( this::selectCurrentDB ) );
+	}
+
+	@Test
+	public void testWithStatelessSessionThrowsExceptionWithoutTenantWithMutiny(TestContext context) {
+		thrown.expectCause( isA( CompletionException.class ) );
```
DavideD, line 305: This assertion is wrong for a couple of different reasons:
1. You are using the `Mutiny.SessionFactory` so there is not going to be a `CompletionException` anywhere (hopefully).
2. If you were using the `Stage.SessionFactory`, the error iwould be a `CompletionStageException` but the cause is going to be a `HibernateException` (like in `testWithStatelessSessionThrowsExceptionWithoutTenant`). Check the difference bewtween `expect` and `expectCause`.

You need to change this to:
```
	@Test
	public void testWithStatelessSessionThrowsExceptionWithoutTenantWithMutiny(TestContext context) {
		thrown.expect( isA( HibernateException.class ) );
		thrown.expectMessage( "no tenant identifier" );

		test( context, getMutinySessionFactory().withStatelessSession( this::selectCurrentDB ) );
	}
```
```

DavideD (changes requested):
Thanks!

DavideD:
@blafond you need to rebase to the latest changes in main, or the build on CI will fail

DavideD (changes requested):
Thanks!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
gavinking: @blafond if this is testing the _mutiny_ API then I don’t see why a `ConpletionException` would be correct.

DavideD:
This PR is obsolete now. @blafond is working on the new one

PR #805, opened 2021-05-26
Testing: fix calls to session.close() + remove warnings

DavideD:
Several places in the tests where calls to session.close weren't chained correctly


gavinking (approved):
yeah, good catch

Issue #465, opened 2020-12-02
exceptions get wrapped in CompletionException


gavinking:
I don't like this `CompletionException` thing at all, but I guess there's no way to avoid it when using `CompletionStage`s.

But I guess they're also leaking into our Mutiny API. We should try to prevent that.

gavinking:
Digging through our stack traces is actually pretty painful. We need to figure out how to improve this.

gavinking:
I investigated this and there doesn't seem to be a problem (anymore?).

PR #810, opened 2021-05-27
test the exception type after a UK violation

gavinking:
See #465.

Apparently Mutiny *does* filter out the `CompletionException`s, which is *great*.

Is that right, @cescoffier?

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
@@ -10,6 +10,7 @@ import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.persistence.Entity;
 import javax.persistence.Id;
+import javax.persistence.PersistenceException;
 import javax.persistence.Table;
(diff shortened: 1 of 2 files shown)


DavideD:
I think so, we test already exceptions for Mutiny and CompletionStage in different places and that's what happens

gavinking:
> I think so, we test already exceptions for Mutiny and CompletionStage in different places and that's what happens

Yeah, but we didn't have any tests for a proper `PersistenceException`. At least not AFAICS.

DavideD (approved):
Looks good but don't we already have this test?

Mmh...probably not with `withTransaction`

gavinking:
Oh, we did have one in `StageExceptionsTest` and `MutinyExceptionsTest`. But they don't actually test the thing I was most interested in.

gavinking:
Anyway, might as well push these because in principle they are slightly different (they use `withTransaction()` so there are extra levels of wrapping).

DavideD:
Yes, not complaining :-D

cescoffier:
> Apparently Mutiny does filter out the CompletionExceptions, which is great.

Yes, we unwrap the "cause" when we see these.

gavinking:
Great, perfect, thanks.

PR #813, opened 2021-05-27
Close session if an exception happens in withSession

### LINKED ISSUE #811: withSession doesn't close the session in case of exception
@gavinking shouldn't we close the session if an exception occurs?

Example:
```
	@Test
	public void testExceptionInWithSession(TestContext context) {
		final Stage.Session[] savedSession = new Stage.Session[1];
		test( context, getSessionFactory().withSession( session -> {
			context.assertTrue( session.isOpen() );
			savedSession[0] = session;
			throw new RuntimeException( "No Panic: This is just a test" );
		} ).handle( (o, t) -> context.assertFalse( savedSession[0].isOpen(), "Session should be closed" ) ) );
	}
```

DavideD:
Fixes #811

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -73,14 +73,31 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 	Uni<Mutiny.Session> newSession() throws HibernateException {
 		SessionCreationOptions options = options();
 		return uni( () -> connection( options.getTenantIdentifier() ) )
-				.map( reactiveConnection -> new ReactiveSessionImpl( delegate, options, reactiveConnection ) )
+				.chain( reactiveConnection -> create( reactiveConnection, () -> new ReactiveSessionImpl( delegate, options, reactiveConnection ) ) )
 				.map( s -> new MutinySessionImpl(s, this) );
 	}
 
 	Uni<Mutiny.Session> newSession(String tenantId) throws HibernateException {
 		return uni( () -> connection( tenantId ) )
-				.map( reactiveConnection -> new ReactiveSessionImpl( delegate, options( tenantId ), reactiveConnection ) )
+				.chain( reactiveConnection -> create( reactiveConnection, () -> new ReactiveSessionImpl( delegate, options( tenantId ), reactiveConnection ) ) )
 				.map( s -> new MutinySessionImpl(s, this) );
+	}
+
+	/**
+	 * Close the connection if something goes wrong during the creation of the session
+	 */
+	private <S> Uni<S> create(ReactiveConnection connection, Supplier<S> supplier) {
+		try {
+			return Uni.createFrom().item( supplier.get() );
+		}
+		catch (Throwable t) {
+			return Uni.createFrom()
+					.completionStage( connection.close() )
+					// There has been an error already,
+					// if another happens while closing the connection we will ignore it
+					.onItemOrFailure()
+					.transformToUni( (unused, throwable) -> Uni.createFrom().failure( t ) );
+		}
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -5,7 +5,12 @@
  */
 package org.hibernate.reactive.mutiny.impl;
 
-import io.smallrye.mutiny.Uni;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.metamodel.Metamodel;
 
 import org.hibernate.Cache;
 import org.hibernate.HibernateException;
@@ -19,12 +24,7 @@ import org.hibernate.reactive.session.impl.ReactiveCriteriaBuilderImpl;
 import org.hibernate.reactive.session.impl.ReactiveSessionImpl;
 import org.hibernate.reactive.session.impl.ReactiveStatelessSessionImpl;
 
-import javax.persistence.criteria.CriteriaBuilder;
-import javax.persistence.metamodel.Metamodel;
-import java.util.concurrent.CompletionStage;
-import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.function.Supplier;
+import io.smallrye.mutiny.Uni;
 
 import static org.hibernate.reactive.common.InternalStateAssertions.assertUseOnEventLoop;
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -141,19 +141,11 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
 	@Override
 	public <T> Uni<T> withSession(Function<Mutiny.Session, Uni<T>> work) {
-		Mutiny.Session current = context.get(Mutiny.Session.class, sessionId);
-		if ( current!=null && current.isOpen() ) {
+		Mutiny.Session current = context.get( Mutiny.Session.class, sessionId );
+		if ( current != null && current.isOpen() ) {
 			return work.apply( current );
 		}
-		return newSession()
-				.chain( session -> {
-					context.put(Mutiny.Session.class, sessionId, session);
-					return work.apply( session )
-							.eventually( () -> {
-								context.remove(Mutiny.Session.class, sessionId);
-								return session.close();
-							} );
-				} );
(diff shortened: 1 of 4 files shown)


gavinking:
I don't understand what this is doing. Why do we care about exceptions that occur while creating the session? (If we fail to create the session, there's no need to close it.) But that doesn't seem to be the case you were trying to fix, so I just don't understand the fix at all.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -78,23 +80,42 @@ public StageSessionFactoryImpl(SessionFactoryImpl delegate) {
...
 	CompletionStage<Stage.StatelessSession> newStatelessSession() {
 		SessionCreationOptions options = options();
 		return stage( v -> connection( options.getTenantIdentifier() )
-				.thenApply( connection -> new ReactiveStatelessSessionImpl( delegate, options, connection ) )
+				.thenCompose( connection -> create( connection, () -> new ReactiveStatelessSessionImpl( delegate, options, connection ) ) )
 				.thenApply( s -> new StageStatelessSessionImpl(s, this) ) );
 	}
 
+	/**
+	 * Close the connection if something goes wrong during the creation of the session
```
gavinking, line 101: What kind of things could go wrong?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -146,30 +155,46 @@ private ReactiveConnection proxyConnection(String tenantId) {
...
+							return closeSession.apply( session )
+									.onItemOrFailure()
+									.transformToUni( (unused, throwable) ->  Uni.createFrom().failure( e ) );
+						}
+						catch (Exception removeEx ) {
+							return closeSession.apply( session )
+									.onItemOrFailure()
+									.transformToUni( (unused, throwable) ->  Uni.createFrom().failure( e ) );
+						}
+					}
```
gavinking, line 197: Oh so the issue was that even though `work` looks like a `Uni`, it's actually a function that can throw its own exception, that is *not* captured in the stream. That much makes perfect sense. But there seems to be other stuff going on here.

gavinking:
Wouldn't a *much* simpler fix just be to write:

```java
voidFuture().thenApply(v -> work.apply(session))
```

and that way the exception from the `work` function gets naturally captured into the stream.

I dunno everything you're thinking here, but the fix naively looks way to complicated.

gavinking:
I just tried using `voidFuture().thenCompose(v -> work.apply(session))` and your new tests seem to pass.

That's a *much* smaller change.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
DavideD: Multitenancy.
Check the constructor for `AbstractSharedSessionContract`. 

If we don't close the we are going to see the exception but the connection will never get closed, leading to the app getting stuck or throwing  TimeOut exceptions that are not easy to find.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
DavideD: I was thinking that `context.remove` might throw an exception. Mostly because it is a service and might reimplemented.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
DavideD: I mean, Multitenancy is the easier to see, but there are all sort of initialisation when we create the session. So, there might be something else

DavideD: Probably, a better solution would be to wait to get the connection until we are ready with the session.

DavideD:
Obviously... I didn't think about it :-)

DavideD:
I will update the PR

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
gavinking: > Probably, a better solution would be to wait to get the connection until we are ready with the session.

Hrm so you're worried about closing the connection because in `withSession()` we don't use a `ProxyConnection`. I see.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
gavinking: I think we could reasonably require that `remove()` not throw exceptions. It's just a cleanup method after all.

gavinking: My philosophy on this would be: a problem that occurs cleaning up resources _after_ a txn has already completed successfully _might_ be something that is of interest to a system administrator (so should be logged) but is almost certainly _not_ something that should be reported to the user. What the user needs to see is that their txn was successful.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -73,16 +73,33 @@ public MutinySessionFactoryImpl(SessionFactoryImpl delegate) {
...
+				.chain( reactiveConnection -> create( reactiveConnection, () -> new ReactiveSessionImpl( delegate, options( tenantId ), reactiveConnection ) ) )
 				.map( s -> new MutinySessionImpl(s, this) );
 	}
 
+	/**
+	 * Close the connection if something goes wrong during the creation of the session
+	 */
+	private <S> Uni<S> create(ReactiveConnection connection, Supplier<S> supplier) {
+		try {
+			return Uni.createFrom().item( supplier.get() );
```
gavinking, line 91: @DavideD this looks weird to me. Did you mean to write `Uni.createFrom().item(supplier)`?

gavinking: Oh no, you didn't, because you want to force the error to happen immediately.

But then why wrap it in a `Uni`?

gavinking: Oh I see the point is to convert a `CompletionStage` to a `Uni`. But is that even necessary here?

Hard to make sense of this change from the diff.

gavinking: @DavideD why didn't you just use `onFailure()` to catch the exception and close the session. Wouldn't make the code a little easier to understand?

DavideD:
Ok, it seems that I didn't understand how the method `handler` work, now it makes more sense.

But I have one question, aren't we going to loose the real exception it something happens when we close the session?

I'm talking about [this code](https://github.com/hibernate/hibernate-reactive/blob/2bb8edc626c5302f5e222f2620b580084bb29d17/hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java#L127):
```
return session.close().thenApply(handler);
```

if `session.close` fails for whatever reason, `thenApply` doesn't get called and if an exception happened before this point (in the `handler`) we are going to lose it. Or am I missing something?

gavinking:
> if `session.close` fails for whatever reason, `thenApply` doesn't get called and if an exception happened before this point (in the handler) we are going to lose it. Or am I missing something?

As I posted above:

> a problem that occurs cleaning up resources after a txn has already completed successfully might be something that is of interest to a system administrator (so should be logged) but is almost certainly not something that should be reported to the user. What the user needs to see is that their txn was successful.

gavinking:
Oh, I guess I misread what you're saying: you're saying that the exception from `session.close()` suppresses the one from the handler.

OK, sure, so probably that should be done in `handle()` instead of in a `thenApply()`.

(Really `session.close()` should not throw exceptions, but it probably does, 'cos 1991.)

gavinking:
Notice that the API of `CompletionStage` is _utterly garbage_ as far as allowing for clean exception handling. Mutiny is _soooo_ much better from this respect.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
DavideD: Sorry, I'm still working a bit around it. I've already applied your suggestions.

DavideD:
Closing this pr because it's already changed quite a bit, I will reopen it when I'm ready

gavinking:
Ok, sorry, take your time, didn't mean to be a PITA.

DavideD:
No, no, you are not a PITA (this time at least :-P). I just don't want to waste your time reviewing something that's already changed.
Your comments helped me understand a couple of things that I didn't realize before. Thanks.

PR #814, opened 2021-05-27
Close session in case of exception in 

### LINKED ISSUE #811: withSession doesn't close the session in case of exception

DavideD:
Follows #813 
Fixes #811 

Hopefully, this will look a bit better.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -103,7 +103,7 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 	Uni<Mutiny.StatelessSession> newStatelessSession() throws HibernateException {
 		SessionCreationOptions options = options();
 		return uni( () -> connection( options.getTenantIdentifier() ) )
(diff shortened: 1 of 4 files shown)


gavinking (approved):
This looks fine to me now, thanks.

The only thing I'm thinking is that we should add back some sort of `Closeable` interface, like we had before, in order to avoid having to pass around method refs.

But I can do that as a separate task if you prefer.

gavinking:
Wait, @DavideD I think you missed a similar change to `newStatelessSession()` for Mutiny.

DavideD:
> The only thing I'm thinking is that we should add back some sort of `Closeable` interface, like we had before, in order to avoid having to pass around method refs.

Definitely, I was going to propose it :-)
Yes, please,  have a look at it as a separate issue.

DavideD:
> Wait, @DavideD I think you missed a similar change to `newStatelessSession()` for Mutiny.

Good catch, I've probably lost the change during a rebase

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/814

Thanks

PR #819, opened 2021-05-29
Clarifications for the contextual Sessions

Sanne:
I've played with some very minor performance improvements for this, and renamed the variables + added some comments as I found it confusing.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
@@ -24,9 +24,13 @@ public class VertxContext implements Context, ServiceRegistryAwareService {
         return vertxInstance.getVertx().getOrCreateContext();
     }
 
+    private io.vertx.core.Context currentContext() {
+        return Vertx.currentContext();
+    }
+
     @Override
     public void injectServices(ServiceRegistryImplementor serviceRegistry) {
-        vertxInstance = serviceRegistry.getService(VertxInstance.class);
+        vertxInstance = serviceRegistry.getService( VertxInstance.class );
     }
 
     @Override
@@ -36,12 +40,21 @@ public class VertxContext implements Context, ServiceRegistryAwareService {
 
     @Override
     public <T> T get(Class<T> key, String id) {
-        return getOrCreateContext().getLocal( id );
+        final io.vertx.core.Context context = currentContext();
+        if ( context != null ) {
+            return context.getLocal( id );
+        }
+        else {
+            return null;
+        }
     }
 
     @Override
     public void remove(Class<?> key, String id) {
-        getOrCreateContext().removeLocal( id );
+        final io.vertx.core.Context context = currentContext();
+        if ( context != null ) {
+            context.removeLocal( id );
+        }
     }
 
     @Override
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
@@ -5,6 +5,7 @@
  */
 package org.hibernate.reactive.mutiny.impl;
 
+import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 import java.util.function.BiFunction;
 import java.util.function.Function;
@@ -49,6 +50,7 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 	private final String contextKeyForStatelessSession;
 
 	public MutinySessionFactoryImpl(SessionFactoryImpl delegate) {
+		Objects.requireNonNull( delegate );
 		this.delegate = delegate;
 		context = delegate.getServiceRegistry().getService( Context.class );
 		connectionPool = delegate.getServiceRegistry().getService( ReactiveConnectionPool.class );
@@ -71,6 +73,7 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
 	@Override
 	public Mutiny.Session openSession(String tenantId) {
+		Objects.requireNonNull( tenantId, "parameter 'tenantId' is required" );
 		return new MutinySessionImpl(
 				new ReactiveSessionImpl( delegate, options( tenantId ), proxyConnection( tenantId ) ),
 				this
@@ -139,6 +142,7 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
 	@Override
 	public <T> Uni<T> withSession(Function<Mutiny.Session, Uni<T>> work) {
+		Objects.requireNonNull( work, "parameter 'work' is required" );
 		Mutiny.Session current = context.get( Mutiny.Session.class, contextKeyForSession );
 		if ( current != null && current.isOpen() ) {
 			return work.apply( current );
@@ -148,6 +152,8 @@ public class MutinySessionFactoryImpl implements Mutiny.SessionFactory {
 
 	@Override
 	public <T> Uni<T> withSession(String tenantId, Function<Mutiny.Session, Uni<T>> work) {
+		Objects.requireNonNull( tenantId, "parameter 'tenantId' is required" );
+		Objects.requireNonNull( work, "parameter 'work' is required" );
 		String id = contextKeyForSession + '/' + tenantId;
 		Mutiny.Session current = context.get(Mutiny.Session.class, id);
 		if ( current!=null && current.isOpen() ) {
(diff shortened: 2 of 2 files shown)


Sanne:
resolved conflicts

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -16,17 +16,21 @@
  *
  * @author Gavin King
  */
-public class VertxContext implements Context, ServiceRegistryAwareService {
+public final class VertxContext implements Context, ServiceRegistryAwareService {
```
gavinking, line 19: Do we really want to stop people subclassing this? I would have thought that was useful.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -36,12 +40,21 @@ public void injectServices(ServiceRegistryImplementor serviceRegistry) {
...
     @Override
     public <T> T get(Class<T> key, String id) {
-        return getOrCreateContext().getLocal( id );
+        final io.vertx.core.Context context = currentContext();
+        if ( context != null ) {
+            return context.getLocal( id );
+        }
+        else {
+            return null;
+        }
```
gavinking, line 49: Are you sure about this? Isn't it useful to create a context for them if there is none active?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -36,12 +40,21 @@ public void injectServices(ServiceRegistryImplementor serviceRegistry) {
...
     }
 
     @Override
     public void remove(Class<?> key, String id) {
-        getOrCreateContext().removeLocal( id );
+        final io.vertx.core.Context context = currentContext();
+        if ( context != null ) {
+            context.removeLocal( id );
+        }
```
gavinking, line 57: +1

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
```diff
@@ -151,33 +164,36 @@ private ReactiveConnection proxyConnection(String tenantId) {
...
 
 	private<S extends Mutiny.Closeable, T> Uni<T> withSession(
-			Class<S> sessionType,
-			Uni<S> sessionUni,
-			Function<S, Uni<T>> work,
-			String id) {
+			final Class<S> sessionType,
+			final Uni<S> sessionUni,
+			final Function<S, Uni<T>> work,
+			final String contextId) {
```
gavinking, line 179: Why id this necessary? Surely these are marked implicitly-final by the compiler.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
Sanne: When it's a fairly simple `Service` implementation, I'd encourage people to implement their own from scratch by implementing the base interface (`Context` in this case).

This is also more flexible as you might not want to make it `ServiceRegistryAwareService` for example, which is implied by this particular implementation.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
Sanne: Well not creating a Context when one isn't required is precisely the point of my performance fix.

Besides, with a method called like this I wouldn't expect to use it to create a context if there isn't one? Would people want to rely on such side-effects?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
Sanne: I agree, they are most likely unnecessary. Personally I prefer such things to be explicit - matter of style and taste :)

Can revert if you don't like it, no strong opinion.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
gavinking: Well it seemed to me pretty likely that you might just want to override `execute()` without overriding any of the other methods.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
gavinking: But then this is a change to functionality that we had already discussed and agreed on a while ago. We had said that we would do this for you if you called the `Session` outside of a Vert.x context.

I mean, I'm perfectly happy to re-discuss it...

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
gavinking: Well we don't do that anywhere else in the codebase: we rely on implicit final _all over the place_. It's weird to not trust it in just one method.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
DavideD: As a user, I find it pretty annoying having to re-implement a whole class when I want to change only one method of the existing one.

DavideD: If anything, let's clarify that the recommended approach is to implement the interface but let's not make life harder for developers that know what they are doing.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
DavideD: @gavinking But why do we need to create the context here? Isn't it created anyway in the `execute`?

gavinking: This is called from `withSession()`.

gavinking: Well, actually perhaps there's room for some cleanup here: `newSession()` _does_ call `execute()` but that's just slightly too late. We should look a little more carefully at that.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
Sanne: well fine I'll remove the `final` ... but this is an `impl` class, it seems intended for people to back off. And I just don't expect many users to have any need to change how this works.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
gavinking: Anyway, if the idea is that calling `getOrCreateContext()` is a performance bug, then we should be getting rid of it from all the `Session` methods that call it (which is .... _all of them_). Getting rid of it from `Context.get()` is pointless, since that's called only once per session.

Sanne: I'm not arguing about where it's used: if opening a Session needs to create a context, it can of course do so.

 Just in terms of API contract of `Context#get` it seems not its responsibility to create a Context automagically - and we can save some cycles in doing so.

DavideD: We might also want to update the javadoc of `Context` or `VertxContext`. it seems we don't clarify when we expect  a new context to be created.

Sanne:
removed all newly introduced `final` ..

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
gavinking: Oh, wait, @Sanne, I think I understand what you're saying now: you're saying it doesn't need to be created in `get()` because it will be created later in `put()`. 

OK, sure, fine. Indeed, since `newSession()` calls `execute()`, I guess we don't even need to call `getOrCreateContext()` from `put()`, since it will have already been called from `execute()`.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionFactoryImpl.java
Sanne: Removed it.

But I generally never trust it and when I code I prefer it explicit . Especially when we have many lambdas, there's a high degree of risk of unintentionally creating a performance problem as it's very easy to miss cases in which the code is no longer eligible for "effectively final" treatment - and when it's common practice to not use them it's kinda guaranteed we'll have issues with it eventually.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
Sanne: Right :)

Sanne:
ok looks like we're all on the same page now... I'll merge. Of course feel free to amend if there's any need.

PR #831, opened 2021-06-01
Avoid use of InStream and AtomicInteger  for loops

DavideD:


DavideD:
Actually, there are a couple of things I need to check

DavideD:
It should be fine

DavideD:
@gavinking This should be an improvement because now we don't need the `AtomicInteger` nor the `InStream`.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveFlushingEventListener.java
```diff
@@ -129,9 +128,10 @@ private ReactiveActionQueue actionQueue(EventSource session) {
 		IdentitySet copiedAlready = new IdentitySet( 10 );
 		//safe from concurrent modification because of how concurrentEntries() is implemented on IdentityMap
 		return CompletionStages.loop(
-				Arrays.stream( persistenceContext.reentrantSafeEntityEntries() )
-						.filter( entry -> flushable( entry.getValue() ) ),
-				entry -> cascadeOnFlush( session, entry.getValue().getPersister(), entry.getKey(), copiedAlready )
+				persistenceContext.reentrantSafeEntityEntries(),
+				entry -> flushable( entry.getValue() )
+						? cascadeOnFlush( session, entry.getValue().getPersister(), entry.getKey(), copiedAlready )
+						: voidFuture()
```
DavideD, line 134: I don't think this pattern will cause any issue but peraphs I could filter the array before doing the loop.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -729,19 +727,20 @@ else if ( !delegate().isModifiableEntity( entry ) ) {
...
+								fields,
+								oldFields,
+								table == 0 ? rowId : null,
+								propsToUpdate,
+								table,
+								oldVersion,
+								updateStrings[table],
+								session
+						)
+						: voidFuture()
```
gavinking, line 743: See the thing about this change is that it means we get a longer chain, with a bunch of noop stages that the previous impl filtered out.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/DefaultReactiveMergeEventListener.java
```diff
@@ -456,20 +455,25 @@ private boolean existsInDatabase(Object entity, EventSource source, EntityPersis
...
+			Object[] mergeState,
+			Object[] managedState,
+			Integer i) {
+		// Cascade-merge mappings do not determine what needs to be fetched.
+		// The value only needs to be fetched if the incoming value (mergeState[i])
+		// is initialized, but its corresponding managed state is not initialized.
+		// Initialization must be done before copyValues executes.
+		return Hibernate.isInitialized( mergeState[i] ) && !Hibernate.isInitialized( managedState[i] )
+					? session.reactiveFetch( managedState[i], true )
+					: voidFuture();
```
gavinking, line 474: Again here, this results in a longer chain, doesn't it?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
```diff
@@ -155,25 +152,25 @@ public static void logSqlException(Throwable t, Supplier<String> message, String
...
-		}
-		else {
-			return voidFuture();
+			return AsyncTrampoline.asyncWhile(
+					index -> iterator.hasNext(),
+					index -> consumer
+							.apply( iterator.next(), index )
+							.thenApply( u -> index + 1 ),
+					0
+			).thenAccept( CompletionStages::voidFuture );
```
gavinking, line 163: OK, so I like getting rid of the `AtomicInteger`, for sure.

But if you're expecting this to be a performance optimization, I doubt that's true. Because here we're boxing an integer for each step of the loop, which creates garbage.

I would rather see an implementation that avoided boxing loop indices, which I believe is conceptually possible, but we would have to replace `BiFunction` with a different SAM.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
```diff
@@ -222,8 +206,17 @@ public static void logSqlException(Throwable t, Supplier<String> message, String
...
+		if ( start < end ) {
+			return AsyncTrampoline.asyncWhile(
+					index -> index < end,
+					index -> consumer
+							.apply( index )
+							.thenApply( u -> index + 1 ),
+					start
+			).thenCompose( CompletionStages::voidFuture );
+		}
+		return voidFuture();
```
gavinking, line 219: Again: I think a much more interesting optimization would be to get rid of the boxing by replacing `Function` with a different SAM.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveFlushingEventListener.java
gavinking, line 134: So I'm not sure but this looks like the same story: by not filtering out the entities which aren't flushable before calling `loop()`, we create a much longer chain.

Unless I'm missing something.

gavinking: Why don't you think it's an issue? It looks like a potentially serious degredation to me (but I might be wrong).

gavinking:
(Should clarify that with trampoline it's not literally a longer chain, but I'm supposing it still has some sort of cost compared to filtering out the noops upfront.)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
DavideD: Maybe... I don't know... I'm not sure how much we should worry about longer chains. This is  a CompletionStage that does nothing. Does it really have such a heavy impact?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
DavideD: Well, the goal was to remove the various castings, I guess now we can see if there are some changes compared to before

DavideD: > we would have to replace `BiFunction` with a different SAM.

I'd like to try that if you can give me some more details about it

DavideD: What's a SAM?

DavideD: Although, I did this change because I prefer this approach then having an accumulator around.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
gavinking: Well even if the only impact is a deeper stack trace, it's still bad, IMO.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
gavinking: Sorry, SMI, i've no idea what a SAM is either. :-)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveFlushingEventListener.java
DavideD: Sure, I got it the first time. No need to comment in every place I've used the same pattern :-)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
DavideD: That doesn't really help :)

gavinking: Single Method Interface.

i.e. an  `IntBiFunction` interface that accepts `int` instead of `Integer`.

DavideD: Ah, ok, thanks

DavideD:
I've just realized that I can pass the filter as a function, this way the call chain won't get affected. I will have another look at this tomorrow.

DavideD:
I think I have a nicer solution now and I will send a new PR shortly.

Thanks a lot for the feedback!

PR #832, opened 2021-06-01
Updates dependencies for JBang examples

DavideD:
I think it's better if one example doesn't require testcontainer, even if it means starting a database manually.
The other example still uses docker.

It also updates assertJ

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -3,7 +3,7 @@
   "aliases": {
     "example": {
       "script-ref": "tooling/jbang/Example.java",
-      "description": "An application example. Starts PostgreSQL using Testcontainers and Docker"
+      "description": "An application example. The database must be available before running it. See https://github.com/hibernate/hibernate-reactive/blob/main/podman.md"
     },
     "testcase": {
       "script-ref": "tooling/jbang/ReactiveTest.java",
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -26,7 +26,7 @@ dependencies {
     implementation "io.vertx:vertx-sql-client:${vertxVersion}"
 
     // Testing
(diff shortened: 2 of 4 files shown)


jbang-catalog.json
```diff
@@ -3,7 +3,7 @@
   "aliases": {
     "example": {
       "script-ref": "tooling/jbang/Example.java",
-      "description": "An application example. Starts PostgreSQL using Testcontainers and Docker"
+      "description": "An application example. The database must be available before running it. See https://github.com/hibernate/hibernate-reactive/blob/main/podman.md"
```
Sanne, line 6: I'm a big fan of podman, but still shouldn't we at least point to instructions for Docker users as well (and first and foremost) ?

I'm aware that the podman.md document also mentions things of interest for Docker users, but people will only notice that if they read it with attention - if I go there as a Docker user I don't think I'd notice it. Maybe we need a new document which is useful for both users?

Sanne: BTW `podman` can work with Testcontainers as well.

Just run this command first (and keep it running) in another terminal:

```
DOCKER_HOST="unix:/run/user/1000/podman/podman.sock" podman system service -t 3600
```

DavideD: Thanks, I couldn't remember which one was the command. But if I remember correctly you also need to pull the images first, isn't it?

Sanne: well pulling images isn't strictly necessary but if you don't then the Testcontainer boot seems prone to timeout w/o a sensible error message - but it seems to work ok most of the times.

DavideD: At the moment, we don't have them now for Hibernate Reactive because when we run the build, if you have docker, you can use test containers out of the box.

I will rename and reword the file but I will do it as a separate issue.
But also, if you are a docker user, you can run the other JBang alias :-)

Issue #807, opened 2021-05-26
withStatelessTransaction()


gavinking:
I hate the name, but I think it's best to have the same convenience method for stateless sessions.

DavideD:
Maybe `withStatelessAndTransaction` is less puzzling?

PR #833, opened 2021-06-02
added withStatelessTransaction() methods

blafond:
Fixes [#807](https://github.com/hibernate/hibernate-reactive/issues/807)

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -1622,6 +1622,20 @@ public interface Mutiny {
 		<T> Uni<T> withTransaction(BiFunction<Session, Transaction, Uni<T>> work);
 
 		/**
+		 * Perform work using a {@link StatelessSession reactive session} within an
+		 * associated {@link Transaction transaction}.
+		 * <p>
+		 * The session will be closed automatically and the transaction committed automatically.
+		 *
+		 * @param work a function which accepts the stateless session and returns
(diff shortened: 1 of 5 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1621,6 +1621,21 @@
...
+		 * <p>
+		 * The session will be {@link Session#flush() flushed} and closed
+		 * automatically, and the transaction committed automatically.
+		 *
+		 * @param work a function which accepts the session and returns
+		 *             the result of the work as a {@link Uni}.
+		 *
+		 * @see #withSession(Function)
+		 * @see StatelessSession#withTransaction(Function)
+		 */
```
gavinking, line 1636: @blafond this comment needs to change a bit, because there is no notion of flushing for a stateless session. Also the @see annotations point to the wrong things.

gavinking (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
blafond: working on it

blafond: comments fixed

blafond:
@gavinking  pushed commit that fixes the javadoc (and links) based on related java docs and operations within the reference methods.  I think we're good ?

gavinking (approved):
LGTM, thanks!!

Issue #740, opened 2021-04-29
Add a parameter to enable/disable the standard output stream 


DavideD:
We need it enabled on CI because some errors and warning won't be visible when something fails otherwise.

DavideD:
Thanks!

PR #834, opened 2021-06-03
added parameter to enable/disable standard output streams

blafond:
Fixes [#740](https://github.com/hibernate/hibernate-reactive/issues/740)

Included note in README.md

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -71,7 +71,7 @@ tasks.withType(Test) {
     defaultCharacterEncoding = "UTF-8"
     testLogging {
         displayGranularity 1
-        showStandardStreams = false
+        showStandardStreams = project.hasProperty('showStandardOutput')
         showStackTraces = true
         exceptionFormat = 'full'
         events 'PASSED', 'FAILED', 'SKIPPED'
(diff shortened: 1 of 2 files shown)


DavideD (approved):

hibernate-reactive-core/build.gradle
```diff
@@ -71,7 +71,7 @@ tasks.withType(Test) {
     defaultCharacterEncoding = "UTF-8"
     testLogging {
         displayGranularity 1
-        showStandardStreams = false
+        showStandardStreams = project.hasProperty('showStandardOutput') ? getProperty('showStandardOutput').toBoolean() : false
```
DavideD, line 74: On second thought, I think you can change this to
```
showStandardStreams = project.hasProperty('showStandardOutput')` 
```

We can assume that if somebody pass the property is because they want to enable it. Not reason to check the actual value.

DavideD (changes requested):

hibernate-reactive-core/build.gradle
DavideD: After that I think you can merge it

blafond: agreed

pushed those changes. Will merge when build complete, thx

PR #838, opened 2021-06-04
Adds jbang template to create reproducers

maxandersen:
With this you just do `jbang init -t reproducer@hibernate/hibernate-reactive mytest`
and you will have a `mytest.java`. 

Advantage is you can create any number of reproducers and you get your own file rather than 
editing the file in jbang's internal cache.

Other fixes:
- move #! at top of file as otherwise it breaks ability to run it directly.
- add simple sl4j to get useful output rather than warnings. you can also use sl4fj-nop if you just want it quite.

### DIFF
--- a/tooling/jbang/ReactiveTestTemplate.java.qute
+++ b/tooling/jbang/ReactiveTestTemplate.java.qute
@@ -1,0 +1,250 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+
+//DEPS io.vertx:vertx-pg-client:$\{vertx.version:4.1.0}
+//DEPS io.vertx:vertx-db2-client:$\{vertx.version:4.1.0}
+//DEPS io.vertx:vertx-mysql-client:$\{vertx.version:4.1.0}
+//DEPS io.vertx:vertx-unit:$\{vertx.version:4.1.0}
+//DEPS org.hibernate.reactive:hibernate-reactive-core:$\{hibernate-reactive.version:1.0.0.CR6}
+//DEPS org.assertj:assertj-core:3.19.0
+//DEPS junit:junit:4.13.2
+//DEPS org.testcontainers:postgresql:1.15.3
+//DEPS org.testcontainers:mysql:1.15.3
+//DEPS org.testcontainers:db2:1.15.3
+//DEPS org.testcontainers:mariadb:1.15.3
+//DEPS org.testcontainers:cockroachdb:1.15.3
+//DEPS org.slf4j:slf4j-simple:1.7.30
+//// Testcontainer needs the JDBC drivers to start the containers
+//// Hibernate Reactive doesn't use them
+//DEPS org.postgresql:postgresql:42.2.16
+//DEPS mysql:mysql-connector-java:8.0.25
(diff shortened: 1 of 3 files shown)


tooling/jbang/ReactiveTestTemplate.java.qute
```diff
@@ -0,0 +1,250 @@
...
+//DESCRIPTION   2. Run the test with JBang: `jbang mytest.java`
+//DESCRIPTION   3. (Optional) Edit the file (with IntelliJ IDEA for example):
+//DESCRIPTION             jbang edit --live --open=idea SampleIssueTest.java
+@RunWith(VertxUnitRunner.class)
+public class {baseName} {
+
+	/**
+	 * The database to use for the tests
+	 */
+	public final static Database DATABASE = Database.POSTGRESQL;
```
DavideD, line 82: Out of curiosity, is there a way to pass the database as a parameter during the template generation?
So that I could replace this with something like:
```
public final static Database DATABASE = Database.{database};
```

and maybe only include the dependencies for the selected database.

I'm trying to avoid a single template for each database but maybe that's the way to go.

DavideD (approved):
Thanks, looks great. I will merge it in a bit.

I have to admit that for the use case I had in mind I don't fully understand the benefit of having a template. It seems it requires an extra step.

Doesn't this give me the same result?
```
jbang testcase@hibernate/hibernate-reactive
jbang edit --live --open=idea testcase@hibernate/hibernate-reactive
```

DavideD:
I guess jbang doesn't create the file wihtout the `init` and the template

DavideD:
Merged, thanks

maxandersen:
> Doesn't this give me the same result?

no - it downloads the testcase into jbang cache and what you are editing is the file directly in jbang cache. meaning no guarantees when that file will be nuked/moved/overwritten.

I probably should try and prevent that edit operation in jbang.

tooling/jbang/ReactiveTestTemplate.java.qute
maxandersen: you *should* be able to refer to -Dkey=value in the templates but I haven't really tested/explored it yet. i've been waiting for a usecase to make the templates "smarter" - this one might be it.

feel free opne jbang issue on it and lets see - I'm cautious to avoid too much logic capbailities to the templating as I also want to be sure the templates generates even if no templates specified.

DavideD:
Couldn't you add an option instead?
```
```

DavideD:
```
jbang edit --live --open=idea  testcase@hibernate/hibernate-reactive Issue123.java
```

It's just a minor thing though

maxandersen:
What do you mean default name ? 

You just list a name - which for a reproducer would be something relevant for the issue. 

Or you prefer every producer you get has same name? :)

DavideD:
> Or you prefer every producer you get has same name? :)

Do I really care? Most of the time I want to try something and then throw it away.

Or even if different users send me two reproducer with the same name, I hope that the code will be at least different.
The default name is only for when the user doesn't set it: It's going to be `IssueTest.java` if not set or `HR123Test.java` if set.

Most of the users will only send a reproducer and then disappear.

DavideD:
Maybe I'm wrong but with a default name you could also run a template

DavideD:
By the way, I haven't used JBang that much so it's very possible that what I'm saying doesn't make sense.

For some reason, I just find it weird that I have to use a template when the only things that changes is the name of the class. While, if I use an alias I can just run or open the file without any additional steps.

But if that's the way it is, I will get used to it :-)

PR #839, opened 2021-06-04
Improve loop performance

### LINKED ISSUE #836: Improve performance of CompletionStages.loop
* Remove the `AtomicInteger`s
* Avoid use of InStream
* Avoid casting arrays to streams
* Don't chain unnecessary `CompletionStage`s

DavideD:
Fixes #836 

I still need to add a couple of test cases but otherwise the PR is ready for a review:

* It removes the AtomicInteger
* Avoid chaining emtpy CompletionStages when possible
* Avoid InStream
* Avoid casting arrays to Stream
* Avoid iterators where possible

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
@@ -24,7 +24,6 @@ import org.hibernate.persister.collection.CollectionPersister;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.proxy.HibernateProxy;
 import org.hibernate.reactive.session.ReactiveSession;
-import org.hibernate.reactive.util.impl.CompletionStages;
 import org.hibernate.type.AssociationType;
 import org.hibernate.type.CollectionType;
 import org.hibernate.type.CompositeType;
@@ -35,9 +34,11 @@ import org.hibernate.type.Type;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 
 import static org.hibernate.pretty.MessageHelper.infoString;
+import static org.hibernate.reactive.util.impl.CompletionStages.loop;
 import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
 
 /**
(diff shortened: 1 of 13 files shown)


DavideD:
I've added the missing tests and this is now ready to be reviewed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/Cascade.java
```diff
@@ -483,7 +484,7 @@ private void cascadeToOne(
 			persistenceContext.addChildParent( child, parent );
 			stage = stage.thenCompose( v -> action.cascade( eventSource, child, entityName, context, isCascadeDeleteEnabled ) )
 					.whenComplete( (vv, e) -> persistenceContext.removeChildParent( child ) )
-					.thenAccept( vv -> {} );
+					.thenCompose( CompletionStages::voidFuture );
```
gavinking, line 487: I just sent a PR that makes this change obsolete.

gavinking (approved):
LGTM!

gavinking:
But please don't forget to squash ;-)

DavideD:
> But please don't forget to squash ;-)

Sure, on Monday I will reduce the commits. Have a nice week-end

gavinking:
You too :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/DefaultReactiveDeleteEventListener.java
```diff
@@ -337,59 +337,56 @@ private void disallowDeletionOfDetached(DeleteEvent event) {
...
+										entityEntry.getDeletedState(),
 										persister,
-										isCascadeDeleteEnabled,
-										session
-								)
-						);
-					}
-				} );
-
-		CompletionStage<Void> afterDelete = cascadeAfterDelete( session, persister, entity, transientEntities );
```
gavinking, line 390: Hrm. @Davide I actually had this being called outside of the `thenCompose()` callback intentionally.

So this change actually changes the behavior of the code somewhat.

It probably doesn't break anything but I'm still not sure I like it.

DavideD:
Ok, but you cannot rely on the order of the operations if we don't chain
them this way ( theoretically). Why would you want to keep them separated?

gavinking:
> Ok, but you cannot rely on the order of the operations if we don't chain
them this way ( theoretically).

Well I think the ordering is still well-defined. All the non-async processing happens up-front, before any async process, and then the async stuff happens, with a well-defined order.

>  Why would you want to keep them separated?

Well, for one thing, imagine how impossible it's going to be to debug the cascading logic if it's happening over a bunch of async callbacks. I did it this way because I liked the idea of just doing all the sync stuff upfront.

DavideD:
I didn't expect this to be a controversial commit. I will remove it from this PR because it's not really related and I don't follow  you right now.

I'm going to have a better look at it in the future.

gavinking:
Yeah fine. I'm not completely against the change, I just think it bears more thinking-through.

DavideD:
Squashed things together, waiting for CI

PR #841, opened 2021-06-04
clean up some generic type issues

gavinking:
Some of this is far from perfect, and not actually typesafe but it's an improvement at least.


gavinking:
@DavideD WDYT?

DavideD:
LGTM

DavideD (approved):

PR #842, opened 2021-06-06
Separate JBang template for each database

DavideD:
It looks tidier now when everything is in a separate template and it's clearer what's needed for each database.

The only drawback is that we need to duplicate a lot of stuff among databases.
I didn't play with qute and template to see if it's possible to reuse a single file. 

I will have a look at it in the future

@maxandersen Any idea why we cannot run this class in IDEA as a main (once it has been initialized) but it works fine when we run it with jbang? Some missing configuration peraphs?


DavideD:
I'm having some trouble testing the db2 template because it doesn't start the docker image. I'm not sure if it's just me or something wrong in general. If somebody can try it and let me know, it would be nice.

gavinking (approved):
This looks good.

maxandersen:
What error are you seeing in idea ?

maxandersen (commented):
We should be able to handle this with qute and properties so it would not be this much duplication. I can't look / try out before later tonight / tomorrow though.

DavideD:
> We should be able to handle this with qute and properties so it would not be this much duplication. I can't look / try out before later tonight / tomorrow though.

It's certainly not urgent. Enjoying the weekend :-)

If I run the main from IDEA this is the result:
```
Execution failed for task ':test'.
> No tests found for given includes: [DbTest.main](filter.includeTestsMatching)
```

maxandersen:
Looks like you are trying to run it as a junit test not a main class ?

DavideD:
It's the opposite, I'm trying to run it [as main](https://github.com/hibernate/hibernate-reactive/blob/030195cf71a79dd4c7b6ad107f655d989cf9b434/tooling/jbang/CockroachDBReactiveTest.java.qute#L195) class instead as a JUnit test. Running the class as a JUnit test works fine in idea

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/842

We can improve these when we have time

PR #845, opened 2021-06-07
use NullValue to bind null values

### LINKED ISSUE #843: Handle Null parameters the right way with Vert.x SQL client
Based [on this comment](https://github.com/eclipse-vertx/vertx-sql-client/pull/970#issuecomment-854460745),  it seems that we are  not handling nul values the right way:.

See https://vertx.io/docs/vertx-mssql-client/java/#_handling_null

gavinking:
fixes #843

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
@@ -5,12 +5,15 @@
  */
 package org.hibernate.reactive.pool.impl;
 
+import java.sql.JDBCType;
 import java.sql.ResultSet;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 
+import io.vertx.sqlclient.data.NullValue;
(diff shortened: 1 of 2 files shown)


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/845! Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -264,4 +273,58 @@ public boolean hasNext() {
 	public CompletionStage<Void> executeBatch() {
 		return voidFuture();
 	}
+
+	private static void translateNulls(Object[] paramValues) {
```
tsegismont, line 277: If I understand correctly, this method will be invoked regardless of the DB vendor? In this case, it is a problem because `NullValue` so far is only used in MS SQL Client.

gavinking: Well, I mean, we tested its and nothing broke...

tsegismont: Yeah, not breaking `Tuple` behavior was a goal of the PR that introduced `NullValule` :-)

Sorry for the confusion

PR #850, opened 2021-06-08
[MSSQL] Fix UriConfigTest

DavideD:
We might need to check if it works for the JDBC Url


DavideD:
116 test failures at the moment

But I'm having a look at the one that seems simple

PR #854, opened 2021-06-11
support multiple calls on single eventloop thread

michalszynkiewicz:
Currently, when _operationA_ started a call with _withConnection_, but the connection is not established yet, and _operationB_ calls _withConnection_ on the same thread, _operationB_ will fail.
This change makes _operationB_ wait (non-blocking :) ) for the initialization of the connection.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java
@@ -24,7 +24,7 @@ import static org.hibernate.reactive.common.InternalStateAssertions.assertUseOnE
 final class ProxyConnection implements ReactiveConnection {
 
 	private final ReactiveConnectionPool sqlClientPool;
-	private ReactiveConnection connection;
+	private CompletionStage<ReactiveConnection> connection;
 	private boolean connected;
 	private boolean closed;
 	private final String tenantId;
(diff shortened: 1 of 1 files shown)


gavinking:
This doesn't seem right to me. A `ProxyConnection` belongs to a single `Session`, it's not shared between sessions. So if you're getting calls to the same `ProxyConnection` from different threads, that's because you're sharing a session between different reactive streams (which is certainly wrong).

DavideD:
Note: There are some failures on CI but they are related to the latest ORM snapshot. not this PR

michalszynkiewicz:
@gavinking It's a single thread but different invocations.

gavinking:
> Should these calls use different sessions?

Of course.

Hibernate sessions do not and have never been sharable between threads.

gavinking:
And it makes no difference that you're doing things in multiple "threads" rather than in multiple `Threads`. Logically, you're still sharing mutable state in a way that's not safe.

Sanne:
@michalszynkiewicz I *suspect* you're falling in the problem cathegory of Panache not storing the sessions in the right context yet - @FroMage is exploring some solutions.

But it's hard to be sure if your case is indeed the same - it would be really helpful if you could manage to reproduce this in a test which doesn't use Panache (ideally a test to be merged in this repository)

michalszynkiewicz:
@DavideD @Sanne @gavinking thank you guys! I'll close this PR then

michalszynkiewicz:
actually, should we check sth like if the vert.x context is the same for each usage of the ProxyConnection to fail each time someone tries to do such things?

gavinking:
We could do that indeed ... though probably the main issue is that the error message is not quite helpful. It should blame the user.

Sanne:
I'm understanding this is the source?

https://github.com/michalszynkiewicz/todos-with-rr-and-grpc/blob/2f6a97fdc5d9e27be260e63edcd25a6f9dffa364/src/main/java/com/example/storage/TodoStorage.java#L28-L37

So I see there's two different transactions started, which seems correct at first sight. But the problem is that Panache is obtaining a ReactiveSession instance by asking ARC for one - which follows the CDI convention, so if there  isn't one it will create one, but if there already is a ReactiveSession in the same scope you'll get the existing one.

In practice the two reactive methos are executed in "parallel" by interleaving  - even if in in the same thread - so what could happens - occasionally - is that the second method is initializing its own Transaction, but re-using the existing ReactiveSession which it finds in the scope.

This leads to having two "threads", with the intent of running two transactions, but sharing the same session: very bad.

A simple solution seems could be to patch Panache here: 

https://github.com/Sanne/quarkus/blob/18f1e8d424e9fb1a9e2f40d4b5190e30943844e5/extensions/panache/hibernate-reactive-panache/runtime/src/main/java/io/quarkus/hibernate/reactive/panache/Panache.java#L37-L39

I think it should always consistently open a new Session. @gavinking WDYT?

gavinking:
I mean the CDI model of session affinity simply doesn't support having two transactions against the same datasource in a single request. It should not have even let you do such a thing.

michalszynkiewicz:
Hmmm... no expert at CDI, but I'm quite sure request context for both invocations should be different

michalszynkiewicz:
fyi, we found the issue in request context propagation in grpc that was the cause of my problems that I tried to solve with this PR

PR #857, opened 2021-06-14
hibernate.jdbc.time_zone support for Time

### LINKED ISSUE #856: hibernate.jdbc.time_zone unsupported on TIME columns
Hi, we are trying to add the property `<property name="hibernate.jdbc.time_zone" value="UTC"/>` to our application, but we got `UnsupportedOperationException` on some MySQL `TIME` columns, may I know is there any plan to support that?

## Stacktrace

`java.lang.UnsupportedOperationException: null
	at org.hibernate.reactive.adaptor.impl.PreparedStatementAdaptor.setTime(PreparedStatementAdaptor.java:246)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: `

## Environment Information

- Version: 1.0.0.CR6

DavideD:
Fixes #856 ... sort of

This is what I would do and it seems consistent to what happens with ORM and MySQL (after a quick test).

I think OffsetTime is the one that's weird, because when we return the value it won't match the original Timezone.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/PreparedStatementAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/PreparedStatementAdaptor.java
@@ -158,6 +158,11 @@ public class PreparedStatementAdaptor implements PreparedStatement {
 	}
 
 	@Override
+	public void setTime(int parameterIndex, Time x, Calendar cal) {
+		put( parameterIndex, x.toLocalTime() );
+	}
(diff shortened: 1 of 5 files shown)


gavinking:
Well, OK, I agree that this is the most sane thing to do if we really have to support this setting, but it's *not* compatible with what care does, it's a different behavior.

DavideD:
> Well, OK, I agree that this is the most sane thing to do if we really have to support this setting, but it's not compatible with what care does, it's a different behavior.

Good enough for me :-)

PR #861, opened 2021-06-15
Add and describe build properties in gradle.properties

DavideD:
I've copied the properties we have there so that it make it easier to enable a specific build configuration when working in the IDE.
They are supposed to stay uncommented and only enabled when working locally.


gavinking (approved):
awesome, much better!

PR #865, opened 2021-06-17
Add support for MS SQL Server

### LINKED ISSUE #255: MS SQL Server support
Support for SQL Server is only available in Vert.x 4 at present ~~, and we're currently tied to Vert.x 3.9 by the need to support Quarkus~~.

Also, last time we checked, the SQL Server client was not very robust, and will need some bugfixes to work with HR.

DavideD:
Fixes #255 
Follows #852 

Support for MS SQL Server.

I had to refactor the code so that we could run the creation of temporary tables as queries and not preparedQueries:
See [StatementsWithParameters#execute](https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:mssql-rebase?expand=1#diff-cead1f530da1841e5e7e5108fcbf221200062f455968cdf18ae3600de6280a51R61) and [CreateTempTablesHandler](https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:mssql-rebase?expand=1#diff-89960259ba041b73dafa764ce1f08c072f46017561438a7dff7173271d2c6853R159).

I've also created a separate class for keeping the logic of the insert-and-select queries: https://github.com/hibernate/hibernate-reactive/commit/06528d56118d4bf99142ebb959ff3615e1457da3

We need to wait the next Vert.x SQL client release before we can merge this.


DavideD:
It seems I broke something on Db2... checking

DavideD:
Uff... everything works fine for me locally... I will check again tomorrow.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/bulk/impl/ReactiveIdTableSupport.java
```diff
@@ -28,7 +29,7 @@ public ReactiveIdTableSupport(Dialect dialect) {
 
     @Override
     public String generateIdTableName(String baseName) {
-        return "ht_" + baseName;
+        return (dialect instanceof SQLServerDialect ?  "#" : "ht_") + baseName;
```
gavinking, line 32: So ... we're not using SQL Server temporary tables anymore? We're using a regular table?

gavinking: Ooops, sorry, my mistake, I misread the diff.

DavideD:
I've found the problem... working on it

DavideD:
I think this is ready for a review now.
The failure on Db2 was happening because I wasn't ignoring the errors like in the original code.
In the end I've decided to refactor the way we handle the queries for the temporary tables because I was having a hard time understanding it. I think now it's clearer what happens and when.

I've kept the refactoring in a [separate commit](https://github.com/hibernate/hibernate-reactive/pull/865/commits/6bba87743da34285c79be784343a6c4f7310b4d6). @gavinking I would appreaciate if you could have a look at it.

It still need a release of the Vert.x client. @tsegismont Is there any idea about when it's going to happen?

tsegismont:
@DavideD we are working on bugfix release (`4.1.1`). At best, we would release by the end of this week.

tsegismont:
@DavideD @gavinking if I read correctly the build results, we have now 90% of the HR test suite passing? Do you know what's problematic in the remaining 10%? cc @vietj

DavideD:
> @DavideD we are working on bugfix release (`4.1.1`). At best, we would release by the end of this week.

Sounds great



> @DavideD @gavinking if I read correctly the build results, we have now 90% of the HR test suite passing? Do you know what's problematic in the remaining 10%? 

It's probably less than that. We have some tests that are not db dependent and we only run them for PostgreSQL.
I will let you know which tests we skipped for MSSQL specifically

DavideD:
@tsegismont, 
These are the only tests left: https://github.com/hibernate/hibernate-reactive/blob/32d4658035b99a389f24d1c8a1c5533d7726be9f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/BasicTypesForSelectedDBTest.java

When testing `@Lob` types it just gets stuck and JsonObject causes unsupported exception.
I suppose is expected.

tsegismont:
SQL Server doesn't have a json data type so we must keep the corresponding tests ignored.

For `@Lob`, do you use an `image` column or `binary`/`varbinary` ?

gavinking:
> @gavinking I would appreaciate if you could have a look at it.

OK let me find some time to try it out properly.

DavideD:
We create the following column:
```
        book varchar(MAX) // @Lob String
        pic varbinary(MAX) // @Lob byte[]
```
and then we try to do an insert of a string or an array of bytes (length between 87000 and 100000) we pass the values as parameters of type `String` or `BufferImpl` respectively. 

I said that it gets stuck but it might be just slow. The same tests doesn't take that long with the other databases though

tsegismont:
@DavideD I updated https://github.com/eclipse-vertx/vertx-sql-client/issues/608#issuecomment-629390027 about `varchar(max)` and `varbinary(max)`

DavideD:
Nice, thanks

gavinking:
@DavideD we're still waiting for a release of Vert.x before we can merge this patch, correct?

DavideD:
Yep

DavideD:
For the people unfamiliar with our test suite:

* This is the workflow that uses the latest Vert.x version: https://github.com/hibernate/hibernate-reactive/pull/865/checks?check_run_id=2883572435

* In all the other workflows, only the tests related to MS SQL Server fail (as expected because they are not using the latest vert.x snapshot).

gavinking:
OK thanks

DavideD:
Ah... do you need the configuration?

You can open it in the ide using these properties (in gradle.properties):
```
db = MSSQL
vertxVersion = 4.1.1-SNAPSHOT
enableSonatypeOpenSourceSnapshotsRep = true
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/InsertAndSelectParser.java
```diff
@@ -0,0 +1,78 @@
...
+	private InsertAndSelectParser() {
+	}
+
+	/**
+	 * Matches a SQL server query of this type ({@code default} is optional):
+	 * <pre>
+	 *     insert into entity_table default values select id, name from other_table;
+	 * </pre>
+	 */
+	private static final Pattern MSSQL_PATTERN = Pattern.compile( "^(insert into .*?)(( default)? values.*)" );
```
gavinking, line 35: Can we please not use a regex for this?

_Especially_ not a regex which matches `.*`.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/InsertAndSelectParser.java
```diff
@@ -0,0 +1,78 @@
...
+				: sql;
+	}
+
+	private static String parseDb2Query(String sql, String idColumnName) {
+		return "select " + idColumnName + " from NEW TABLE (" + sql + ")";
+	}
+
+	private static String parsePostgresSQLQuery(String sql, String idColumnName) {
+		return sql + " returning " + idColumnName;
+	}
```
gavinking, line 77: I don't think "parse" is quite the right name for these methods.

gavinking:
Apart from those comments it looks good.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/InsertAndSelectParser.java
DavideD: `convert`? `update`?

DavideD: I will go with `Processor` and `process`, a nice generic term

gavinking: That's fine.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/InsertAndSelectParser.java
DavideD: This should solve the issue: https://github.com/hibernate/hibernate-reactive/pull/865/commits/196658305fc17a2f413c6345f22fc6a61c0da3bb

It still uses a regex but only to check the last part of the string end now queries are generated once.

gavinking: What exactly is that regex supposed to be doing? Matching parentheses? **And** assuming that you never have more than one level of nesting?

DavideD: It matches the pattern `values (?, ?, ?, ?, ?)` in a query of type `insert into Table (col1, col2) values (?,?)`. Where the number of parameters varies. We only use it for query that have to return the id after an insert. What kind of nesting can we have?

DavideD: But I guess this is not going to work if somebody uses a `@ColumnTransformer`. Uff, I will add a test tomorrow

gavinking: I think regexes are just terrible in at almost everything, but they're particularly terrible at parsing programming languages, since they simply can't handle any language with a recursive grammar (which is every language).

And look, _perhaps_ there's no scenario right now where we let you have an expression in an `insert` list, but SQL certainly allows it, so it's just not robust. I already ripped out the use of broken regexes from some of the dialects in H6, and used handwritten parsers instead. The results was less code, and less-broken code. And it's easier to understand.

DavideD: This approach should work in all cases: https://github.com/hibernate/hibernate-reactive/pull/865/commits/80f7691e06464cad18fc2e35a708783d371344e3

Although, it feels like we should send a patch to ORM but I haven't looked too close on why the queries are different. For now I was interested in finding a solution for Reactive.

DavideD:
@tsegismont Any idea why now that I've updated to vert.x 4.1.1 MS SQL Server gives this error?

```
    java.lang.UnsupportedOperationException: Unsupported token: 0
    	at io.vertx.mssqlclient.impl.codec.MSSQLCommandCodec.decode(MSSQLCommandCodec.java:80)
    	at io.vertx.mssqlclient.impl.codec.TdsMessageDecoder.decodeMessage(TdsMessageDecoder.java:65)
    	at io.vertx.mssqlclient.impl.codec.TdsMessageDecoder.channelRead(TdsMessageDecoder.java:43)
    	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
    	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
    	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
    	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:296)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
    	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
    	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
    	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
    	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
    	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166)
    	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:719)
    	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:655)
    	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:581)
    	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
    	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
    	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
    	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
```

It seems to happen with several basic queries like `drop table MapsIdPerson` but from the error I cannot understand what's wrong.

tsegismont:
No idea. I will look into it tomorrow morning.

>

tsegismont:
@DavideD this is a regression introduced after refactoring the MS SQL codec. It happens when executing a prepared query that leads to an error (e.g. `delete from basic` when the table does not exist).

I filed https://github.com/eclipse-vertx/vertx-sql-client/issues/1000

tsegismont:
@DavideD after building https://github.com/eclipse-vertx/vertx-sql-client/issues/1001 and running the HR suite using these Gradle props:

```
db = MSSQL
enableMavenLocalRepo = true
vertxVersion = 4.2.0-SNAPSHOT
```

We need to discuss with @vietj when this fix can be delivered. In the meantime, you can workaround the issue by not running prepared queries if table does not exist.

DavideD:
> We need to discuss with @vietj when this fix can be delivered. In the meantime, you can workaround the issue by not running prepared queries if table does not exist.

I'm not sure we can do this Wouldn't it be possible to release a `4.1.2` with this fix? This seems like a serious issue.

DavideD:
It seems that everything is working fine with the latest Vert.x 4.1.2.
I will merge it if there are no objections.

gavinking:
alright!

DavideD:
Merged!

tsegismont:
Awesome news!

gavinking:
> Awesome news!

Indeed.

DavideD:
Thank you!

By the way, @tsegismont, is there a roadmap somewhere about which databases are going to be next? It would be nice to have something on the website to let users know.

I've seen a [PR for Oracle](https://github.com/eclipse-vertx/vertx-sql-client/pull/1011), so I guess that will be next.

tsegismont:
There is a PR for Oracle and another for ClickHouse DB (contribution by a community member). Other than that we don't plan to support new databases soon.

PR #873, opened 2021-06-25
JoinColumnTest now works as expected on Db2

DavideD:
I think it works now because we fix all those tests where we weren't closing the session correctly


gavinking:
Excellent!

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/873

PR #876, opened 2021-06-28
Introduce a new test to verify multi-line parsed import.sql scripts

Sanne:
This seems to work fine already; doesn't work in Quarkus for some reason.


DavideD:
Thanks

PR #884, opened 2021-07-01
Update complexMultilineImports.sql

DavideD:
Now it works for all the databases


gavinking:
Heh, was just about to mention this.

DavideD:
Yes, sorry, there were a couple of errors that I missed because they were hidden by others. I'm working on making the build stable again

DavideD:
This should fix the main build. I've disabled the test for Db2 because I'm unable to start it and I don't know how to update the query. I will get back to it in the future

PR #888, opened 2021-07-06
Enable sanity assertions by default

Sanne:
What would you all think of this?

Untested as my local DBs are busy.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/InternalStateAssertions.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/InternalStateAssertions.java
@@ -5,6 +5,8 @@
  */
 package org.hibernate.reactive.common;
 
+import java.util.Locale;
+
 import io.vertx.core.Context;
 
 /**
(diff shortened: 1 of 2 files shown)


DavideD:
Could you rebase this to the latest changes, please?

The only thing that worries me about this change is that people will see an exception after the next upgrade.
But not fixing it might have unexpected results and  they can still disable the check if they can't, or don't want to, fix it

DavideD:
Another question before I merge this. Are this checks already enabled/disabled in Quarkus? (Basically, when we upgrade in Quarkus, will everything still work out of the box?)

Sanne:
No not all tests are passing in Quarkus, for example Panche Reactive was apparently still running tests on RESTEasy "blocking".

I've sent a PR https://github.com/quarkusio/quarkus/pull/18464 , but let's wait to merge this one as I couldn't verify the full suite yet.

DavideD:
ok

gavinking:
I'm not sure I understand this.

DavideD:
The check we have in place to make sure that the session is used in the right thread is disabled as a default in Hibernate Reactive. This pull request changes it by having it enabled as a default.

Or is something else that's not clear?

gavinking:
Well, OK. I don't have any specific objection, just a general feeling of unease that I don't have really clear in my head what are the implications of this.

DavideD:
It's possible the somebody is using the session is the wrong thread but doesn't see any error. After the upgrade they will have an exception and they will need to figure out what happened and how to fix it. Or they can just disable the check.

it shouldn't affect anybody else.

But I remembered that in Quarkus we kept this property disabled for the time being because there were some tests failing, that's why I asked about it.

Sanne:
@gavinking I had initially made it a "softer" opt-in because it would have broken many different modules in Quarkus.

But this required me to explciitly enable the checks on selected modules as we fixed them; I think we're finally done with that, so would at very least want to switch the default to make sure there is no module hiding somewhere without the appropriate changes.

In general, I also believe end users should really make sure to run this all in the right thread, as encoded by these assertions - otherwise there's racy access to mutable state by multiple threads, which is very tricky to diagnose: better to fail fast and provide useful guidance.

If I'm off here with excessive guards for some reasonable use case I didn't think of, I think it would be good to get details; at very least this makes sure people come forward.

Sanne:
[I've rebased it and added an additional commit.]

Sanne:
Ok let's merge this, it has been very useful for me to catch some integration issues still with Panache Reactive, and I don't think it is likely to get "in the way" of people doing reasonable things.

PR #891, opened 2021-07-07
upgrade Mutiny to 0.19.2

gavinking:
@DavideD is there any reason our Mutiny is so far behind?

### DIFF
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -14,7 +14,7 @@ dependencies {
         exclude group: 'javax.activation', module: 'javax.activation-api'
     }
 
-    api 'io.smallrye.reactive:mutiny:0.18.0'
+    api 'io.smallrye.reactive:mutiny:0.19.2'
 
     //Logging
     implementation 'org.jboss.logging:jboss-logging:3.4.1.Final'


DavideD:
No reason, I didn't realise they were already at 19.2.
@Sanne Is it OK to upgrade this, isn't it?

DavideD:
Quarkus is still on 0.18 but I suppose is safe to assume they will upgrade soon

gavinking:
What actually *happens* when we have a difference in a dependency version compared to Quarkus?

Do they just override the one we declare? @Sanne?

DavideD:
I think new dependencies might not work out of the box with GraalVM because they might introduce new classes that needs special handling to work in native mode.

That's why we tend to be extra careful when dealing with these type of upgrades

Sanne:
Yes Quarkus has the final word on all dependencies, overriding what's specified here.

So we need to keep HR building and testing with the dependency versions we expect to have in the Quarkus version in which we'll integrate the next HR version.

We can of course block component upgrades in Quarkus, or suggest upgrades in there if it's important, but this needs to work for other components as well.

Sanne:
So specifically about this PR: yes it should be a relatively safe bet to integrate.

There is a remote possibility that Mutiny `0.19.2` doesn't get in Quarkus as soon as we'd expect (some critical regression?) , but this event would need to happen while we also want to upgrade urgently for whatever reason - and in such highly unlikely case we can always revert.

gavinking:
> Quarkus has the final word on all dependencies, overriding what's specified here.

Well, then honestly it seems to me it's OK to do this upgrade. I don't really see any realistic risk that we'll stop working against the older version of Mutiny. Our usage of Mutiny APIs is *very* minimalistic.

Sanne:
Right. Also there's good changes we're still compatible with the old version.

gavinking:
Thanks

Sanne:
FYI https://github.com/quarkusio/quarkus/pull/18518

PR #894, opened 2021-07-09
Use JBoss logging annotations

### LINKED ISSUE #571: Move to JBoss Logging annotations
We can use `HR` as code and for the list of ids this should do:

* Core: 1-1000
* PostgreSQL: 1101-1200
* MySQL: 1201-1300
* MariaDB: 1301-1400
* Db2: 1401-1500
* CockroachDB: 1501-1600
* Sql Server: 1601-1700

and so on...

We need to match the dependency version in Quarkus.

DavideD:
Fixes #571 

If it passes the build this should be ready to be merged.
I think I've replaced all the places where we use the Logger from ORM with ours.

Plus, I've created new entries for the errors of type: HibernateException, PersistenceException, LazyInitializationException and IllegalStateException.


Sanne (approved):
Ok, I was a bit worried about the new usage of `MethodHandles.lookup()` in the scope of native-images but it seems to work fine.

Issue #886, opened 2021-07-04
Stateless update doesn't work


Xset-s:
I have two sample entities:

SampleEntity:

```java
@Entity
@Table(name = "sample_entities")
public class SampleEntity implements Serializable {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    public Long id;

    @Column(name = "sample_field")
    public String sampleField;
}
```

SampleJoinEntity:

```java
@Entity
@Table(name = "sample_join_entities")
public class SampleJoinEntity implements Serializable {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    public Long id;

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "sample_entity_id", referencedColumnName = "id")
    public SampleEntity sampleEntity;
}
```

and after that I would like to execute this code:

```java
<!-- SessionFactory -->
Mutiny.SessionFactory sessionFactory =
        createEntityManagerFactory( "mysql" )
                .unwrap(Mutiny.SessionFactory.class);

<!-- Persist everything -->

// sampleEntity
SampleEntity sampleEntity = new SampleEntity();
sampleEntity.sampleField = "test";

sessionFactory.withTransaction((session, transaction) ->
        session.persist(sampleEntity)
).await().indefinitely();

// sampleJoinEntity
SampleJoinEntity sampleJoinEntity = new SampleJoinEntity();
sampleJoinEntity.sampleEntity = sampleEntity;

sessionFactory.withTransaction((session, transaction) ->
        session.persist(sampleJoinEntity)
).await().indefinitely();

<!-- Find just persisted entity again -->
Long targetId = sampleJoinEntity.id;

SampleJoinEntity sampleJoinEntityFromDatabase = sessionFactory.withTransaction((session, transaction) ->
        session.find(SampleJoinEntity.class, targetId)
                .call((entity) -> session.fetch(entity.sampleEntity))
).await().indefinitely();

<!--Change something and update it using stateless session-->
SampleEntity sampleEntityFromDatabase = sampleJoinEntityFromDatabase.sampleEntity;
sampleEntityFromDatabase.sampleField = "test";

// EXCEPTION IS HERE!
sessionFactory.withStatelessTransaction((session, transaction) ->
        session.update(sampleEntityFromDatabase)
).await().indefinitely();
```

[...]

PR #900, opened 2021-07-20
handle update of detached entity use case

blafond:
Fixes [#886](https://github.com/hibernate/hibernate-reactive/issues/886)

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
@@ -223,23 +223,24 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl
     public CompletionStage<Void> reactiveUpdate(Object entity) {
         checkOpen();
         if ( entity instanceof HibernateProxy ) {
+            Object proxy = entity;
             final LazyInitializer hibernateLazyInitializer = ( (HibernateProxy) entity ).getHibernateLazyInitializer();
             if ( hibernateLazyInitializer.isUninitialized() ) {
-                return reactiveFetch( entity, true ).thenCompose(
-                        fetchedEntity -> executeReactiveUpdate( fetchedEntity ) );
+                return reactiveFetch( entity, true )
+                        .thenCompose( fetchedEntity -> executeReactiveUpdate(proxy, fetchedEntity) );
             }
             else {
-                return executeReactiveUpdate( hibernateLazyInitializer.getImplementation() );
+                return executeReactiveUpdate( proxy, hibernateLazyInitializer.getImplementation() );
             }
         }
 
-        return executeReactiveUpdate( entity );
+        return executeReactiveUpdate( null,  entity );
     }
 
-    private CompletionStage<Void> executeReactiveUpdate(Object entity) {
+    private CompletionStage<Void> executeReactiveUpdate(Object proxy, Object entity) {
         ReactiveEntityPersister persister = getEntityPersister( null, entity );
         Serializable id = persister.getIdentifier( entity, this );
-        Object[] state = persister.getPropertyValues( entity );
+        Object[] state = proxy != null ? persister.getPropertyValues( proxy ) : persister.getPropertyValues( entity );
         Object oldVersion;
         if ( persister.isVersioned() ) {
             oldVersion = persister.getVersion( entity );
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
@@ -223,24 +223,23 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl
     public CompletionStage<Void> reactiveUpdate(Object entity) {
         checkOpen();
         if ( entity instanceof HibernateProxy ) {
-            Object proxy = entity;
             final LazyInitializer hibernateLazyInitializer = ( (HibernateProxy) entity ).getHibernateLazyInitializer();
             if ( hibernateLazyInitializer.isUninitialized() ) {
                 return reactiveFetch( entity, true )
-                        .thenCompose( fetchedEntity -> executeReactiveUpdate(proxy, fetchedEntity) );
+                        .thenCompose( fetchedEntity -> executeReactiveUpdate(entity, fetchedEntity) );
             }
             else {
-                return executeReactiveUpdate( proxy, hibernateLazyInitializer.getImplementation() );
+                return executeReactiveUpdate( entity, hibernateLazyInitializer.getImplementation() );
             }
         }
 
-        return executeReactiveUpdate( null,  entity );
+        return executeReactiveUpdate( entity,  entity );
     }
 
-    private CompletionStage<Void> executeReactiveUpdate(Object proxy, Object entity) {
+    private CompletionStage<Void> executeReactiveUpdate(Object stateEntity, Object entity) {
         ReactiveEntityPersister persister = getEntityPersister( null, entity );
         Serializable id = persister.getIdentifier( entity, this );
-        Object[] state = proxy != null ? persister.getPropertyValues( proxy ) : persister.getPropertyValues( entity );
+        Object[] state = persister.getPropertyValues( stateEntity );
         Object oldVersion;
         if ( persister.isVersioned() ) {
             oldVersion = persister.getVersion( entity );
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
@@ -222,6 +222,21 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl
     @Override
     public CompletionStage<Void> reactiveUpdate(Object entity) {
         checkOpen();
+        if ( entity instanceof HibernateProxy ) {
+            final LazyInitializer hibernateLazyInitializer = ( (HibernateProxy) entity ).getHibernateLazyInitializer();
+            if ( hibernateLazyInitializer.isUninitialized() ) {
+                return reactiveFetch( entity, true ).thenCompose(
+                        fetchedEntity -> executeReactiveUpdate( fetchedEntity ) );
+            }
+            else {
+                return executeReactiveUpdate( hibernateLazyInitializer.getImplementation() );
+            }
+        }
+
+        return executeReactiveUpdate( entity );
+    }
+
+    private CompletionStage<Void> executeReactiveUpdate(Object entity) {
         ReactiveEntityPersister persister = getEntityPersister( null, entity );
         Serializable id = persister.getIdentifier( entity, this );
         Object[] state = persister.getPropertyValues( entity );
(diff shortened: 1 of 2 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -222,20 +222,28 @@ public ReactiveEntityPersister getEntityPersister(String entityName, Object obje
...
-        ReactiveEntityPersister persister = getEntityPersister( null, entity );
-        Serializable id = persister.getIdentifier( entity, this );
+        if ( entity instanceof HibernateProxy ) {
+            return reactiveFetch( entity, true ).thenCompose(
+                    fetchedEntity -> executeReactiveUpdate( entity, fetchedEntity ) );
+        }
+        return executeReactiveUpdate( entity, entity );
+    }
+
+    private CompletionStage<Void> executeReactiveUpdate(Object entity, Object updatableEntity) {
```
DavideD, line 232: Why do you need both entities as parameters?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveStatelessUpdateTest.java
```diff
@@ -0,0 +1,101 @@
...
+				)
+
+				.chain( joinEntityFromDatabase -> getMutinySessionFactory().withStatelessTransaction(
+						(s, t) -> {
+							SampleEntity entityFromDb = joinEntityFromDatabase.sampleEntity;
+							entityFromDb.sampleField = "updated field";
+							// Without the fix it fails here
+							return s.update( entityFromDb );
+						} )
+				)
```
DavideD, line 64: I would add a query to check that the value has been actually updated

DavideD (changes requested):

DavideD:
Thanks, looks good. Just a couple of minor changes.

@gavinking is this  what you had in mind?

gavinking:
Thanks. Give me a couple of hours to have a look at this i'm a bit swapped out and I need to remember what was going on here.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
blafond: The **proxy** `entity` is contains in with the updated properties but the **fetched** entity has the non-updated properties. The Object[] state values are pulled from the `entity` but applied to the **fetched** entity. I could pass in the **state** array but the **persister.getPropertyValues(entity)** is used to extract them.

blafond: The **proxy** `entity` is contains in with the updated properties but the **fetched** entity has the non-updated properties. The Object[] state values are pulled from the `entity` but applied to the **fetched** entity. I could pass in the **state** array but the **persister.getPropertyValues(entity)** is used to extract them.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveStatelessUpdateTest.java
blafond: I'll add it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
gavinking: I think it might be a little bit clearer if we renamed the method parameters to `proxy, entity`, or `reference, entity`or something like that.

gavinking (approved):
This looks pretty reasonable to me. Thanks!

gavinking:
Wait, wait, perhaps I spoke too fast. I'm not sure we really want to actually call `fetch()` @blafond, though certainly that would *work*.

All we really need to do is unwrap the proxy, right? Which we should be able to do without actually calling `fetch()` but by using similar logic to what `fetch()` does internally, something like `((HibernateProxy) entity).getHibernateLazyInitializer().getImplementation()`.

Does that make sense?

blafond:
@DavideD pushed the changes which include some comments in the tests.

@gavinking added logic to handle the managed proxy use-case as well as a second test to cover that use-case.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveStatelessUpdateTest.java
```diff
@@ -0,0 +1,147 @@
...
+				.chain( targetId -> getMutinySessionFactory()
+						.withTransaction( (session, transaction) -> session
+								.find( SampleJoinEntity.class, sampleJoinEntity.id ) )
+				)
+
+				.chain( joinEntityFromDatabase -> getMutinySessionFactory().withStatelessTransaction(
+						(s, t) -> s
+								// fetching entity in this transaction to trigger logic for
+								// updating a managed proxy entity
+								// @See ReactiveStatelessSessionImpl.updateReactive()
```
DavideD, line 94: I would add this comment on the javadoc for the class, so that it's actually linkable:
```
/**
 * @see org.hibernate.reactive.session.impl.ReactiveStatelessSessionImpl#reactiveUpdate(Object)
 */
public class ReactiveStatelessUpdateTest  extends BaseReactiveTest {
```

DavideD (approved):
A couple of minor comments but other than that it looks good.

@gavinking In this particular use case, using the `( (HibernateProxy) entity ).getHibernateLazyInitializer().getImplementation()` doesn't work because the proxy is not initialized. But the changes @blafond  applied to the PR cover that case as well.

gavinking:
> @gavinking In this particular use case, using the `( (HibernateProxy) entity ).getHibernateLazyInitializer().getImplementation()` doesn't work because the proxy is not initialized. But the changes @blafond applied to the PR cover that case as well.


Well, wait, I *must* be missing something because that just sounds completely wrong!

If the entity is unfetched, then `update()` would be a noop! So why fetch it?!

I mean, at most `update()` should simply *ignore* an unfetched entity, right?

DavideD:
> If the entity is unfetched, then `update()` would be a noop! So why fetch it?!

Mmmh... I will have a second look at the tests

DavideD:
Ooops... the tests are wrong. It seems I've checked them too quickly

DavideD:
> I mean, at most update() should simply ignore an unfetched entity, right?

Because of a problem with the test we didn't notice we were overriding the  state of the entity before doing the update.
But if we keep track of the state correctly, we can fetch the entity and then update it, even if the proxy has not been initialized (I think). Or there is a problem with this approach?

@blafond is working on fixing this and we should be able to discuss over some code soon

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -223,23 +223,24 @@ public ReactiveEntityPersister getEntityPersister(String entityName, Object obje
     public CompletionStage<Void> reactiveUpdate(Object entity) {
         checkOpen();
         if ( entity instanceof HibernateProxy ) {
+            Object proxy = entity;
```
DavideD, line 226: There is no need for this additional variable, just pass `entity`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -223,23 +223,24 @@ public ReactiveEntityPersister getEntityPersister(String entityName, Object obje
...
-        return executeReactiveUpdate( entity );
+        return executeReactiveUpdate( null,  entity );
     }
 
-    private CompletionStage<Void> executeReactiveUpdate(Object entity) {
+    private CompletionStage<Void> executeReactiveUpdate(Object proxy, Object entity) {
         ReactiveEntityPersister persister = getEntityPersister( null, entity );
         Serializable id = persister.getIdentifier( entity, this );
-        Object[] state = persister.getPropertyValues( entity );
+        Object[] state = proxy != null ? persister.getPropertyValues( proxy ) : persister.getPropertyValues( entity );
```
DavideD, line 243: I know that I said that passing the same value `entity` for both values was weird. It turns out I was wrong and it would make sense in this case. There  is really no need to add an if-then here. But, please, add a comment to explain that we expect proxy to contains the state that we want to save in the database

blafond: So yes, I can remove the if-then and change the non-proxy call to:  `executeReactiveUpdate( entity,  entity );`

For me, it's a little confusing just looking in the `executeReactiveUpdate(proxy, entity)` method because it makes it appear to always be a proxy object... but I can do that.

DavideD:
@gavinking The reason the proxy is uninitialized is not because there hasn't been any change, it's because the user is updating the field without using a setter.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
DavideD: I think renaming a parameter is easier than to add extra logic, but that's me :-)

DavideD:
After some more thinking, I'm not sure we should merge this The user is trying to update an entity that they've obtain from a detached instance. I'm not sure how the stateless update is supposed to work in this scenario but I suspect it's not a use-case we should cover. We should have a better error message for this situations, though

DavideD:
Note that this only works because they are not getting the entity they are trying to update using a getter, otherwise they would see a `LazyInitializationException`

gavinking:
>  The user is trying to update an entity that they've obtain from a detached instance. I'm not sure how the stateless update is supposed to work in this scenario but I suspect it's not a use-case we should cover. 

Well in a stateless session *every* entity instance is detached. So there's nothing a priori wrong with what you've just said.

So if they have an object which 

- was fetched (initialized) and 
- they have changed a field value 

then it's perfectly reasonable to expect that they should be able to call `update()` on it.

When I first saw the description of this issue I thought that the problem was simply that the stateless session wasn't unwrapping the proxy to get at the fetched entity instance underneath (which would be something trivial to fix).

But if it's something more than that then I guess I don't actually understand the issue at all.

DavideD:
> Well in a stateless session every entity instance is detached

Right... :-)
In the original test the fetch wasn't done the right way so the entity wasn't initialized.

OK, so I guess it makes sense to make it work after the fetch.
I will update the PR

gavinking:
> In the original test the fetch wasn't done the right way so the entity wasn't initialized.

Yeah, so if it was unfetched then this is a kinda uninteresting bug. Legitimate things to do would be:

- just ignore it and do nothing, or
- throw a meaningful exception that says "entity is uninitialized" or whatever.

I don't have an especially strong opinion on that, but perhaps the second is better, I'm not sure.

But I don't think it makes sense to go fetch it and then do nothing. That doesn't seem useful.

DavideD:
> But I don't think it makes sense to go fetch it and then do nothing.

We were actually doing an update. there are some changes. Not sure why you think we are doing nothing

DavideD:
Anyway, I'm closing this one and send a new PR.

gavinking:
Well if it was unfetched I don't see how it could possibly have any changes in it worth sending to the database.

I mean, sure, you could send off a useless update statement with the exact same state you just fetched from the table, but that wouldn't be useful at all.

DavideD:
This is the part of the test doing the update:
```
// This entity hasn't been fetched, so the proxy is uninitialized
SampleEntity entityFromDb = joinEntityFromDatabase.sampleEntity;
// Setting the new value
entityFromDb.sampleField = "updated field";
// What do we expect here? The proxy is uninitialized (because  we aren't using a setter, I think)
return s.update( entityFromDb );	
```

gavinking:
> we aren't using a setter, I think

We unless we're in Quarkus, isn't it just a plain user error to access entity state directly via fields?

It's only in Quarkus that you're allowed to use direct field access, right?

DavideD:
> We unless we're in Quarkus, isn't it just a plain user error to access entity state directly via fields?

Ok, It's actually explained in both documentations (ORM and Reactive). I didn't remember this to be a strong requirement.

PR #902, opened 2021-07-23
add a gotcha

gavinking:
Let's try making this even more prominent.

### DIFF
--- a/documentation/src/main/asciidoc/reference/introduction.adoc
+++ b/documentation/src/main/asciidoc/reference/introduction.adoc
@@ -393,6 +393,10 @@ It's illegal to access persistent fields from outside the entity class.
 Therefore, external access to persistent fields must be intermediated via
 getter and setter methods defined by the entity class.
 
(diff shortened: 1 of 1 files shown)


DavideD:
I'm not sure that additional paragraph help.
I think the main issue is that all these kind of info is spread inside the documentation and it's easy to miss (or forget) unless you read the whole document from start to finish multiple times.

What I think would help is to have a box at the end of the documentation with the most important (and easily missable things to remember when working with Hibernate Reactive (for example: session is not thread safe, associations should be updated both sides, use of getter/setter for accessing fields is not optional). This way we wil be able to link it when a user asks questions.

gavinking:
> I think the main issue is that all this kind of info is spread inside the documentation 

Well, sure, but there's not much we can do about that. If there's lots of information in the docs, then by nature it's spread out.

At least with this new callout here we maximize the probability of people actually reading it, and also we tell them what sort of thing could go wrong if they do the wrong thing.

Issue #536, opened 2021-02-04
schema update or validation


yrodiere:
See https://github.com/quarkusio/quarkus/issues/14264#issuecomment-772690764

Whenever one sets `hibernate.hbm2ddl.auto` to `validate` or `update`, the schema management tool apparently tries to get a JDBC connection and ultimately fails with an error like this:

```
java.lang.RuntimeException: java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.test.QuarkusUnitTest.beforeAll(QuarkusUnitTest.java:502)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeBeforeAllCallbacks$8(ClassBasedTestDescriptor.java:368)
        at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeBeforeAllCallbacks(ClassBasedTestDescriptor.java:368)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.before(ClassBasedTestDescriptor.java:192)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.before(ClassBasedTestDescriptor.java:78)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:136)
        at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:129)
        at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:127)
        at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:126)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:84)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
[...]

PR #903, opened 2021-07-24
[536] [913] [910] SchemaMigrator/SchemaValidator support for PostgreSQL and CockroachDB

gbadner:
For https://github.com/hibernate/hibernate-reactive/issues/536, https://github.com/hibernate/hibernate-reactive/issues/913, and https://github.com/hibernate/hibernate-reactive/issues/910 .

Requires ORM PR:
https://github.com/hibernate/hibernate-orm/pull/4106

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -74,55 +74,55 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public String getString(int columnIndex) {
-		String string = row.getString(columnIndex);
+		String string = row.getString( columnIndex - 1 );
 		return (wasNull=string==null) ? null : string;
 	}
 
 	@Override
 	public boolean getBoolean(int columnIndex) {
-		Boolean bool = row.getBoolean(columnIndex);
+		Boolean bool = row.getBoolean(columnIndex - 1);
 		wasNull = bool == null;
 		return !wasNull && bool;
 	}
 
 	@Override
 	public byte getByte(int columnIndex) {
-		Integer integer = row.getInteger( columnIndex );
+		Integer integer = row.getInteger( columnIndex - 1 );
 		wasNull = integer == null;
 		return wasNull ? 0 : integer.byteValue();
 	}
 
 	@Override
 	public short getShort(int columnIndex) {
-		Short aShort = row.getShort( columnIndex );
+		Short aShort = row.getShort( columnIndex - 1 );
 		wasNull = aShort == null;
 		return wasNull ? 0 : aShort;
 	}
 
 	@Override
 	public int getInt(int columnIndex) {
-		Integer integer = row.getInteger( columnIndex );
+		Integer integer = row.getInteger( columnIndex - 1 );
 		wasNull = integer == null;
 		return wasNull ? 0 : integer;
 	}
 
 	@Override
 	public long getLong(int columnIndex) {
-		Long aLong = row.getLong(columnIndex);
+		Long aLong = row.getLong( columnIndex - 1 );
 		wasNull = aLong == null;
 		return wasNull ? 0 : aLong;
 	}
 
 	@Override
 	public float getFloat(int columnIndex) {
-		Float real = row.getFloat( columnIndex );
+		Float real = row.getFloat( columnIndex - 1 );
 		wasNull = real == null;
 		return wasNull ? 0 : real;
 	}
 
 	@Override
 	public double getDouble(int columnIndex) {
-		Double real = row.getDouble( columnIndex );
+		Double real = row.getDouble( columnIndex - 1 );
 		wasNull = real == null;
 		return wasNull ? 0 : real;
 	}
@@ -134,32 +134,32 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public byte[] getBytes(int columnIndex) {
-		Buffer buffer = row.getBuffer(columnIndex);
+		Buffer buffer = row.getBuffer( columnIndex - 1 );
 		wasNull = buffer == null;
 		return wasNull ? null : buffer.getBytes();
 	}
 
 	@Override
 	public Date getDate(int columnIndex) {
-		LocalDate localDate = row.getLocalDate(columnIndex);
+		LocalDate localDate = row.getLocalDate( columnIndex - 1 );
 		return (wasNull=localDate==null) ? null : java.sql.Date.valueOf(localDate);
 	}
 
 	@Override
 	public Time getTime(int columnIndex) {
-		LocalTime localTime = row.getLocalTime(columnIndex);
+		LocalTime localTime = row.getLocalTime( columnIndex - 1);
 		return (wasNull=localTime==null) ? null : Time.valueOf(localTime);
 	}
 
 	@Override
 	public Time getTime(int columnIndex, Calendar cal) {
-		LocalTime localTime = row.getLocalTime(columnIndex);
+		LocalTime localTime = row.getLocalTime( columnIndex - 1 );
 		return (wasNull=localTime==null) ? null : Time.valueOf(localTime);
 	}
 
 	@Override
 	public Timestamp getTimestamp(int columnIndex) {
-		LocalDateTime localDateTime = row.getLocalDateTime(columnIndex);
+		LocalDateTime localDateTime = row.getLocalDateTime( columnIndex - 1 );
 		return (wasNull=localDateTime==null) ? null : Timestamp.valueOf(localDateTime);
 	}
 
@@ -214,7 +214,28 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public long getLong(String columnLabel) {
-		Long aLong = row.getLong( columnLabel );
+		// PostgreSQL stores sequence metadata in information_schema as Strings.
+		// First try to get the value as a Long; if that fails, try to get
+		// as a String and try to parse it as a Long.
+		Long aLong;
+		try {
+			aLong = row.getLong(columnLabel);
+		}
+		catch (ClassCastException ex) {
+			// Check if the value is a String that can be converted to a Long
+			final String aString = row.getString(columnLabel);
+			// aString won't be null; check just because...
+			try {
+				aLong = aString != null ? Long.parseLong( aString ) : null;
+			}
+			catch (ClassCastException exNotAString) {
+				// The value is neither a long nor a String that can be
+				// parsed as a long.
+				// Throw the original exception.
+				throw ex;
+			}
+		}
+
 		wasNull = aLong == null;
 		return wasNull ? 0 : aLong;
 	}
@@ -311,7 +332,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public <T> T getObject(int columnIndex, Class<T> type) {
-		T object = row.get(type, columnIndex);
+		T object = row.get(type, columnIndex - 1);
 		return (wasNull=object==null) ? null : object;
 	}
 
@@ -430,6 +451,8 @@ public class ResultSetAdaptor implements ResultSet {
 
 			@Override
 			public String getColumnTypeName(int column) {
+				// This information is in rows.columnDescriptors().get( column-1 ).dataType.name
+				// but does not appear to be accessible.
 				return null;
 			}
 
@@ -467,7 +490,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public Object getObject(int columnIndex) {
-		Object object = row.getValue( columnIndex );
+		Object object = row.getValue( columnIndex - 1 );
 		return (wasNull=object==null) ? null : object;
 	}
 
@@ -485,7 +508,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public BigDecimal getBigDecimal(int columnIndex) {
-		BigDecimal decimal = row.getBigDecimal(columnIndex);
+		BigDecimal decimal = row.getBigDecimal(columnIndex - 1);
 		return (wasNull=decimal==null) ? null : decimal;
 	}
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
@@ -51,6 +51,22 @@ public interface ReactiveConnection {
 	CompletionStage<Result> select(String sql, Object[] paramValues);
 	CompletionStage<ResultSet> selectJdbc(String sql, Object[] paramValues);
 
+	/**
+	 * This method is intended to be used only for queries returning
+	 * a ResultSet that must be executed outside of any "current"
+	 * transaction (i.e with autocommit=true).
+	 * <p/>
+	 * For example, it would be appropriate to use this method when
+	 * performing queries on information_schema or system tables in
+	 * order to obtain metadata information about catalogs, schemas,
+	 * tables, etc.
+	 *
+	 * @param sql - the query to execute outside of a transaction
+	 * @param paramValues - a non-null array of parameter values
+	 * @return the CompletionStage<ResultSet> from executing the query.
+	 */
+	CompletionStage<ResultSet> selectJdbcOutsideTransaction(String sql, Object[] paramValues);
+
 	CompletionStage<Long> insertAndSelectIdentifier(String sql, Object[] paramValues);
 	CompletionStage<Long> selectIdentifier(String sql, Object[] paramValues);
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
@@ -1,0 +1,330 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.provider.service;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hibernate.boot.model.naming.Identifier;
+import org.hibernate.tool.schema.extract.internal.AbstractInformationExtractorImpl;
+import org.hibernate.tool.schema.extract.spi.ExtractionContext;
+import org.hibernate.tool.schema.extract.spi.InformationExtractor;
+
+/**
+ * An implementation of {@link InformationExtractor} that obtains metadata
+ * information from a database's information_schema.
+ *
+ * @author Gail Badner
+ */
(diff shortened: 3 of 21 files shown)


DavideD:
I will convert this to a draft, because it contains changes to the `gradle.properties` file that we don't want on the main branch. And also because we have to wait for ORM to include the PR that we need.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdatePostgreSqlTestBase.java
```diff
@@ -0,0 +1,339 @@
...
+																	 )
+												)
+										)
+										.thenCompose( v -> s.createNativeQuery( indexDefinitionQuery, String.class )
+												.setParameter( 1, "aanother" )
+												.getSingleResult()
+												.thenAccept( result ->
+																	 context.assertEquals(
+																			 result,
+																			 "CREATE UNIQUE INDEX aanother_pkey ON public.aanother USING btree (id)"
```
DavideD, line 210: Not critical, but when using `context.equals(expected, actual)`, the firs value is the expected one and the second the actual one.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
```diff
@@ -50,6 +50,7 @@
 	CompletionStage<Result> select(String sql);
 	CompletionStage<Result> select(String sql, Object[] paramValues);
 	CompletionStage<ResultSet> selectJdbc(String sql, Object[] paramValues);
+	CompletionStage<ResultSet> selectJdbcOutsideTransaction(String sql, Object[] paramValues);
```
DavideD, line 53: Could you remind us why you need the additional method?

gavinking:
I have not tested it but the code looks great! Thanks @gbadner 

(The only very minor comment I have is I would prefer that SQL keywords were in lowercase like the rest of the SQL we generate.)

gbadner:
@DavideD, thanks for doing making it a draft. I thought I defined it as a draft when I created it, but maybe I missed that.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdatePostgreSqlTestBase.java
gbadner: I normally do it that way, I don't know why I didn't in this test. I'll change it.

gbadner: @DavideD, I pushed a [commit|https://github.com/hibernate/hibernate-reactive/pull/903/commits/99eb01a207506b1536aee6ca7ed4a2252db57632] that corrects the order.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
gbadner: @DavideD , since `SchemaMigrator` executes select queries on system tables and executes DDL statements (which update system tables), I assumed that these queries should be executed outside of a transaction (i.e., with `autocommit=true`). IIUC, this would help avoid contention with other work the DB is doing on schemas used by various applications.

If you look at the corresponding ORM code in [GenerationTargetToDatabase]
(https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/exec/GenerationTargetToDatabase.java#L74-L85): you can see that the statement is created using a connection obtained from a `DdlTransactionIsolator`.

Javadoc for `DdlTransactionIsolator` says:
```
 * Provides access to a Connection that is isolated from
 * any "current transaction" with the designed purpose of
 * performing DDL commands
```
If you look at the `DdlTransactionIsolator` implementations, you will see that `autocommit` is set to `true` on the `Connection` that is obtained:
`DdlTransactionIsolatorJtaImpl`
`DdlTransactionIsolatorNonJtaImpl`
`DdlTransactionIsolatorProvidedConnectionImpl` relies on `JdbcConnectionAccessProvidedConnectionImpl` which sets `autocommit` to `true`

I didn't see any other way to execute a query outside of a transaction that returned a `CompletionStage<ResultSet>`, so I created a new method.

Is there some other method that should be used instead?

gbadner:
> I have not tested it but the code looks great! Thanks @gbadner

Thanks @gavinking!

> 
> (The only very minor comment I have is I would prefer that SQL keywords were in lowercase like the rest of the SQL we generate.)

That's fine. I can change that.

gbadner:
I'm currently using the Hibernate ORM codestyle. Which of the [codestyles](https://github.com/hibernate/hibernate-ide-codestyles/tree/main/intellij-14) do you use for Hibernate Reactive?

gbadner:
I finally got around to taking a good look at what else needs to be done for `SchemaValidator`, and, AFAICT, it's already working with the code provided here!?!

gbadner:
@gavinking, I added a new [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/ba2e5f9411c1af09e61c65e6ad7946f8e9aeae3f) that changes SQL keywords to lower-case.

gavinking:
Thanks, @gbadner

DavideD:
> @DavideD, thanks for doing making it a draft. I thought I defined it as a draft when I created it, but maybe I missed that.

no problem

> I'm currently using the Hibernate ORM codestyle. Which of the [codestyles](https://github.com/hibernate/hibernate-ide-codestyles/tree/main/intellij-14) do you use for Hibernate Reactive?

Yes, that's what we should use. Thanks.

Although, our build doesn't check the code style (except of a couple of basic rules) so you might find that we are a bit inconsistent sometimes.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
DavideD: I don't think we had this requirement before. I don't see any problem adding a new method.

But I mainly wanted a bit of an explanation written somewhere so that others know as well (maybe it wasn't obvious only to me though ;-)

At some point, we will need to add some javadoc to that class. Anyway, not something you need to worry about at the moment.

gbadner: Yes, I know this stuff is lacking in documentation. I will be adding that.

gbadner:
s@DavideD, I just pushed a commit that adds Javadoc and a couple of small tweaks.

I forgot that I had already pushed a commit for encapsulating methods into `ExtractionTool` and ended up squashing the two commits, so I had to force-push.

Sanne is still looking at the ORM PR. I think this PR is ready to review, although we will obviously have to wait for the ORM PR to be finalized.

Unless I hear of any required changes to this PR, I will start work on adding support for MySQL.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -430,6 +430,8 @@ public String getCatalogName(int column) {
 
 			@Override
 			public String getColumnTypeName(int column) {
+				// This information is in rows.columnDescriptors().get( column-1 ).dataType.name
+				// but does not appear to be accessible.
```
gavinking, line 434: @gbadner this is used by `SchemaValidator`, is that right?

Should we ask the vert.x guys to expose this info on `ColumnDescriptor`?

gavinking:
This all looks very nice, @gbadner!

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gbadner: @gavinking,

TL; DR: yes, it would help to expose this info on `ColumnDescriptor`.

For the record, here are the details:

If they expose it, then we can use `ResultSetMetaData` like what is done in ORM [here](https://github.com/hibernate/hibernate-orm/blob/a64d0102812474898059f1143bf09764f95556f1/hibernate-core/src/main/java/org/hibernate/tool/schema/extract/internal/InformationExtractorJdbcDatabaseMetaDataImpl.java#L152-L167) in the ORM [PR](https://github.com/hibernate/hibernate-orm/pull/4106).

Currently, since that information is not available in `ResultSetMetaData`, [`AbstractReactiveInformationSchemaBasedExtractorImpl#processColumnsResultSet`](https://github.com/hibernate/hibernate-reactive/pull/903/files#diff-149c9cc5fe97eafff78515120d487c5934371492bd13929253d0ec88278e9316R220-R259) is executed instead. 

Notice also that the SQL type code gets returned as `null` by `AbstractReactiveInformationSchemaBasedExtractorImpl#processColumnsResultSet` (which ends up being 0 (zero), which is java.sql.Types.NULL) because it is not available. It would be helpful to get that SQL code; otherwise, this [line](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaValidator.java#L165) constructing an error message might not make sense (or may throw an exception?).

From looking at the code, it appears that `ResultSetAdapter`'s implementation of `ResultSetMetaData#getColumnType` returns the SQL type code properly, but I haven't debugged into it to be sure.

AFAICT, `ColumnInformation#getNullable`, `#getColumnSize`, and `#getDecimalDigits` are not used by ORM's main branch, so if `ColumnDescriptor#dataType#name` gets exposed, we'll have everything ORM currently needs. I know that there has been talk about using some of this information in the future (in ORM 6?), but I'm not sure about that.

gbadner:
Thanks @gavinking, I'm pretty happy with it.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
```diff
@@ -0,0 +1,93 @@
...
+		service = registry.getService( ReactiveConnectionPool.class );
+	}
+
+	@Override
+	public <T> T getQueryResults(
+			String queryString,
+			Object[] positionalParameters,
+			ResultSetProcessor<T> resultSetProcessor) throws SQLException {
+
+		final CompletionStage<ReactiveConnection> connectionStage = service.getConnection();
```
gbadner, line 69: @gavinking, I noticed that [`ReactiveGenerationTarget#executeCommands`](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveGenerationTarget.java#L91-L105) uses the same `ReactiveConnection` to execute all of the DDL statements.

Here, `Service#getConnection` gets called for each query that gets executed. That `ReactiveConnection` gets closed in the `finally` block below.

Would it be better to keep a `ReactiveConnection` around so the same one is used for all of the queries?

gavinking: I can't think of a reason why it would matter.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: I opened https://github.com/eclipse-vertx/vertx-sql-client/issues/1020.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
gbadner: OK, I'll leave it as is.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
```diff
@@ -0,0 +1,260 @@
...
+			Object parameter,
+			StringBuilder sb,
+			List<Object> parameters) {
+
+		if ( parameter != null ) {
+			parameters.add( parameter );
+			sb.append( clause );
+			// TODO: PostgreSQL does not seem to accept "?" parameter placeholders.
+			//       It does accept $<position>. Will this work for all dialects?
+			sb.append( "$").append( parameters.size() );
```
gbadner, line 171: I see that this doesn't work for MySQL. It looks like `PostgresParameters` knows to make this change.

I'm looking into how I can change this to use `Parameters`...

gbadner: I just added a [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/cd964b9b5609045249954280779c117e85d2d452) to use `Parameters` to get the dialect-specific placeholder.

gbadner:
I just added a [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/82fb199aa1e0b4e62f91c7a9704a079a455e9fb1) to deal with a non-standard information_schema.columns column name used by PostgreSQL.

gbadner:
I pushed a [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/63041ad00c0e90119b5df48f8cfcfdeb2e7920ea) that deals with PostgreSQL's storing sequence metadata as a String when ORM expects a Long.

This was the issue I mentioned [here](https://github.com/hibernate/hibernate-reactive/issues/536#issuecomment-848386513).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -214,7 +214,20 @@ public int getInt(String columnLabel) {
...
+		Long aLong;
+		try {
+			aLong = row.getLong(columnLabel);
+		}
+		catch (ClassCastException ex) {
+			// Check if the value is a String that can be converted to a Long
+			final String aString = row.getString(columnLabel);
+			// aString won't be null; check just because...
+			aLong = aString != null ? Long.parseLong(aString) : null;
+		}
```
gavinking, line 229: @gbadner won't this swallow the original CCE and cause it to be rethrown as something potentially misleading if it's *not* a string?

gbadner: Yes, you're right! I'll fix it to throw the original CCE in that case.

gavinking: Perfect, thanks 🙏

gbadner: @gavinking, I pushed a [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/a4ec4b3e14b75639acbef9c8ff20a384534f20e6) to fix this. Does it look OK to you?

gavinking: Yup that's fine.

gbadner:
Pushed a [commit](https://github.com/hibernate/hibernate-reactive/pull/903/commits/f0639b1d1a5088be83965c18c417000e19cfe68a) for fixing on CockroachDB.

gbadner:
There are small differences in results returned from querying the `pg_indexes` system table using CockroachDB compared to PostgreSQL. This only affected the CockroachDB unit test.

It would be better to change the test to not rely on `pg_indexes`. I think that would allow changing `SchemaUpdatePostgreSqlTestBase` to be used for both PostgreSQL and CockroachDB.

gbadner:
@gavinking, rebased and settings in build.properties commented out.

gbadner:
@gavinking,I'm seeing failures in tests that restrict ORM versions [5.5,5.6).

Should I remove .github/workflows/tracking-orm-5.build.yml ?

gavinking:
Umm, I'm not sure TBH. Perhaps @Sanne knows what to do.

gbadner:
@gavinking, I'd like to add the commits from https://github.com/hibernate/hibernate-reactive/pull/940 into this PR, then close https://github.com/hibernate/hibernate-reactive/pull/940.

Any reason why I shouldn't do that?

gavinking:
Fine by me.

gbadner:
Merged commits from https://github.com/hibernate/hibernate-reactive/pull/940.

gavinking:
> Merged commits from #940.

Great.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
```diff
@@ -124,17 +126,17 @@ protected Configuration constructConfiguration() {
 			doneTablespace = true;
 		}
 		//Use JAVA_TOOL_OPTIONS='-Dhibernate.show_sql=true'
-		configuration.setProperty( Settings.SHOW_SQL, System.getProperty(Settings.SHOW_SQL, "false") );
+		configuration.setProperty( Settings.SHOW_SQL, System.getProperty(Settings.SHOW_SQL, "true") );
```
gavinking, line 129: @gbadner could you roll back this change please?

P.S. A more convenient way to enable SQL logging during development is to set:

    logger.hibernate.level = debug

in `log4j.properties`.

gavinking (changes requested):

gavinking:
> I'm seeing failures in tests that restrict ORM versions [5.5,5.6).

Is this the only thing blocking us from merging this?

Sanne:
We should no longer test with Hibernate ORM 5.5 since this requires those changes which are exclusively in 5.6

Sanne:
P.S. I've already upgraded Quarkus to use ORM `5.6.0.Beta1` so I'm not concerned about dropping compatibility with 5.5

gavinking:
> We should no longer test with Hibernate ORM 5.5 since this requires those changes which are exclusively in 5.6

Yes, I agree.

gavinking:
So am I understanding correctly that as of this change, we now have two redundant runs of the test suite for every CI build?

gavinking:
Anyway, OK, right, @gbadner, this seemed to fix it:

<https://github.com/gavinking/hibernate-reactive/commit/7d8e19b88206e3b32697397ca804045762192e42>

gbadner:
I pushed a commit that disables SQL logging and includes @gavinking 's  commit to fix DI. Crossing my fingers that everything passes...

gbadner:
Hurray, all passed!

gavinking:
WDYT shall we merge it?

gbadner:
@gavinking, yes, please do!

gavinking:
It's still a "Draft" :)

gbadner:
Oops, it's "Open" now.

gavinking:
There's a lot of commits here, should we squash?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
gbadner: @gavinking, this is done.

gbadner:
Sure, I'll squash and push to my branch...

gavinking (approved):

gavinking:
Excellent! Fantastic to finally have this done!

Thanks @gbadner

Sanne:
Nice!

PR #904, opened 2021-07-26
fix JDBC batching

### LINKED ISSUE #877: Value of `hibernate.jdbc.batch_size` is ignored
It doesn't matter what we set it to, it's always 0.

The reason is that in `SessionFactoryOptionsBuilder`, [there's a check to see if the database supports batching](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/boot/internal/SessionFactoryOptionsBuilder.java#L428) and it always return false. It happens because we don't extract the info from the database

gavinking:
Attempt to fix #877.

WDYT @DavideD?


DavideD:
It seems to work.

Thanks @gavinking.
We still need a way to test that batching is actually used when we want to use it.

gavinking:
yeah well. that's much harder

DavideD:
I think testing that the connection is an instance of `BatchingConnection`  should be doable.
Not foolproof but better than nothing. @blafond can you have a look at it please?

PR #905, opened 2021-07-26
Issue 886

DavideD:
Follows https://github.com/hibernate/hibernate-reactive/pull/900

I've tried to apply what we discussed in the previous PR, but there is still something that doesn't work.
I've added 4 tests, all of which updates the proxy:

-  :heavy_check_mark:  1. Unfetched  poxy and value is set using setter: Throws lazy initialization exception
-  :heavy_check_mark: 2. Unfetched  poxy and value is assigned to the field: Throws an exception about the value being unfetched
-  :heavy_check_mark: 3. Fetched proxy and  assign the new value to the field: It works as expected ( no errors and value gets updated)
-  :x: 4. Fetched proxy and  assign the value using setter: No exceptions but the value doesn't get udpated

I can't figure out why it doesn't work when using setter/getter (case 4). Any idea?
This is the test: https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:issue-886?expand=1#diff-b4f970573d11d1e8b5616b5c7b8a93f89d8cdf37ae235d9648ec01145e1a9bb9R150

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
@@ -209,6 +209,9 @@ public interface Log extends BasicLogger {
 	@Message(id = 71, value= "No name provided and multiple persistence units found")
 	PersistenceException noNameProvidedAndMultiplePersistenceUnitsFound();
 
+	@Message(id = 72, value= "Cannot update an uninitialized proxy. Make sure to fetch the value before trying to update it: %1$s")
+	HibernateException uninitializedProxyUpdate(Object entity);
+
 	// Same method that exists in CoreMessageLogger
 	@LogMessage(level = WARN)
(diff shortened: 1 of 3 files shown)


gavinking:
I mean from what you just described it sounds like you're passing down a proxy instead of the underlying instance. it's the 3rd case that should fail, not the 4th.

DavideD:
I've figured it out, I was checking the values in the wrong object. It should be fixed now.

What should we do with the 3rd test? Should we remove it or should  I leave it to check if something changes in the future?

DavideD:
Nevermind, it doesn't work without the setter now. I will remove the test

DavideD:
It think this is ready to be merged if it passes on CI

gavinking:
> What should we do with the 3rd test?

Yep, remove it.

gavinking (approved):
Looks fine.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/905

Thanks

PR #907, opened 2021-07-28
add built-in support for Vert.x JsonObject as a type

### LINKED ISSUE #869: Support JsonObject as a built-in type
We have a test in `BasicTypesForSelectedDBTest` testing that we can save a `JsonObject` via a `UserType`.
But a user needs to create a Json user type and apply it to the column:
```
		@Type(type = "org.hibernate.reactive.types.Json")
		@Column(columnDefinition = "json")
		private JsonObject jsonObj;
```

Should we provide support for this type out of the box?

gavinking:
fixes #869


DavideD:
Merged, thanks

PR #921, opened 2021-08-02
add section about equals()/hashCode() to doc

gavinking:


DavideD:
I will check this in a bit.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -401,6 +401,41 @@ values!
...
+
+TIP: We therefore recommend identifying a _natural key_ for each entity,
+that is, a combination of fields that uniquely identifies an instance of
+the entity, from the perspective of the data model of the program. The
+business key should correspond to a unique constraint on the database,
+and to the fields which are included in `equals()` and `hashCode()`.
+
+IMPORTANT: If you can't identify a natural key, it's _highly likely_
+that you have not thought hard enough about your data model, or that
+there's something actually wrong with your model. If your entity doesn't
```
DavideD, line 431: Maybe it's just me, but I don't feel comfortable telling a user that they haven't thought hard enough about an issue.

DavideD: We could keep it short and say that without a natural identifier, it's not possible to distinguish between two objects with the same fields and that usually, they could add a column containing a generated identifier.

DavideD (changes requested):
It's good but it needs an example.
Showing an entity with several fields (mutable, immutable and an identifier) would go a long way.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: But that's not what I'm trying to say here: I've had too many arguments with people who insist that their table has no natural key, and so they want to use the surrogate key as the hashcode, which causes problems. 

What I'm trying to get them to do is think harder about their model and identify the natural key.

And yeah, I have never ever been shy about telling users when they're wrong.

DavideD: I don't see how telling somebody "Think harder" is going to help them.
Has it ever worked in the past?

Anyway, I think you can rewrite the same paragraph without mentioning the user:
```
IMPORTANT: If your model doesn't have a natural key, it's _highly likely_ that there's something wrong with it.
```
> And yeah, I have never ever been shy about telling users when they're wrong.

Really, I haven't noticed :-)
That said, no user has made any statement, so I think we can give them the benefit of the doubt.  This is not twitter after all.

gavinking: > Has it ever worked in the past?

Absolutely yes!

gavinking: Really, I think it's much more important to give people good advice that will help them have success with Hibernate than to avoid the potential for momentary slight emotional discomfort.

DavideD: I don't see how the two things are related. It's possible to give good advice without being patronizing. 

> > Has it ever worked in the past?
> 
> Absolutely yes!

I doubt it.

I think it's something we will never agree on.
Feel free to merge whichever version you prefer.

Sanne (changes requested):

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -401,6 +401,41 @@ values!
 When you use Hibernate Reactive in Quarkus, these requirements are relaxed,
 and you can use public fields instead of getters and setters if you prefer.
 
+=== `equals()` and `hashCode()`
+
+Entity classes should override `equals()` and `hashCode()`. People new to
```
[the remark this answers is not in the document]
DavideD: > Also there's loads of literature on this topic, do we need to add more here?

I think we do. I don't see why somebody should look at this information somewhere else. I understand that other places could explain everything in deeper detail, but the documentation should at least give an idea about how we expect those method to be implemented for an entity.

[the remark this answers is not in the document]
gavinking: > Also I've never seen a solid and persuasive example in defence of the idea that entities really should declare hashcode and equals like you're stating here. 

I mean, I think it's just generally excepted that "data classes" should have "value equality", otherwise known as "equality".

Really "identity equality" is a sort-of garbage concept that doesn't really have a place outside of _very_ technical code, in my opinion.

(Try explaining the notion of "identity equality" to a functional programmer. It doesn't even _make sense_ in an environment which promotes immutability.)

Anyway, I would *always* define a proper notion of equality for all my entities. I need to anyway, because I certainly need to add the corresponding unique constraints to the database.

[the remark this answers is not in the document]
Sanne: You both have good points and I'm OK to soften my position, but please bear in mind that "identity equality" just works fine in all cases which don't involve re-attaching detached entities (which I dislike anyway) - and generally doesn't even need to be explained.

So for me it's still like the preferred approach and also simplest to implement.

This conversation really makes me think that after all these years we haven't solved this yet - perhaps we should take a step back and try to make this easier.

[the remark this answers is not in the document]
gavinking: Yes, sure, identity equality works, and I don't mind adding a clarification to state that.

But it seems to me that most users are going to want to define value equality, and it is very well-worth having clear guidance for those who do.

And y'know, I don't really trust the sources that they will find on google.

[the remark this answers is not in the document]
Sanne: Ok agreed. Thanks!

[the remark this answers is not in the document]
gavinking: I toned down the language a little, and added the proposed clarification.

I'm scared it's a bit too weak now, but whatever. Sure wouldn't want to hurt anyone's precious feelings.

PR #924, opened 2021-08-02
native queries returning Tuple

### LINKED ISSUE #919: createNativeQuery(queryString,Tuple.class).getResultList()  method is not returning list of Tuple data
I am using reactive hibernate (1.0.0.CR7) for my spring webflux project. I am getting the object array data while executing native queries and it is working fine. But I need it as a tuple object. So I passed Tuple.class as a second argument to createNativeQuery method. But it will throw an exception as it cannot cast an object to tuple. I just need the data with column alias name. So I choose tuple instead of object array. Kindly help.

Sample code :

```
String queryString= "select c.country_name as country_name, c.phone_code as phone_code
from erp_country c"; // MySql query
Mutiny.Query<Tuple> query = session.createNativeQuery(queryString, Tuple.class);
List<Tuple> list = query.getResultList().await().indefinitely();
for (Tuple tuple : list) { // here it will throw an exception as object cannot be cast to class javax.persistence.Tuple
     System.out.println(tuple.get("country_name") + "-" + tuple.get("phone_code"));
}
```
Exception Message:

`Ljava.lang.Object; cannot be cast to class javax.persistence.Tuple ([Ljava.lang.Object; is in module java.base of loader 'bootstrap'; javax.persistence.Tuple is in unnamed module of loader 'app')
`

I have another project using JPA- hibernate 5.5. There I can pass the second parameter of createNativeQuery as tuple class instead of giving an entity class. So I thought may be there is a chance to use Tuple in reactive hibernate also. If we use tuple class then we can retrieve the data by using alias name in the query string.

Few Reference :

[https://stackoverflow.com/questions/16296853/jpa-result-list-cast-object-to-tuple](url)
[https://stackoverflow.com/questions/44626609/getting-column-names-from-a-jpa-native-query](url)

gavinking:
fixes #919

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
@@ -207,11 +207,7 @@ public class MutinySessionImpl implements Mutiny.Session {
 
 	@Override
 	public <R> Mutiny.Query<R> createNativeQuery(String sql, Class<R> resultType) {
-		boolean knownType = delegate.getFactory().getMetamodel()
-				.entityPersisters().containsKey( resultType.getName() );
-		return knownType
-				? new MutinyQueryImpl<>( delegate.createReactiveNativeQuery(sql, resultType), factory )
-				: new MutinyQueryImpl<>( delegate.createReactiveNativeQuery(sql), factory );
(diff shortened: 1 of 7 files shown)


DavideD:
This looks alright, but for some reason the CI build didn't complete. I don't think it's related to the PR but I've restarted it to be sure

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/924

Thanks

PR #925, opened 2021-08-02
Tests to check for BatchingConnection if batch size > 1

blafond:
Tests to support [#877](https://github.com/hibernate/hibernate-reactive/issues/877)

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
@@ -6,11 +6,16 @@
 package org.hibernate.reactive;
 
 import io.vertx.ext.unit.TestContext;
+
 import org.hibernate.LockMode;
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.cfg.Configuration;
 import org.hibernate.reactive.common.AffectedEntities;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.mutiny.impl.MutinySessionImpl;
+import org.hibernate.reactive.pool.BatchingConnection;
 import org.hibernate.reactive.stage.Stage;
+import org.hibernate.reactive.stage.impl.StageSessionImpl;
 
 import org.junit.After;
 import org.junit.Test;
@@ -26,6 +31,7 @@ import java.util.Objects;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
 
 public class ReactiveSessionTest extends BaseReactiveTest {
@@ -698,6 +704,28 @@ public class ReactiveSessionTest extends BaseReactiveTest {
 								)
 						)
 		);
+	}
+
+	@Test
+	public void testBatchingConnection() {
+		Stage.Session session = getSessionFactory().openSession();
+		try {
+			assertThat(((StageSessionImpl)session).getReactiveConnection()).isInstanceOf( BatchingConnection.class );
+		}
+		finally {
+			closeSession( session );
+		}
+	}
+
+	@Test
+	public void testBatchingConnectionMutiny() {
+		Mutiny.Session session = getMutinySessionFactory().openSession();
+		try {
+			assertThat(((MutinySessionImpl)session).getReactiveConnection() ).isInstanceOf( BatchingConnection.class );
+		}
+		finally {
+			closeSession( session );
+		}
 	}
 
 	@Test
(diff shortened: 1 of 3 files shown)


gavinking:
Hrm, @blafond I was really hoping to not add `unwrap()` methods to our public API. JPA has them because there are multiple vendor implementations, but we don't have that need here.

gavinking:
Hrm, @blafond I was really hoping to not add `unwrap()` methods to our public API. JPA has them because there are multiple vendor implementations, but we don't have that need here.

blafond:
The `ReactiveSessionImpl.unwrap()` isn't available on our `Mutiny.Session` or `Stage.Session` interfaces and are not directly available in our tests. Not sure how to pull that info out of a `ReactiveSession` without going through those interfaces.

.. and I do agree... seems a little heavy change just to perform a test.

gavinking:
Yeah, well, right, and this is exactly why I didn't add tests for this when I fixed the bug.

So an alternative, I suppose, would be to typecast to ``MutinySessionImpl`, and get the connection from there.

blafond:
that would work but require accessing the `delegate` session to get the `delegate.getReactiveConnection()`

Would there be a future need/benefit to implementing a `ReactiveConnection` enum to apply to **Proxy, Batching and SqlClient** connection types and add a simple `ReactiveSession.getConnectionType()` method?

gavinking:
> that would work but require accessing the `delegate` session to get the `delegate.getReactiveConnection()`

Right. I think it's fine to expose the `delegate` on the `XxxxImpl` classes, just for this particular test.

> Would there be a future need/benefit to implementing a `ReactiveConnection` enum to apply to **Proxy, Batching and SqlClient** connection types and add a simple `ReactiveSession.getConnectionType()` method?

I mean that just seems like too much extra code, and I'm not sure how it would ever be useful to the _user_.

blafond:
removed the api changes and added a `getReactiveConnection()` method which pulls from the `delegate` reactive session.

DavideD:
I think you've squashed your fix over the last commit upstream

blafond:
yes :(  I'm going to revert and recommit correctly

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -700,6 +706,18 @@ public void testBatching(TestContext context) {
 		);
 	}
 
+	@Test
+	public void testBatchingConnection(TestContext context) {
```
DavideD, line 710: I don't think you need the parameter for this test because nothing is reactive

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -700,6 +706,18 @@ public void testBatching(TestContext context) {
...
+		Stage.Session session = getSessionFactory().openSession();
+		Assertions.assertThat(((StageSessionImpl)session).getReactiveConnection()).isInstanceOf( BatchingConnection.class );
+	}
+
+	@Test
+	public void testBatchingConnectionMutiny(TestContext context) {
```
DavideD, line 716: I don't think you need the parameter for this test because nothing is reactive

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -700,6 +706,18 @@ public void testBatching(TestContext context) {
...
+		Mutiny.Session session = getMutinySessionFactory().openSession();
+		Assertions.assertThat(((MutinySessionImpl)session).getReactiveConnection()).isInstanceOf( BatchingConnection.class );
```
DavideD, line 718: Not critical, but you can create an import static of `Assertions.assertThat`

DavideD (commented):
There are a couple of very minor things but otherwise I think we can merge this

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -700,6 +706,18 @@ public void testBatching(TestContext context) {
 		);
 	}
 
+	@Test
+	public void testBatchingConnection() {
+		Stage.Session session = getSessionFactory().openSession();
```
DavideD, line 711: Sorry, one thing I've just noticed... You are not using the method `openSession` in the superclass, meaning that the session doesn't get closed at the end of the test.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -700,6 +706,18 @@ public void testBatching(TestContext context) {
...
+		assertThat(((StageSessionImpl)session).getReactiveConnection()).isInstanceOf( BatchingConnection.class );
+	}
+
+	@Test
+	public void testBatchingConnectionMutiny() {
+		Mutiny.Session session = getMutinySessionFactory().openSession();
```
DavideD, line 717: You are not using the method `openMutinySession` in the superclass, meaning that the session doesn't get closed at the end of the test.

DavideD (changes requested):

gavinking:
LGTM now.

gavinking:
LGTM now.

blafond:
I added try/finally to insure session close is called.

PR #927, opened 2021-08-03
[#877] Refactor tests for batching connection

DavideD:
I think this version is more consistent with the other tests.
I also prefer, if possible,  to not have places where we ignore the returned value of `closeSession` (although in this case it was correct).


blafond (approved):
Looks good and I agree it keeps consistency within the tests.

DavideD:
Thanks!

Issue #901, opened 2021-07-22
UTCTest.testOffsetDateTime() test failure locally


blafond:
running this test locally by itself or with a local build throws the same exception:
```
Not equals : -05:00 != Z
java.lang.AssertionError: Not equals : -05:00 != Z
	at io.vertx.ext.unit.impl.TestContextImpl.reportAssertionError(TestContextImpl.java:362)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:258)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:245)
	at org.hibernate.reactive.UTCTest.lambda$testOffsetDateTime$5(UTCTest.java:118)
	at org.hibernate.reactive.UTCTest.lambda$testField$10(UTCTest.java:161)
```

It doesn't appear to fail in the CI builds.

Note I'm in the America/Chicago time-zone with an offset of ` -05:00`

blafond:
@DavideD I don't understand the use the constant: `ZoneOffset.UTC` in the assertion that's failing?

DavideD:
Which assertion is failing?

blafond:
line 118:  `context.assertEquals( entity.offsetDateTime.getOffset(), ZoneOffset.UTC );`

DavideD:
[In the configuration of the test](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UTCTest.java#L40) we set the property `hibernate.jdbc.time_zone` to `UTC`.

So the timezone of the date should be set to `UTC` before saving it to the database.
This is what I see when I debug.
You can add a break point here: https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/PreparedStatementAdaptor.java#L172

The `Calendar` parameter, used to set the timezone, should be set to `UTC`.

We also set the time zone to UTC when we read the value in ResultSetAdaptor (as far as I can see). So I'm not sure why you have a different time zone.

DavideD:
OK, I can recreate the error if I change the time zone on my machine. As an experiment, I think it will work if you set the time zone  on your pc to UTC

DavideD:
I think this is a bug in ORM. When it converts the Timestamp to the OffsetDateTime, it uses the system default time zone instead of the one we set with the property.

[...]

PR #930, opened 2021-08-04
[#901] removed failing assertion

blafond:
Fixes [#901](https://github.com/hibernate/hibernate-reactive/issues/901)

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UTCTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UTCTest.java
@@ -115,9 +115,8 @@ public class UTCTest extends BaseReactiveTest {
 				"offsetDateTimeType",
 				thing::getOffsetDateTime,
 				entity -> {
-					context.assertEquals( entity.offsetDateTime.getOffset(), ZoneOffset.UTC );
 					context.assertEquals( thing.offsetDateTime,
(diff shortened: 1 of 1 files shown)


DavideD:
I have this exception on my machine:
```
Not equals : Z != +01:00
java.lang.AssertionError: Not equals : Z != +01:00
	at io.vertx.ext.unit.impl.TestContextImpl.reportAssertionError(TestContextImpl.java:362)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:258)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:245)
	at org.hibernate.reactive.UTCTest.lambda$testOffsetDateTime$5(UTCTest.java:122)
```

blafond:
I changed my laptop time zone to different values and it does fail only on BTC (i.e. Z)

Fails in `ZoneOffset.equals()` on this check:  `return this.totalSeconds == ((ZoneOffset)obj).totalSeconds;`
where `this.totalSeconds` is **0** and `((ZoneOffset)obj).totalSeconds` is **3600**

DavideD:
As usual, dealing with dates is a pain. Honestly, I don't have a solution at the moment.
I'm fine with just removing the assertion for now.

blafond:
I removed the first assertion in that test.  The complexity of time zone rules seems to prevent a reasonable/working check on zone offset.

PR #933, opened 2021-08-06
Efficiency improvements: improve overhead of ContextKey comparisons

### LINKED ISSUE #932: Improve efficiency of org.hibernate.reactive.context.Context#Key
Some remarks from profiling:
- the use of `Objects.hash(type, id)` is hiding an array allocation; it's small but it's wasting ~3 GB/sec while doing 100000 R/s.
- might benefit from a constant (pre-computed hash), especially if we can reuse the same intance of Key more often.

While at it, it should really be made immutable.

Sanne:
Fixes #932


DavideD:
Thanks!

PR #934, opened 2021-08-08
Fix issues with `setFirstResult(0)` and MSSQL

### LINKED ISSUE #931: Parameter substitution error when limiting results on MS SQL Server and Hibernate Reactive 1.0.0.CR9
Hello, I am encountering an issue with MS SQL query parameters in 1.0.0.CR9 when using Paging or Ranges (SQL Exception, message='Incorrect syntax near '?'.'), possibly because the parameter is in the Select clause?

Stack: Quarkus 2.1.0.Final / Panache Reactive / Mutiny 4.1.2 / Hibernate Reactive 1.0.0.CR9 / Hibernate Core 5.5.5.Final

From my rest resource, running this:
`[PanacheRepository].find("firstName LIKE ?1").page(Page.of(25)).list();`

With SQL logging on in Hibernate, this is the produced SQL:
```
Hibernate:
    select
        top(?) request0_.id as id1_1_,
        request0_.caseId as caseid2_1_,
        request0_.caseNumber as casenumb3_1_,
        request0_.expires as expires4_1_,
        request0_.addressLine1 as addressl5_1_,
        request0_.addressLine2 as addressl6_1_,
        request0_.agency as agency7_1_,
        request0_.city as city8_1_,
        request0_.email as email9_1_,
        request0_.firstName as firstna10_1_,
        request0_.lastName as lastnam11_1_,
        request0_.phone as phone12_1_,
        request0_.relationship as relatio13_1_,
        request0_.state as state14_1_,
        request0_.zipcode as zipcode15_1_,
        request0_.submitted as submitt16_1_
    from
        Request request0_
    where
        request0_.firstName like @P1
```
Note the correct parameter substitution in the where clause, but substitution did not occur for the "top(?)" in the select clause.
If I remove the paging and re-run, the query executes successfully. 

Trace logging shown below, showing HQL and SQL. I would guess that Panache is using `setMaxResults()` on the query. Happy to open an issue with Quarkus or Mutiny if that's where the issue is, and thank you in advance!
```
2021-08-05 12:55:57,342 DEBUG [org.hib.hql.int.ast.QueryTranslatorImpl] (vert.x-eventloop-thread-1) HQL: FROM myproject.entities.Request WHERE requestor.firstName LIKE ?1
2021-08-05 12:55:57,342 DEBUG [org.hib.hql.int.ast.QueryTranslatorImpl] (vert.x-eventloop-thread-1) SQL: select request0_.id as id1_1_, request0_.caseId as caseid2_1_, request0_.caseNumber as casenumb3_1_, request0_.expires as expires4_1_, request0_.addressLine1 as addressl5_1_, request0_.addressLine2 as addressl6_1_, request0_.agency as agency7_1_, request0_.city as city8_1_, request0_.email as email9_1_, request0_.firstName as firstna10_1_, request0_.lastName as lastnam11_1_, request0_.phone as phone12_1_, request0_.relationship as relatio13_1_, request0_.state as state14_1_, request0_.zipcode as zipcode15_1_, request0_.submitted as submitt16_1_ from Request request0_ where request0_.firstName like ?
2021-08-05 12:55:57,342 DEBUG [org.hib.hql.int.ast.ErrorTracker] (vert.x-eventloop-thread-1) throwQueryException() : no errors
2021-08-05 12:55:57,342 TRACE [org.hib.rea.ses.imp.ReactiveHQLQueryPlan] (vert.x-eventloop-thread-1) Find: FROM myproject.entities.Request WHERE requestor.firstName LIKE ?1
2021-08-05 12:55:57,344 TRACE [org.hib.eng.spi.QueryParameters] (vert.x-eventloop-thread-1) Named parameters: {1=%j%}
2021-08-05 12:55:57,345 TRACE [org.hib.typ.des.sql.BasicBinder] (vert.x-eventloop-thread-1) binding parameter [2] as [VARCHAR] - [%j%]
2021-08-05 12:55:57,345 TRACE [org.hib.loa.Loader] (vert.x-eventloop-thread-1) Bound [3] parameters total
2021-08-05 12:55:57,355 DEBUG [org.hib.SQL] (vert.x-eventloop-thread-1)
    select
        top(?) request0_.id as id1_1_,
        request0_.caseId as caseid2_1_,
        request0_.caseNumber as casenumb3_1_,
        request0_.expires as expires4_1_,
        request0_.addressLine1 as addressl5_1_,
        request0_.addressLine2 as addressl6_1_,
        request0_.agency as agency7_1_,
        request0_.city as city8_1_,
        request0_.email as email9_1_,
        request0_.firstName as firstna10_1_,
        request0_.lastName as lastnam11_1_,
        request0_.phone as phone12_1_,
        request0_.relationship as relatio13_1_,
        request0_.state as state14_1_,
        request0_.zipcode as zipcode15_1_,
        request0_.submitted as submitt16_1_
    from
        Request request0_
    where
        request0_.firstName like @P1
2021-08-05 12:55:57,358 ERROR [org.hib.rea.errors] (vert.x-eventloop-thread-1) HR000057: Failed to execute statement [$1select request0_.id as id1_1_, request0_.caseId as caseid2_1_, request0_.caseNumber as casenumb3_1_, request0_.expires as expires4_1_, request0_.addressLine1 as addressl5_1_, request0_.addressLine2 as addressl6_1_, request0_.agency as agency7_1_, request0_.city as city8_1_, request0_.email as email9_1_, request0_.firstName as firstna10_1_, request0_.lastName as lastnam11_1_, request0_.phone as phone12_1_, request0_.relationship as relatio13_1_, request0_.state as state14_1_, request0_.zipcode as zipcode15_1_, request0_.submitted as submitt16_1_ from Request request0_ where request0_.firstName like @P1]: $2could not execute query: java.util.concurrent.CompletionException: io.vertx.mssqlclient.MSSQLException: {number=102, state=1, severity=15, message='Incorrect syntax near '?'.', serverName='aa3ae4a0fd3c', lineNumber=1, additional=[io.vertx.mssqlclient.MSSQLException: {number=8180, state=1, severity=16, message='Statement(s) could not be prepared.', serverName='aa3ae4a0fd3c', lineNumber=1}]}
```

DavideD:
Fixes #931 

Not an elegant solution but should fix the issues


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SQLServerParameters.java
```diff
@@ -34,24 +36,48 @@ public String process(String sql, int parameterCount) {
...
             return sqlProcessed;
         }
-
+        // Different Hibernate ORM versions may use different lowecase/uppercase letters
+        if ( sql.toLowerCase( Locale.ROOT ).startsWith( "select top(?)" ) ) {
+            // 13 is the length og the string "select top(?)"
+            String sqlProcessed = "select top(@P" + index + ")" + sql.substring( 13 );
+            shiftValues( parameterArray );
+            return sqlProcessed;
+        }
```
gavinking, line 68: It's unfortunate that ORM sometimes generates such ugly SQL. This limit handler stuff is such a mess in 5. I almost feel like back porting the cleanups I did for H6. :-/

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SQLServerParameters.java
```diff
@@ -34,24 +36,48 @@ public String process(String sql, int parameterCount) {
...
+        // See SQLServer2012LimitHandler#getOffsetFetch
+        String offsetQueryString;
+        int pos = sql.indexOf( " offset ? rows fetch next ? rows only" );
+        if ( pos > -1 ) {
+            offsetQueryString = " offset @P" + index++;
+        }
+        else {
+            pos = sql.indexOf( " offset 0 rows fetch next ? rows only" );
+            offsetQueryString = " offset 0";
+        }
```
gavinking, line 54: @DavideD wouldn't it be good enough to just match `" offset "` and then grab the next two characters?

DavideD: Nice idea.
It bothers me a bit that we check such a short string but I've updated the code anyway

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SQLServerParameters.java
DavideD: It would be nice, but I would wait and see if other more complicated use cases pop up before refactoring 5.
Unless is something relatively easy to do.

gavinking:
@Davide, I'm inclined to say: simplify this patch + just not support SQL Server versions earlier than 2012.

Instead of maintaining that rather ugly, (and also currently untested) code for the `SQLServer2005LimitHandler`.

gavinking:
> Sounds good, I will update the pr in a bit

👍

DavideD:
Wait ... this issue doesn't seem related to the version of the limit handler. We are  using `SQLServer2012LimitHandler` and the test fails anyway. Maybe I didn't understand what you are suggesting?

gavinking:
I mean the specific bit of the fix which handles `select top(?)`. `SQLServer2012LimitHandler` doesn't use `top`.

DavideD:
> I mean the specific bit of the fix which handles `select top(?)`. `SQLServer2012LimitHandler` doesn't use `top`.

I've checked and `SQLServer2012LimitHandler` still uses `top` when the query doesn't have a order by clause.
I've rebased and updated the comment in my previous pr

DavideD:
I'm going to merge this one. We can revert it or fix bugs later

PR #942, opened 2021-08-25
fix #928 problem with JsonType converter

### LINKED ISSUE #928: Throw java.lang.IllegalArgumentException when merge null value
When merge entity into database
```
public class ShippingCarrier extends BaseOwnerWithParentEntity {

    @EqualsAndHashCode.Exclude
    @Convert(converter = MultiLocaleConverter.class)
    @Column(name = "NAME")
    private MultiLocaleValue<String> name;
}
```

and Converter is 
```
@Converter
public class MultiLocaleConverter implements AttributeConverter<MultiLocaleValue<String>, JsonObject> {

```

when insert
```
session.merge(entity).flatMap(ent -> session.flush().then(Mono.just(ent))));
```
then it will throw exception
```
		at java.sql/java.sql.JDBCType.valueOf(JDBCType.java:258)
		at org.hibernate.reactive.adaptor.impl.PreparedStatementAdaptor.setNull(PreparedStatementAdaptor.java:97)
		at org.hibernate.type.descriptor.sql.BasicBinder.bind(BasicBinder.java:60)
		at org.hibernate.type.descriptor.converter.AttributeConverterSqlTypeDescriptorAdapter$1.bind(AttributeConverterSqlTypeDescriptorAdapter.java:88)
		at org.hibernate.type.AbstractStandardBasicType.nullSafeSet(AbstractStandardBasicType.java:276)
		at org.hibernate.type.AbstractStandardBasicType.nullSafeSet(AbstractStandardBasicType.java:271)
		at org.hibernate.type.AbstractSingleColumnStandardBasicType.nullSafeSet(A
```

I just compare the CR6(OK) and CR9 (NG)

org.hibernate.reactive.adaptor.impl.PreparedStatementAdaptor

CR6
```
@Override
public void setNull(int parameterIndex, int sqlType)
 { put( parameterIndex, null ); }

 ```

CR9

 
```
@Override
public void setNull(int parameterIndex, int sqlType)


 { put( parameterIndex, JDBCType.valueOf(sqlType) ); }
```
 

it will cause exception, because sqltype is object id

gavinking:


gavinking:
Well *that's* weird: the new test works fine on MySQL but not on Maria...

gavinking:
Well, I dunno ... there seems to be some subtle difference in the behavior of the SQL client on Maria vs MySQL but it's not obvious what's going on and I can't be bothered tracking it down right now. (The client returns a `String` instead of a `JsonObject`.)

I've disabled the new test on Maria instead.

PR #948, opened 2021-08-31
Take advantage of the Vert.x Context API improvements

Sanne:
Follows up on https://github.com/eclipse-vertx/vert.x/pull/4073

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
@@ -32,15 +32,11 @@ public class VertxContext implements Context, ServiceRegistryAwareService {
         vertxInstance = serviceRegistry.getService( VertxInstance.class );
     }
 
-    static ContextInternal currentContext() {
-        return (ContextInternal) Vertx.currentContext();
-    }
-
     @Override
     public <T> void put(Key<T> key, T instance) {
-        final ContextInternal context = currentContext();
+        final io.vertx.core.Context context = Vertx.currentContext();
         if ( context != null ) {
-            context.localContextData().put( key, instance );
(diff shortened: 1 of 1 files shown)


gavinking:
ah yes, i had forgotten about this

gavinking (approved):

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/948

Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -50,9 +46,9 @@ static ContextInternal currentContext() {
     @Override
     @SuppressWarnings("unchecked")
     public <T> T get(Key<T> key) {
-        final ContextInternal context = currentContext();
+        final io.vertx.core.Context context = Vertx.currentContext();
         if ( context != null ) {
-            return (T) context.localContextData().get( key );
+            return (T) context.get( key );
```
gavinking, line 51: @Sanne why did you change this to use the main context instead of the local context?

Isn't that wrong for what we're using it for?

Sanne: yes it's wrong - it wasn't intentional, I wasn't aware of the "global context" and all tests passed fine.

Ironically my benchmarks also showed no issue, as we don't use the APIs which use the context. Sorry :(

gavinking: @Sanne, we all share in the blame for this one. I approved it, Davide merged it. I *certainly* should have noticed but I just didn't read the diff carefully enough.

It's embarrassing but that's a think that can happen with this stuff that's hard to test. Hopefully we can figure out some way to test it.

PR #949, opened 2021-09-01
Track latest available ORM 5.6.x

DavideD:
The purpose of the build with 5 is to track the latest available ORM 5 (that's why it was an interval). Any particular reason for it to be set to 5.6.0.Beta1? I'm assuming @gbadner needed that version while working on the schema issue.

@Sanne Do we care about other minors (like 5.5.x or 5.4.x)?

### DIFF
--- a/.github/workflows/tracking-orm-5.build.yml
+++ b/.github/workflows/tracking-orm-5.build.yml
@@ -1,3 +1,5 @@
+# Run the build using the latest ORM 5.x snapshots
+# so that we can spot integration issues early
 name: Latest ORM 5.x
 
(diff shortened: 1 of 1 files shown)


gavinking:
👍 right, fine, I wasn't sure about that.

DavideD:
I've added a comment so that it is clearer

DavideD:
> @Sanne Do we care about other minors (like 5.5.x or 5.4.x)?

I've found the answer in a previous PR: https://github.com/hibernate/hibernate-reactive/pull/903#issuecomment-909055382

Issue #951, opened 2021-09-03
merging a @ManyToOne association can call non-reactive code


nivinps:
Hi, 
I need to update 'erp_campus_id' in 'erp_academic_year_detail' table. The same code will works If I used normal hibernate. But In hibernate-reactive it throws an exception. I am using 1.0.0.CR9 version.

```java
AcademicYearDetaisDBO dbo = sessionFactory.withSession(session -> session.createQuery("select dbo from AcademicYearDetaisDBO dbo where dbo.id=386",AcademicYearDetaisDBO.class).getSingleResult()).await().indefinitely();

dbo.setRecordStatus('A');
System.out.println(dbo.getCampusId().getId());//for example here campus id is 11
CampusDBO campusDBO = new CampusDBO();
campusDBO.setId(5);
dbo.setCampusId(campusDBO); // need to update as 5

sessionFactory.withTransaction((session, transaction) -> session.merge(dbo)).await().indefinitely(); // here it will throw an exception
```

```java
@Getter
@Setter
@Entity
@Table(name = "erp_academic_year_detail")
public class AcademicYearDetaisDBO implements Serializable {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    @Column(name = "erp_academic_year_detail_id")
    private Integer id;

    @ManyToOne
    @JoinColumn(name = "erp_campus_id")
    private CampusDBO campusId;

    @Column(name = "record_status")
    private char recordStatus;

    @Column(name = "created_users_id")
    private Integer createdUsersId;

    @Column(name = "modified_users_id")
    private Integer modifiedUsersId;
}
```

```java
@Entity
@Getter
@Setter
@Table(name="erp_campus")
public class CampusDBO implements Serializable {
        @Id
        @GeneratedValue(strategy = GenerationType.IDENTITY)
        @Column(name="erp_campus_id")
        private  Integer id;

        @Column(name="campus_name")
        private String campusName;
}
```

Exception:

```
javax.persistence.PersistenceException: org.hibernate.HibernateException: java.util.concurrent.CompletionException: org.hibernate.reactive.event.impl.UnexpectedAccessToTheDatabase: Unexpected access to the database
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:154) ~[hibernate-core-5.5.7.Final.jar:5.5.7.Final]
[...]

PR #952, opened 2021-09-03
Testcase for merge with eager association

DavideD:
Testcase for #951

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ManyToOneMergeTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ManyToOneMergeTest.java
@@ -1,0 +1,167 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


gavinking:
I'm looking at it now.

gavinking:
Hrm. Well, I see what's happening.

Basically, Hibernate is merging the properties of the `CampusDBO`, and finds the reference to the (proxy) `AcademicYearDetailsDBO`. Now, it wants to swap that reference for a ref to a *managed* `AcademicYearDetailsDBO`. Since the association is nonlazy, it loads it. I can't say that's wrong. It would probably be "surprising" for a non-lazy association to hold a ref to a proxy.

So the *problem* is that the method that does the swapping is nonreactive, and it might be hard to make it reactive.

So this is a bug, and one that might take some work to fix.

On the other hand I think we dealt with a very similar issue in #506 and there is code at `ReactiveResultSetProcessor:75` which does something quite similar to what I guess we must need to do here. So perhaps it's not so bad.

gavinking:
Closing, superseded by #954.

Issue #950, opened 2021-09-01
Consider forcing Uni return types on all Session opening methods


Sanne:
I've got burned once again by chaining the wrong style of Session opening.

This is what I had as a general helper:

     @Inject
    Mutiny.SessionFactory sf;

    public <T> Uni<T> inSession(Function<Mutiny.Session, Uni<T>> work) {
        final Mutiny.Session session = sf.openSession();
        return sf.withSession( work ).eventually( session::close );
    }

(This is very similar to the `withSession` helper offered on `MutinySessionImpl`, but I was needing to intentionally side-step the use of the `org.hibernate.reactive.context.Context` - details not relevant now)
    
When I'm passing as `work` argument a trivial work which does a single operation on the session, this works as expected.

But if I use such an helper method with a slightly less trivial work, such as needing to load more than one entity, the second load operation is likely going to hit the infamous error we already know quite well:

```
Caused by: java.lang.IllegalStateException: Session/EntityManager is closed
	at org.hibernate.internal.AbstractSharedSessionContract.checkOpen(AbstractSharedSessionContract.java:393)
	at org.hibernate.engine.spi.SharedSessionContractImplementor.checkOpen(SharedSessionContractImplementor.java:148)
	at org.hibernate.reactive.session.impl.ReactiveSessionImpl.checkOpen(ReactiveSessionImpl.java:1556)
	at org.hibernate.internal.AbstractSharedSessionContract.checkOpenOrWaitingForAutoClose(AbstractSharedSessionContract.java:399)
```

In this case, I am sure that this was all run on the same vert.x context (and the same thread); with @DavideD 's help we figured out the problem is that the `Mutiny.Session` instance opened this way is not fit to be used more than once, because the composition of multiple operations on it can't guarantee that the session proxy is being initialized once and atomically.

I would propose to remove the `Session openSession();` method and its siblings dealing with tenant-id and stateless sessions, to be replaced with methods which return the `Uni<Session>`, `Uni<StatelessSession>`, etc... 

Impact:
 - it would seem the Hibernate Reactive code would get simpler
 - most Quarkus integration code would also benefit from it
 - Panache Reactive would also be simpler, as it is more common to require a `Uni<Session>` - the only exception being the `boolean isPersistent(Object entity) ` method which Panache Reactive exposes, but we can either deal with that by using the context, or update the signature to also become a reactive method.

Sanne:
cc/ @gavinking I'd be keen to hear you opinion on this one.

[...]

PR #953, opened 2021-09-03
Amend ProxyConnection to deal with multiple requests being scheduled

Sanne:
Draft, to think about it.

This would avoid issues when issuing muliple requests to a Session which haven't been strictly chained: for example this would allow to "send" a second SQL to load a second entity while the first load hasn't been completed yet.


Sanne:
After talking with @gavinking , we agreed it would be more compelling to avoid the problem entirely by removing the needd to defer acquiring the connection: let's implement #950 .

However since the exception being thrown here was our only "safeguard" allowing to detect illegal usage patterns, we also need to improve our ability to detect such things, and throw a more helpful error.

PR #963, opened 2021-09-16
Upgrade JDK 18 tests to JDK 18-ea+14

yrodiere:


yrodiere:
Actually, forget about JDK 17 GA, it's not available on Eclipse Adoptium yet.

Issue #964, opened 2021-09-16
Remove IBM's asyncutil library dependency


DavideD:
This: https://github.com/hibernate/hibernate-reactive/blob/297ddc867e0d3e21b01c2f7e7b8a12e996ce041b/hibernate-reactive-core/build.gradle#L27

DavideD:
Done

PR #965, opened 2021-09-17
Remove IBM asyncutil library

DavideD:
#964

I've copied the classes we need and made some changes to reuse some methods already in `CompletionStages` (instead of copying additional classes from the library).

Considering that I've copied classes from a library licensed under Apache 2.0,
is there something specific I need to pay attention to when I copy the files?


Sanne:
> Considering that I've copied classes from a library licensed under Apache 2.0,
is there something specific I need to pay attention to when I copy the files?

It's not required by the license but I'd consider it "good manners" to add a comment on the files attributing credit to the original project (and authors if you know the names).

DavideD:
I've updated the javadocs and added a package-info. It should be good now

Sanne (approved):

gavinking:
Thanks @DavideD this is good 👍

PR #967, opened 2021-09-21
912 : SchemaMigrator/SchemaValidator support for MS SQL Server

gbadner:
https://github.com/hibernate/hibernate-reactive/issues/912

Includes https://github.com/hibernate/hibernate-reactive/pull/966

Requires ORM PR: https://github.com/hibernate/hibernate-orm/pull/4235

PR #970, opened 2021-09-22
MSSQL Schema update and validation

### LINKED ISSUE #912: SchemaMigrator/SchemaValidator support for MS SQL Server
Follows #536

See #903 for an example with PostgreSQL

DavideD:
Fixes #912 
A test for the following pr: https://github.com/hibernate/hibernate-reactive/pull/967

I've rebased it to the latest changes and it's using the available HIbernate ORM (5.6.0.Beta2)

Waiting for [this PR](https://github.com/hibernate/hibernate-orm/pull/4235) to be merged and released

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
@@ -99,11 +99,11 @@ public abstract class SchemaUpdateSqlServerTestBase extends BaseReactiveTest {
 
 	@Test
 	public void testValidationSucceed(TestContext context) {
-		Configuration createHbm2ddlConf = constructConfiguration( "validate" );
-		createHbm2ddlConf.addAnnotatedClass( ASimpleFirst.class );
-		createHbm2ddlConf.addAnnotatedClass( AOther.class );
+		Configuration configuration = constructConfiguration( "validate" );
+		configuration.addAnnotatedClass( ASimpleFirst.class );
+		configuration.addAnnotatedClass( AOther.class );
 
-		test( context, setupSessionFactory( createHbm2ddlConf ) );
+		test( context, setupSessionFactory( configuration ) );
 	}
 
 	//TODO: I'm just checking that the validation fails because the table is missing, but we need more tests to check that
@@ -111,10 +111,10 @@ public abstract class SchemaUpdateSqlServerTestBase extends BaseReactiveTest {
 	//      actually checks).
 	@Test
 	public void testValidationFails(TestContext context) {
-		Configuration createHbm2ddlConf = constructConfiguration( "validate" );
-		createHbm2ddlConf.addAnnotatedClass( AAnother.class );
+		Configuration configuration = constructConfiguration( "validate" );
+		configuration.addAnnotatedClass( AAnother.class );
 
-		test( context, setupSessionFactory( createHbm2ddlConf )
+		test( context, setupSessionFactory( configuration )
 				.handle( (unused, throwable) -> {
 					context.assertNotNull( throwable );
 					context.assertEquals( throwable.getClass(), SchemaManagementException.class );
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
@@ -7,6 +7,7 @@ package org.hibernate.reactive;
 
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.concurrent.CompletionStage;
 import javax.persistence.CascadeType;
 import javax.persistence.Entity;
 import javax.persistence.ForeignKey;
@@ -80,8 +81,24 @@ public abstract class SchemaUpdateSqlServerTestBase extends BaseReactiveTest {
 		createHbm2ddlConf.addAnnotatedClass( ASimpleFirst.class );
 		createHbm2ddlConf.addAnnotatedClass( AOther.class );
 
-		test( context, setupSessionFactory( createHbm2ddlConf )
-				.thenCompose( v -> factoryManager.stop() ) );
+		test( context, dropSequenceIfExists( createHbm2ddlConf )
+				.thenCompose( ignore -> setupSessionFactory( createHbm2ddlConf )
+				.thenCompose( v -> factoryManager.stop() ) ) );
+	}
+
+	// See HHH-14835: Vert.x throws an exception when the catalog is specified.
+	// Because it happens during schema creation, the error is ignored and the build won't fail
+	// if one of the previous tests has already created the sequence.
+	// This method make sure that the sequence is deleted if it exists, so that these tests
+	// fail consistently when the wrong ORM version is used.
+	private CompletionStage<Void> dropSequenceIfExists(Configuration createHbm2ddlConf) {
+		return setupSessionFactory( createHbm2ddlConf )
+				.thenCompose( v -> getSessionFactory()
+						.withTransaction( (session, transaction) -> session
+								.createNativeQuery( "drop sequence if exists dbo.hibernate_sequence" )
+								.executeUpdate() ) )
+				.handle( (res, err) -> null )
+				.thenCompose( v -> factoryManager.stop() );
 	}
 
 	@After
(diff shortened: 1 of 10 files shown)


DavideD:
@gbadner, @blafond I was expecting this PR to fail but it seems to work fine. What am I missing?

gbadner:
I'm not sure how it could be passing?!? I'm looking into it...

gbadner:
@DavideD, take a look at the standard output for the SQL Server test:
hibernate-reactive-core/build/reports/tests/test/classes/org.hibernate.reactive.SchemaUpdateSqlServerTestBase$IndividuallySchemaUpdateSqlServerTestBase.html

The `drop sequence` statement fails:
```
?[33mWARN?[m ?[34mReactiveGenerationTarget?[m [vert.x-eventloop-thread-0] HR000021: DDL command failed [io.vertx.mssqlclient.MSSQLException: {number=166, state=1, severity=15, message=''DROP SEQUENCE' does not allow specifying the database name as a prefix to the object name.', serverName='8ec40056a401', lineNumber=1, additional=[io.vertx.mssqlclient.MSSQLException: {number=8180, state=1, severity=16, message='Statement(s) could not be prepared.', serverName='8ec40056a401', lineNumber=1}]}]
```
Later in the output, you'll see that the `create sequence` statement fails for the same reason.

Later in the test, you'll see that `select next value for master.dbo.hibernate_sequence` statements succeed.
(you can actually search for that string in the output because of all the embedded "?[34m".)

Apparently, the `hibernate_sequence` is left over from a previous test, and that is what is being used.

Using a local ORM 5.6.0-SNAPSHOT, these messages are gone, and the drop/create sequence statements succeed.

DavideD:
Thanks @gbadner,  makes sense.

gbadner:
Supersedes https://github.com/hibernate/hibernate-reactive/pull/967.

gbadner:
We need to wait for https://github.com/hibernate/hibernate-orm/pull/4235 to get merged and released before this PR can be merged.

gbadner:
> Using a local ORM 5.6.0-SNAPSHOT, these messages are gone, and the drop/create sequence statements succeed.

To clarify, after publishing the hibernate-core jar from https://github.com/hibernate/hibernate-orm/pull/4235 (which is not merged yet) to a local repository, and setting gradle.properties to use that jar, the drop/create sequence statements succeed.

gbadner:
Pushed a [commit](https://github.com/hibernate/hibernate-reactive/pull/970/commits/22c213e56ae80cfd9c072e8d25e1bf3c651ca5b3) to fix inconsistent test arguments.

gbadner:
@DavideD, is the DB2 test failure expected?

DavideD:
I've added a commit that should make the tests fail consistently.
It will delete the sequence if it exists before running the creation of the schema.
@gbadner or @blafond could you check that it works when using the right ORM snapshot, please?

gbadner:
@DavideD, I pulled your commit, confirmed the tests fail using 5.6.0.Beta2, and confirmed the tests pass using https://github.com/hibernate/hibernate-orm/pull/4235.

FWIW, specifying the catalog when creating a sequence on ORM worked fine. Vert.x throws an exception when the catalog is specified.

SQL Server documentation of the `create sequence` syntax only allows the schema, not the catalog:
 https://docs.microsoft.com/en-us/sql/t-sql/statements/create-sequence-transact-sql?view=sql-server-ver15&viewFallbackFrom=sql-server-ver12 . This is mentioned in ORM code that that was changed.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
```diff
@@ -0,0 +1,357 @@
...
+			configuration.setProperty( Settings.HBM2DDL_JDBC_METADATA_EXTRACTOR_STRATEGY, GROUPED.toString() );
+			return configuration;
+		}
+	}
+
+	protected Configuration constructConfiguration(String hbm2DdlOption) {
+		Configuration configuration = constructConfiguration();
+		configuration.setProperty( Settings.HBM2DDL_AUTO, hbm2DdlOption );
+		// DEFAULT_CATALOG needs to be set to something other than an empty string;
+		// otherwise, NullPointerException is thrown.
```
DavideD, line 67: I've just noticed this, why do we have a NullPointerException if we don't set the catalog?

DavideD: It seems that if we don't set the catalog name and solve the issue with the NullPointerException, we can merge this without the fix for ORM and treat the Catalog name as a separate issue.

I've added a commit: @gbadner or @blafond could you have a look at it and tell me if it makes sense?

blafond:
tested successfully locally with default ORM 5.6.0.Beta2 as well as my local 5.6.0-SNAPSHOT

gbadner:
@Davide, I agree that what you did will make the test pass. I'm not sure if it will always be correct though. The problem is that ORM doesn't have a way to find out the catalog name (which is the database name) unless Hibernate Reactive provides it.

I think it would be fine to deal with the catalog name separately.

gbadner:
@DavideD, I have some ideas about how the catalog name might be resolved. Should I go ahead and create the follow-on issue for catalog, or would like to do that?

DavideD:
> @DavideD, I have some ideas about how the catalog name might be resolved. Should I go ahead and create the follow-on issue for catalog, or would like to do that?

Yes, please. Create a separate issue. I've tried to play a bit with it but could figure out what to do.
It seems the property also causes issues with MySQL and MariaDB

DavideD:
> @davide, I agree that what you did will make the test pass. I'm not sure if it will always be correct though. The problem is that ORM doesn't have a way to find out the catalog name (which is the database name) unless Hibernate Reactive provides it.

Sure but I see this as a separate issue. If I'm not mistaken, everything will work for users that don't set the catalog name

gbadner:
@DavideD, I'm really not sure if it will always work. 

SQL Server's default catalog name is "master". I've read that it is possible to change it. I am not sure if/how that would impact what gets returned by information_schema views.

I have not figured out how to change it with `podman`, so I haven't tested it.

DavideD:
We can just test with  the default name and make sure that it works with that

gbadner:
Your change will make the queries not take the catalog name into account at all. If it is possible for information_schema to have more than one catalog, then query results that ORM relies on might not be correct.

This may not be an issue; I don't know...

DavideD:
What I mean is that the issue with the catalog is not just for schema validation/update. It's something that won't work in other cases as well. I would expect this test to pass with MSSQL: https://github.com/DavideD/hibernate-reactive/commit/5b979c267cfe6b46a8e9b4ff579b76c2f22d80df

gbadner:
@DavideD, I think it really only matters for SchemaMigrator and SchemaValidator.

gbadner:
Yes, the catalog name will be used if provided for SchemaExport, but it is not needed SchemaExport.

gbadner:
I suppose it might be needed if the catalog is a non-default, but honestly, I don't know. This is not an issue for ORM because ORM can get the information it needs.

DavideD:
I've refactored the code and added back the test with the default catalog name.
The latest changes in the PR test the following cases:

1. Property `hibernate.default_catalog` is not set and MSSQL uses the default catalog name -> Works
2. Property `hibernate.default_catalog` is set to `master` (the default catalog name for MSSQL) -> requires [this fix for ORM](https://github.com/hibernate/hibernate-orm/pull/4235)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateSqlServerTestBase.java
```diff
@@ -0,0 +1,403 @@
...
+		@Id
+		private int id1;
+
+		@Id
+		private String id2;
+
+		private String anotherString;
+	}
+
+	public static class AOtherId implements Serializable {
```
DavideD, line 373: @gbadner I've just noticed this class but I don't think we are using it. Is it necessary for the test or is this a leftover?

gbadner: @DavideD, yes, it is needed. It is an `IdClass` used by `AOther`.

DavideD: Ah yes, sorry

DavideD:
Merged!

gavinking:
Excellent!

PR #971, opened 2021-09-22
Switch from Java 8 to Java 11 in the README

### LINKED ISSUE #969: Switch requirements to Java 11
At least in the documentation, we can still test with Java 8 for now

DavideD:
We will continue to test with Java 8 for the time being
but at least we don't commit to it for the future.

And we are aligned to the base version used by the products.

Fixes #969

I will update the website before the release

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Learn more at <http://hibernate.org/reactive>.
 
 Hibernate Reactive has been tested with:
 
-- Java 8
+- Java 11, 16 and 17
 - PostgreSQL 13
 - MySQL 8
 - MariaDB 10


README.md
```diff
@@ -28,7 +28,7 @@ Learn more at <http://hibernate.org/reactive>.
 
 Hibernate Reactive has been tested with:
 
-- Java 8
+- Java 11
```
Sanne, line 31: since the wording is "has been tested with" , shouldn't we list Java 16 and 17 as well?

Sanne (approved):

README.md
Sanne: @DavideD up

DavideD: Added 11,16 and 17

DavideD:
Merged!

PR #973, opened 2021-09-23
Have Uni return types on all Session opening methods

DavideD:
**_DON'T MERGE_**

@Sanne,  I've squashed the commits together.

I've added a commit that removes `ProxyConnection` because I don't think we need it now but it's optional.

\cc @gavinking


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1515,62 +1516,15 @@
...
-		 * <p>
-		 * The underlying database connection is obtained lazily when
-		 * the returned {@link Session} needs to access the database.
-		 * <p>
-		 * The client must explicitly close the session by calling
-		 * {@link Session#close()}.
-		 *
-		 * @see #withSession(Function)
-		 */
-		Session openSession();
```
Sanne, line 1531: We'll need to restore the javadoc

DavideD: Yes, I'm going to do it today

DavideD:
Replaced by https://github.com/hibernate/hibernate-reactive/pull/983

Issue #936, opened 2021-08-10
SchemaMigrator/SchemaValidator - add tests for corner cases


gbadner:
There are some corner cases that are covered by ORM tests. They should be migrated to Reactive to make sure they work properly.

Here are some ORM `SchemaMigrator` tests that should be reviewed for possible inclusion in Hibernate Reactive:
```
SchemaUpdateTableBackedSequenceTest
CrossSchemaForeignKeyGenerationTest
UniqueConstraintDropTest
MySQLDropConstraintThrowsExceptionTest
SchemaMigrationTargetScriptCreationTest
SchemaMigrationToOutputScriptTest     -> @RequiresDialect(H2Dialect.class)
SchemaMigratorHaltOnErrorTest        -> skip DB2 & Derby
SchemaUpdateProceedOnErrorTest
SchemaUpdateSchemaNameTest     ->  @RequiresDialect(MySQL5Dialect.class)
```
Here are some ORM `SchemaValidator` tests that should be reviewed for possible inclusion in Hibernate Reactive:
```
LongVarcharValidationTest
NumericValidationTest
GroupedSchemaValidatorImplTest       -> @RequiresDialect(H2Dialect.class)
IndividuallySchemaValidatorImplConnectionTest       -> @RequiresDialect(H2Dialect.class)
IndividuallySchemaValidatorImplTest       -> @RequiresDialect(H2Dialect.class)
SchemaToolTransactionHandlingTest
```

gbadner:
I don't think they all need to be added. I'll take a look at these and see which ones we need.

DavideD:
I think we cover these cases, we will fix errors when they present themselves

PR #977, opened 2021-09-28
[936] SchemaMigrator/SchemaValidator tests for corner cases

blafond:
Addresses [#936](https://github.com/hibernate/hibernate-reactive/issues/936)

Test cases for corner cases of schema update, migration and validation

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
@@ -1,0 +1,185 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.schema;
+
+import static org.hibernate.tool.schema.JdbcMetadaAccessStrategy.GROUPED;
+import static org.hibernate.tool.schema.JdbcMetadaAccessStrategy.INDIVIDUALLY;
+
+import java.math.BigDecimal;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.BaseReactiveTest;
+import org.hibernate.reactive.containers.DatabaseConfiguration;
+import org.hibernate.reactive.provider.Settings;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import io.vertx.ext.unit.TestContext;
+
+public abstract class NumericValidationTest extends BaseReactiveTest {
+
+	public static class IndividuallyNumericValidationTestBase extends NumericValidationTest {
+
+		@Override
+		protected Configuration constructConfiguration(String hbm2DdlOption) {
+			final Configuration configuration = super.constructConfiguration( hbm2DdlOption );
+			configuration.setProperty( Settings.HBM2DDL_JDBC_METADATA_EXTRACTOR_STRATEGY, INDIVIDUALLY.toString() );
+			return configuration;
+		}
+	}
+
+	public static class GroupedNumericValidationTestBase extends NumericValidationTest {
+
+		@Override
+		protected Configuration constructConfiguration(String hbm2DdlOption) {
+			final Configuration configuration = super.constructConfiguration( hbm2DdlOption );
+			configuration.setProperty( Settings.HBM2DDL_JDBC_METADATA_EXTRACTOR_STRATEGY, GROUPED.toString() );
+			return configuration;
+		}
+	}
+
+	protected Configuration constructConfiguration(String hbm2DdlOption) {
+		Configuration configuration = constructConfiguration();
+		configuration.setProperty( Settings.HBM2DDL_AUTO, hbm2DdlOption );
+		configuration.setProperty( Settings.DEFAULT_CATALOG, "hreact" );
+		return configuration;
+	}
+
+	@Before
+	@Override
+	public void before(TestContext context) {
+
+		Configuration configuration = constructConfiguration( "create" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		test( context, setupSessionFactory( configuration )
+				.thenCompose( v -> factoryManager.stop() ) );
+	}
+
+	@After
+	@Override
+	public void after(TestContext context) {
+
+		final Configuration configuration = constructConfiguration( "drop" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		test( context, factoryManager.stop()
+				.thenCompose( v -> setupSessionFactory( configuration ) )
+				.thenCompose( v -> factoryManager.stop() ) );
+	}
+
+	@Test
+	public void testValidationSucceed(TestContext context) {
+		Configuration configuration = constructConfiguration( "validate" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		test(
+				context,
+				setupSessionFactory( configuration )
+						.thenCompose( v -> getSessionFactory()
+								.withTransaction(
+										(session, t) -> session.createQuery( "FROM TestEntity", TestEntity.class )
+												.getResultList() ) )
+						.thenAccept( results -> context.assertTrue( results.isEmpty() ) )
+		);
+	}
+
+	@Test
+	public void testDataTypePostgreSql(TestContext context) {
+		final TestEntity testEntity = new TestEntity();
+		testEntity.id = 9;
+		testEntity.number = new BigDecimal( 3.14159 );
+
+		String columnTypeQuery =
+				"select data_type from information_schema.columns " +
+						"where table_name = 'testentity' and column_name = 'numbervalue';";
+
+		final Configuration configuration = constructConfiguration( "update" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		if ( DatabaseConfiguration.dbType() != DatabaseConfiguration.DBType.POSTGRESQL ) {
+			return;
+		}
+		test(
+				context,
+				setupSessionFactory( configuration )
+						.thenCompose( v -> getSessionFactory()
+								.withTransaction( (session, t) -> session.persist( testEntity ) ) )
+						.thenCompose( v1 -> openSession()
+								.thenCompose( s -> s
+										.find( TestEntity.class, testEntity.id )
+										.thenAccept( result -> {
+													context.assertNotNull( result );
+													context.assertEquals( testEntity.number, testEntity.number );
+												}
+										).thenCompose( v -> s.createNativeQuery( columnTypeQuery, String.class )
+												.getResultList()
+												.thenAccept( list -> {
+													context.assertEquals( 1, list.size() );
+													context.assertTrue( list.get( 0 ).equals( "numeric" ) );
+												} )
+										)
+								)
+						)
+		);
+	}
+
+	@Test
+	public void testDataTypeDB2(TestContext context) {
+		TestEntity testEntity = new TestEntity();
+		testEntity.id = 9;
+		testEntity.number = new BigDecimal( 3.14159);
+
+		String columnTypeQuery =
+				"SELECT TYPENAME\n" +
+						"FROM SYSCAT.COLUMNS where TABSCHEMA = 'HREACT' and TABNAME = 'TESTENTITY' and COLNAME = 'NUMBERVALUE'\n" +
+						"ORDER BY TABSCHEMA, TABNAME, COLNAME;";
+
+		final Configuration configuration = constructConfiguration( "validate" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		if( DatabaseConfiguration.dbType() != DatabaseConfiguration.DBType.DB2 ) {
+			return;
+		}
+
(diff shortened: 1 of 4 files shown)


DavideD:
Is this for all databases?

DavideD:
Shouldn't you rebase this on top of this other PR? https://github.com/hibernate/hibernate-reactive/pull/970

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
```diff
@@ -0,0 +1,185 @@
...
+		String columnTypeQuery =
+				"select data_type from information_schema.columns " +
+						"where table_name = 'testentity' and column_name = 'numbervalue';";
+
+		final Configuration configuration = constructConfiguration( "update" );
+		configuration.addAnnotatedClass( TestEntity.class );
+
+		if ( DatabaseConfiguration.dbType() != DatabaseConfiguration.DBType.POSTGRESQL ) {
+			return;
+		}
```
DavideD, line 110: If you do things this way, this test will result in a success even if, in reality, you want to skip it.
That's way I've created the rule. It will also call the before/after methods.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
```diff
@@ -0,0 +1,185 @@
...
+			final Configuration configuration = super.constructConfiguration( hbm2DdlOption );
+			configuration.setProperty( Settings.HBM2DDL_JDBC_METADATA_EXTRACTOR_STRATEGY, GROUPED.toString() );
+			return configuration;
+		}
+	}
+
+	protected Configuration constructConfiguration(String hbm2DdlOption) {
+		Configuration configuration = constructConfiguration();
+		configuration.setProperty( Settings.HBM2DDL_AUTO, hbm2DdlOption );
+		configuration.setProperty( Settings.DEFAULT_CATALOG, "hreact" );
```
DavideD, line 52: Is setting the catalog part of the corner case? Because I don't think it's going to work for all databases

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
blafond: Right.   Some of these tests are not DB specific in their purpose, but DB specific in **query** implementation.  So options are still to find a cleaner (efficient) way to make that happen in one test class, or break them apart into multiple classes. I'm good with either.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
blafond: good point. Will tweak that test.

DavideD: I know that there was an issue where not setting the catalog wil lcause a `NullPointerException`. Iif you rebase on top of [this pr](https://github.com/hibernate/hibernate-reactive/pull/970), that issue should be solved.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
DavideD: Break it in separate classes, please. Maybe inner classes extending an abstract super class.

The rule I've created is not that powerful and it only works for classes.  if the only difference is the query, having an abstract super class with a mehod `getQuery` (and maybe `getResult`) will avoid a lot of duplicated work,

I think we could now switch to JUnit 5 and it will make the management of different test sper database easier. But that's a separate issue for when we have time.

DavideD:
By the way, the reason I sue the `#` and then the issue number is that it's clickable when looking at the history on github. You can see an example here: https://github.com/hibernate/hibernate-reactive/commits/main

DavideD:
If you only use the issue number (for example,` [936] SchemaMigrator/SchemaValidator tests for corner cases`) 936 is not linkable anymore

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
blafond: sure.. I'll  work toward abstraction and getQuery() & getResult().

@DavideD I added a `schema` package since there may be more tests.  Any issue with doing that at this time?

DavideD: > @DavideD I added a `schema` package since there may be more tests. Any issue with doing that at this time?

No, that's fine

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateProceedOnErrorTest.java
```diff
@@ -0,0 +1,96 @@
...
+		final Configuration configuration = constructConfiguration( "drop" );
+		configuration.addAnnotatedClass( From.class );
+
+		test( context, factoryManager.stop()
+				.thenCompose( v -> setupSessionFactory( configuration ) )
+				.thenCompose( v -> factoryManager.stop() ) );
+	}
+
+	@Test
+	public void testHaltOnError(TestContext context) {
```
DavideD, line 75: What's the purpose of this test?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateSchemaNameMySQLTest.java
```diff
@@ -0,0 +1,167 @@
...
+import org.hibernate.reactive.testing.DatabaseSelectionRule;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import io.vertx.ext.unit.TestContext;
+
+public abstract class SchemaUpdateSchemaNameMySQLTest extends BaseReactiveTest {
```
DavideD, line 29: We have this class already in the project: https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SchemaUpdateMySqlTestBase.java

Do we need a separate class for this additional tests or we can add them to the existing class?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
DavideD: Just keep in mind that we already have several test classes in the project and you should move them in the sama package if you want to create a new one

DavideD:
In the project we have several test classes for testing schema update/validation for each database (see `SchemaUpdate*TestBase`).
If I'm not mistaken, it seems you could add the test methods to those classes (But you will need to create a new one for Db2).

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateSchemaNameMySQLTest.java
blafond: It's possible to add the tests to current HR MySQL test but I may need to bring over the the ORM version of the entities which don't have PK/FK defined.  The tests will fails when loading the ASimpleNext metadata during build session factory.

I can do that but I think it muddies the scope of the **SchemaUpdateMySqlTestBase.java**

DavideD: Well OK, it would be nice to have all the common configuration in one place but we can do it later.

blafond:
@gbadner @DavideD  `NumericValidationTest.java`tests validation for a specific entity `BigDecimal` datatype. Since ORM test version does not specify DB type in the tests, so I'm assuming we want to set the test(s) up to do the same in HR.

The ORM test only tests if validation doesn't fail. My HR version actually executes a query to find the specific column and type and verify the DB's expected type.

Is this necessary?  In HR, I don't think we've been writing tests that don't query the data or schema metadata.

DavideD:
> @gbadner @DavideD NumericValidationTest.javatests validation for a specific entity BigDecimal datatype. Since ORM test version does not specify DB type in the tests, so I'm assuming we want to set the test(s) up to do the same in HR.

Ideally, every test should run on every database. The only time we don't do that is when a feature is  not supported by all databases or when we need to run native queries.

For this particular type of tests, I don't think it's possible because you will probably need to check that the column in the table exists and it's of the expected type. And I think you can only do that with native queries.

> The ORM test only tests if validation doesn't fail. My HR version actually executes a query to find the specific column and type and verify the DB's expected type.

Sounds good, but to test that the validation is working we would need the following use case:
1. Validation fails when a table is missing (with and without inheritance)
2. Validation fails when a column is missing
3. Validation doesn't fail if everything is alright

An example of how to test for case 1 is in `SchemaUpdateSqlServerTestBase#testVaslidationFails`.
We could move those tests in a separate class and run it for all databases

For case 2, I would run a native query that drops the column before running the test. You will need a native query for this case but because it's just a query,
you can keep track of them in a constant.

Case 3 is what you have done already

Let me know if this makes sense

DavideD:
I've just realized that we should also check that the validation doesn't fail (even if a column is missing) when `Settings.HBM2DDL_AUTO` is  not set or set to `none`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
gbadner: It is the dialect's implementation of `#getNameQualifierSupport` that determines what *may* be required.

If `NameQualifierSupport#CATALOG` is returned, then `Settings.DEFAULT_CATALOG` may need to be defined.

If `NameQualifierSupport#SCHEMA` is returned, then `Settings.DEFAULT_SCHEMA` may need to be defined.

If `NameQualifierSupport#BOTH` is returned, then both  `Settings.DEFAULT_CATALOG` and `Settings.DEFAULT_SCHEMA` may need to be defined.

I think it depends on the particular database whether it is needed and whether `GROUPED` strategy is required. I'm not sure what happens if, for example `#getNameQualifierSupport` returns `NameQualifierSupport#CATALOG` and `Settings.DEFAULT_SCHEMA` is provided. ORM may simply ignore the setting.

@dreab8, may be able to shed more light on this.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateProceedOnErrorTest.java
gbadner: By default, an error found while executing `SchemaMigrator` should not terminate the bootstrap process. This is controlled by `hibernate.hbm2ddl.halt_on_error` (default is `false`)

This test will try to create a table named `From`, which is a reserved word for most databases. It should cause warnings to get logged, but the `SessionFactory` should still be created. Obviously, there will be problems if an application tries to query or insert data related to the failed DDL.

The test just checks that no exception is thrown.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateSchemaNameMySQLTest.java
gbadner: I think it's worth having this in a separate test because it really is very much a corner case.

Notice that the entities map `schema = "test"`.

For MySQL, `#getNameQualifierSupport` returns `NameQualifierSupport.CATALOG`. 

In other words, the mapping really is incorrect; it should map `catalog="test"`.

Internally, MySQL `DatabaseMetaData` method take the `catalog` value and uses it like it is the schema. That's mentioned [here](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/MySqlReactiveInformationExtractorImpl.java#L21-L25)

I don't remember, but I think that this weird behavior was introduced at some point, and old mappings got broken.

This is a test that the legacy mapping will still work.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/NumericValidationTest.java
dreab8: If I remember correctly if the db does not support schema and we provide a schema in the mapping of a table we have errors not sure about the Settings, but I may be wrong...We need to test this behaviour.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateSchemaNameMySQLTest.java
DavideD: @blafond Can you add some javadoc for this class, please?

From what I understand, we are testing:
1. Combination of catalog + schema names
2. Two different classes with with the same entity name and table name

I also don't see anything specific to MySQL in this test class. I think it will work for MSSQL if we change the name of the catalog to `master` during the build and probably for PostgreSQL without additional changes. @blafond could you also check if it works for the other dbs?

DavideD:
Superseeded by https://github.com/hibernate/hibernate-reactive/pull/1017

PR #978, opened 2021-09-29
Close the connection after the session

### LINKED ISSUE #976: ReactiveSessionImpl#reactiveClose()
The implementation of `reactiveClose()` is first closing the connection, and then invoking close() on super.

I believe it needs to happen in the opposite order, for two reasons:

1. we need to flag the Session as closed ASAP to produce an error on illegal use
2. the super.close() method might need the connection to finalize pending operations

DavideD:
Fixes #976


Sanne (approved):

DavideD:
Merged

PR #981, opened 2021-09-29
Create Implementor interface

DavideD:
New interfaces that will make it easier for integrators
to access internal components. Similar to the SessionFactoryImplementor
interface in ORM.

@Sanne, we will have to update this line in Quarkus: https://github.com/DavideD/quarkus/blob/07c1037c07ace8b93dcb7a3f06ead6a35da965eb/extensions/hibernate-reactive/runtime/src/main/java/io/quarkus/hibernate/reactive/runtime/ReactiveSessionProducer.java#L30

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/spi/Implementor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/spi/Implementor.java
@@ -1,0 +1,19 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.common.spi;
+
+import org.hibernate.reactive.context.Context;
+import org.hibernate.service.ServiceRegistry;
+
+/**
+ * Allows access to object that can be useful for integrators
+ */
+public interface Implementor {
+
+	ServiceRegistry getServiceRegistry();
+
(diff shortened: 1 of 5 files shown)


DavideD:
At the moment we need [the Context](https://github.com/DavideD/quarkus/blob/756248a8cfe09b4ab5d49974156c45a990733230/extensions/hibernate-reactive/runtime/src/main/java/io/quarkus/hibernate/reactive/runtime/ReactiveSessionProducer.java#L31).

But I guess what we actually need is access to the `ServiceRegistry` (@Sanne ?)

gavinking:
I mean, then my preference would be to have a `getContext()` method, instead of this untypesafe `unwrap()`.

gavinking:
And `getServiceRegistry()`

Sanne:
Yes we can expose `getContext()` and `getServiceRegistry()`, but I'd still prefer them to not be visible on the API.

Sanne:
Actually allow me to be clearer: we can't have them on the public API if we want the API to be supportable.

gavinking:
Sure, i think you misunderstood. I'm objecting to `unwrap()`, not to a new `Implementor` interface.

DavideD:
OK, I've  created a single Implemetor interface that has a `getServiceRegistry` and a `getContext`.

gavinking:
> OK, I've created a single Implemetor interface that has a `getServiceRegistry` and a `getContext`.

Looks much better now.

DavideD:
Merged, thanks!

PR #983, opened 2021-09-30
 Have Uni return types on all Session opening methods #973 

### LINKED ISSUE #950: Consider forcing Uni return types on all Session opening methods

DavideD:
Fixes #950
Superseeds #973 

Instead of the previous approach of completely remove the existing methods for opening a session lazily, I've deprecated them.
The main reason is that we are having some issues with quarkus and by deprecating the method in the API we can still have a release of Hibernate Reactive and more flexibility about how to fix the integration with Quarkus.

The drawback is that we have to find a new name for the new method (and we cannot use the current name `openSession`).
I've opted for `newSession` instead but I'm open to suggestions.

As an alternative, I was thinking of `prepareSession` or `prepareNewSession` as an alternative. I think it might gve better the idea that the returned value won't be an instance of `Session`.

@gavinking I think this is something you might want to have a look.


DavideD:
@Sanne, you should be able to test Quarkus with this branch without any additional changes but I'm preparing a branch that uses the new API

Sanne:
> you should be able to test Quarkus with this branch without any additional changes ...

Tested it, that worked fine.

> but I'm preparing a branch Quarkus that uses the new API

Nice, let me know when you have something.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1515,6 +1515,63 @@
...
+		 * <p>
+		 * The client must explicitly close the session by calling
+		 * {@link Session#close()}.
+		 *
+		 * @see #withSession(Function)
+		 */
+		Uni<Session> newSession();
+
+		/**
+		 * Obtain a new {@link Session reactive session} {@link CompletionStage} for a
```
Sanne, line 1533:
```suggestion
		 * Obtain a new {@link Session reactive session} {@link Uni} for a
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1663,10 +1752,10 @@
 		 * <p>
 		 * <il>
 		 * <li>If there is already a stateless session associated with the
-		 * current reactive stream, then the work will be executed using that
+		 * current reactive stream and given tenant id, then the work will be executed using that
 		 * session.
 		 * <li>Otherwise, if there is no stateless session associated with the
-		 * current stream, a new stateless session will be created.
+		 * current stream  and given tenant id, a new stateless session will be created.
```
Sanne, line 1758:
```suggestion
		 * current stream and given tenant id, a new stateless session will be created.
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1681,6 +1770,14 @@
 		 * Perform work using a {@link Session reactive session} for a
 		 * specified tenant within an associated {@link Transaction transaction}.
 		 * <p>
+		 * <il>
+		 * <li>If there is already a session associated with the
+		 * current reactive stream and given tenant id, then the work will be executed using that
+		 * session.
+		 * <li>Otherwise, if there is no session associated with the
+		 * current stream  and given tenant id, a new stateless session will be created.
```
Sanne, line 1778:
```suggestion
		 * current stream and given tenant id, a new stateless session will be created.
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1697,6 +1794,14 @@
 		 * Perform work using a {@link StatelessSession reactive session} for a
 		 * specified tenant within an associated {@link Transaction transaction}.
 		 * <p>
+		 * <il>
+		 * <li>If there is already a stateless session associated with the
+		 * current reactive stream and given tenant id, then the work will be executed using that
+		 * session.
+		 * <li>Otherwise, if there is no stateless session associated with the
+		 * current stream  and given tenant id, a new stateless session will be created.
```
Sanne, line 1802:
```suggestion
		 * current stream and given tenant id, a new stateless session will be created.
```

Sanne (changes requested):

gavinking:
Look, I'm not keen on the idea that we're changing the name of a public API because of some transient integration issue.

If we really think `newSession()` is a better name, then OK, fine. To me it doesn't seem better.

Sanne:
@gavinking we're open to suggestions of course.

I did suggest `newSession()` because the other method one could use - `withSession` - does leverage the vertx context, which confused me initially as it might reuse some existing session in the context (which honestly surprised me).

So I think `newSession` nicely expresses the aspect that this one method actually guarantees it to be a new one. And this conveys a suggestion about when this is a good choice, as otherwise I'd normally always prefer to use `withSession` which IMO looks better when working with reactive.

Would you like to suggest some alternatives?

gavinking:
Well I was very happy with `openSession()` which to me also pretty unambiguously signals a new session, and is the naming that's used in Hibernate ORM.

DavideD:
`createSession` also works for me.

DavideD:
Or `openSessionUni` and `openSessionStage`

Sanne:
> Well I was very happy with `openSession()` which to me also pretty unambiguously signals a new session, and is the naming that's used in Hibernate ORM.

while it's not legal in source code we can have both flavours of `openSession()` in the API; it needs a little tooling gymnastics to make it happen though, so I'd only do that if you feel it's important.

gavinking:
What I'm saying is I don't find your justification for not just removing the previous versions to be very convincing.

gavinking:
> What I'm saying is I don't find your justification for not just removing the previous versions to be very convincing.

(It puts the cart before the horse.)

Sanne:
> What I'm saying is I don't find your justification for not just removing the previous versions to be very convincing.

oh yes sure, we initially tried to remove it.

On the Quarkus side we're having two Panache tests failing for reasons we couldn't understand yet; we spent some days on this and that's why this is coming a week later than when we last discussed it - I then suggested to Davide that if it was possible to release another CR which had both methods, we could then upgrade Quarkus and migrate one module at a time - and bisect the problem, which is eluding us currently.

So another alternative is we try harded to figure it out on the Quarkus side :) if we solve that we could just delete them, but then also deprecating these would feel more like doing the right thing for the few users we already have.

gavinking:
Why don't you just add some method that Panache needs to the new `Implementor` interface and do a typecast in the Quarkus code?

Sanne:
> Why don't you just add some method that Panache needs to the new `Implementor` interface and do a typecast in the Quarkus code?

brilliant :)

Thanks

@DavideD let's do that? It's not as nice to other users but will definitely help us move on.

gavinking:
I mean, to me tat seems *much* better than releasing 1.0 with some operations on a public API that you're already planning on removing later.

Sanne:
> I mean, to me tat seems _much_ better than releasing 1.0 with some operations on a public API that you're already planning on removing later.

well to be clear I was going to make this another CR, and then remove it before tagging Final.

But sure let's use the `Implementor` instead.

PR #987, opened 2021-10-07
Upgrade to Mutiny 1.1.1

jponge:


DavideD:
Thanks a lot.

Please, make sure that there is an issue before merging this next time. I'd like user to know that there is a change in the library in case somebody wants to see what's new in a via the issue tracker (even if it's a micro).

PR #1002, opened 2021-10-15
fix hibernate#1001 

### LINKED ISSUE #1001: DefaultVertxInstance should not create a new Vertx instance if one exists already
[See this discussion.](https://github.com/hibernate/hibernate-reactive/issues/995#issuecomment-944404509)

codingxu97:
fixes #1001.
```DefaultVertxInstance``` creates ```Vertx``` instance only if current vertx is null when ```start()``` invokes, and stop vertx only if it created by ```start()```.


DavideD:
Thanks!

I've created a new PR that contains your commits squashed together and add a new one to check that the tests still works when we don't override the VertxInstance service anymore.

PR #1003, opened 2021-10-16
Create a new Vert.x instance only if one doesn't exist already

### LINKED ISSUE #1001: DefaultVertxInstance should not create a new Vertx instance if one exists already

DavideD:
Fixes #1001 
Superseeds https://github.com/hibernate/hibernate-reactive/pull/1002

Thanks to @codingxu97 fix, we don't need to override the VertxInstance for the tests

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BaseReactiveTest.java
@@ -26,7 +26,6 @@ import org.hibernate.reactive.provider.service.ReactiveGenerationTarget;
 import org.hibernate.reactive.provider.service.ReactiveSchemaManagementTool;
 import org.hibernate.reactive.stage.Stage;
 import org.hibernate.reactive.testing.SessionFactoryManager;
-import org.hibernate.reactive.vertx.VertxInstance;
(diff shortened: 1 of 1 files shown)


codingxu97:
> I don't know if I should merge this before the final but it shouldn't cause any problem: for somebody that already overrides the service, nothing will change. In the other case a new Vert.x instance gets created anyway.

It will not affect the previous implementation IMHO. As you told, this PR just to avoid duplicate creation of Vertx instance, this is the design principles of Vert.x, it will be broken without existence check.

codingxu97:
> I don't know if I should merge this before the final but it shouldn't cause any problem: for somebody that already overrides the service, nothing will change. In the other case a new Vert.x instance gets created anyway.

It will not affect the previous implementation IMHO. As you told, this PR just to avoid duplicate creation of Vertx instance, this is the design principles of Vert.x, it will be broken without existence check.

DavideD:
Ok, Quarkus overrides the default VertxInstance service anyway and I've run the tests for the hibernate reactive extension and everything seems to work fine.

I will merge this tomorrow

gavinking:
Looks fine to me.

Issue #1005, opened 2021-10-19
Schema Validation tests for basic types


blafond:
Schema update and validation tests use entities that only test a few datatypes.  The differences in DB runtime types warrant adding tests to verify our reactive result set processing and schema validation and update.  This will also exercise and check the vertx sql clients for complete datatype processing.

see [#1004](https://github.com/hibernate/hibernate-reactive/issues/1004)

PR #1017, opened 2021-10-21
Schema Update & Validation corner cases

blafond:
Addresses #1005

* Added simple verification of datatype validation
  * Fixed datatype validation error in PG
  * remaining issues with MySQL datatype validation
  * SQL Server issue with **Double** and **double*** datatypes
* Added schema update test for add/remove columns to/from entity
  * issue with update with removed column
  
Also moved Schema Update tests to new org.hibernate.reactive.schema package


blafond:
Pushed latest.  The last commit includes a work-around due to no "double" support in MSSQLServer. Thx @DavideD

DavideD:
Something fails on CI with this PR

DavideD:
Also @blafond, I understand that this is a draft but now we have figure out a way to make it work. Could you clean up the PR and remove all the unrelated changes, please?

blafond:
The CI build fails on MariaDB.  Locally I ran it via our Podman instance and it runs fine.  The **MariaDatabase** extends **MySQLDatabase** so I assumed the datatypes would be identical.  The CI build log says otherwise.  If I'm not "really" running MariaDB locally.... I need figure out what's different and how to run against actual instance.

Yes.. I'll clean this up.

DavideD:
How do you test it locally?
These failures are the same I mentioned to you before. 

MariaDB and MySQL are similar so we can reuse some code.
Doesn't mean that we might not have different behaviour in some areas.

If it helps, CI is using this command to build:
```
./gradlew build -PshowStandardOutput -Pdocker -Pdb=MariaDB
```
But you should be able to see the same error using `./gradlew test` instead of `./gradlew build`

DavideD:
> If I'm not "really" running MariaDB locally.

?? What do you mean you are not really running MariaDB locally?

blafond:
just a config issue.  Got it running and modifying **MariaDatabase** to add the 2 queries and expected types list.  Runs locally now. CI should run OK once I push.

blafond:
@DavideD  So I'm starting to understand that the **ColumnBasicTypesTest** is currently verifying that we can use native queries to access column information and verify the expected type is returned for different databases. This probably isn't true schema validation which occurs in hibernate.  

The **ColumnBasicTypesTest** for MySQL is failing for all types because the MySQL Client is returning the "DATA_TYPE" column type as **LONGBLOB**, where for instance, PG's type is **VARCHAR**.  Resulting error in executing each native query is:
`java.util.concurrent.CompletionException: java.lang.ClassCastException: java.lang.String cannot be cast to io.vertx.core.buffer.Buffer`
... in **ResultSetAdapter.getBlob()**

I don't know if this would be a vertx MySQL client error or if not, how to convert the **String** value to a Blob on the fly.

DavideD:
Please, make sure to rebase to the latest changes.
We merged some critical stuff on `main`.

> @DavideD So I'm starting to understand that the **ColumnBasicTypesTest** is currently verifying that we can use native queries to access column information and verify the expected type is returned for different databases. This probably isn't true schema validation which occurs in hibernate.

The purpose of that test is to check that the type used for the columns on the db  is of the expected type.
We are not really testing the schema validation because too many things happen in between.
For example, we have seen before that even if this test was working something else was causing the validation of the schema to fail.

I will have a look

blafond:
thx... and yes, I'll be rebasing after watching the changes/version upgrades on main.

DavideD:
> I don't know if this would be a vertx MySQL client error or if not, how to convert the String value to a Blob on the fly.

It's not an issue with the Vert.x client, it is returning the value as a `String`. The way i is supposed to.
For some reason, Hibernate Reactive is trying to convert the value into a `Blob`. Not sure why considering that we specify that it has to be a `String`  when we create the query. I'm still looking into it.

DavideD:
Nevermind that. It could be a bug in the Vert.x client. I will write a test case for it to check.

DavideD:
I think there are two bugs in the Vert.x client:
* The main issue is that returning the wrong type for the column on the database ( I still need to write a test for this)
* It doesn't convert a String into a Buffer when using `Tuple#getBuffer` ([I've sent a PR](https://github.com/eclipse-vertx/vertx-sql-client/pull/1062)) but I'm not sure this is really a bug

For now, if you apply this fix it should solve the issue with MySQL: https://github.com/eclipse-vertx/vertx-sql-client/pull/1062

blafond:
K. Thx

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -721,7 +721,14 @@ public Ref getRef(String columnLabel) {
 
 	@Override
 	public Blob getBlob(String columnLabel) {
-		Buffer buffer = (Buffer) row.getValue( columnLabel );
+		final Object value = row.getValue( columnLabel );
+		// Vertx MySQL client is currently not converting String result set values to Buffer
+		// in Tuple.getBuffer(int pos)
+		// Issue logged and PR submitted.
```
DavideD, line 727: OK, but add the issue number, please

blafond: got it

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/ColumnTypesMappingTest.java
```diff
@@ -0,0 +1,226 @@
...
+import java.sql.Blob;
+import java.sql.SQLException;
+
+/**
+ * Check that each property is mapped as the expected type in the database.
+ */
+public class ColumnTypesMappingTest extends BaseReactiveTest {
+
+	@Rule
+	public DatabaseSelectionRule dbRule = DatabaseSelectionRule.skipTestsFor( DB2 );
```
DavideD, line 31: We can change it later but there is no reason for this test not to work with Db2, isn't it?

DavideD: If it doesn't require any change

blafond: The tests pass for DB2 if I remove the rule

DavideD:
Vert.x SQL client is working correctly and I didn't realized I was testing with two different version of MySQL

MySQL 8 define the column  `data_type` for the view `information_schema.columns` as a `longtext` that is returned as a `Blob`.

MySQL 5 define the column  `data_type` for the view `information_schema.columns` as a `varchar` that is returned as a `String`.

I don't think there is much more to do here. it's just a bit unfortunate that we cannot rely on the fact that the column type is a String for all the databases.

hibernate-reactive-core/build.gradle
```diff
@@ -131,7 +131,7 @@ tasks.addRule( "Pattern testDb<id>" ) { String taskName ->
 }
 
 // The dbs we want to test when running testAll
-def dbs = ["MariaDB", 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB']
+def dbs = ["MariaDB", 'MySQL', 'PostgreSQL', 'CockroachDB', 'SqlServer']
```
DavideD, line 134: I don't know what happened here but Db2 has been replaced by SqlServer

blafond:
Closing and resubmitting with cleaner commits.

PR #1018, opened 2021-10-21
Avoid invoking ConfiguredJdbcBatchSize each time a session is created

### LINKED ISSUE #1016: Avoid invoking ConfiguredJdbcBatchSize each time a session is created
> apparently HR is invoking `ConfiguredJdbcBatchSize` on `SharedSessionContractImplementor` each time a Session is constructed. We need to either void that, or make that method more efficient - might be an ORM patch in this case.

This has been patched in ORM so we just need to make sure we called getConfiguredJdbcBatchSize in both sessions (Stateless and normal).

DavideD:
FIxes #1016


gavinking:
Well because it looks like we create the `BatchingConnection` once when the session is created, and save the setting from the factory. And so if you then call `setBatchSize()` the change will just be ignored because the `BatchingConnection` uses its cached value.

DavideD:
Ah, ok. I didn't think about the `setBatchSize` use case.

PR #1033, opened 2021-10-29
Vert.x 4.2

### LINKED ISSUE #1010: Upgrade Vert.x to 4.2
The final has been released: https://search.maven.org/artifact/io.vertx/vertx-sql-client

It seems that this upgrade will require some changes and we could start to have a look.

DavideD:
Fixes #1010 
This change means that the next Hibernate Reactive release is going to be 1.1

I'm going to create a maintainance branch for Hibernate Reactive 1.0

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TenantDependentPool.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TenantDependentPool.java
@@ -20,6 +20,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.sqlclient.Pool;
 import io.vertx.sqlclient.PoolOptions;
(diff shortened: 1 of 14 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/TenantDependentPool.java
```diff
@@ -39,113 +40,118 @@
  */
 public class TenantDependentPool extends DefaultSqlClientPool {
 
-    private MultiplePools pools;
```
Sanne, line 42: what happened here? Formatting changes?

Sanne: ah, yes I see.

DavideD:
Thanks, I will create a separate branch for 1.0 when we have changes to backport

PR #1034, opened 2021-10-29
Re-enable MultipleContextTest

DavideD:
I want to see if this test works now


DavideD:
I've added this commit: https://github.com/hibernate/hibernate-reactive/pull/1034/commits/c8349a0b5bb078633efa2c8e63c7b6fd604a1461

It's a proof of concept about what we need to change.

PR #1035, opened 2021-11-01
Dbs upgrades

DavideD:


DavideD:
I've created a separate issue for Db2: https://github.com/hibernate/hibernate-reactive/issues/1037

PR #1041, opened 2021-11-03
overload withTransaction() for convenience

gavinking:
as suggested by @Sanne

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -1621,13 +1621,38 @@ public interface Mutiny {
 		 * The session will be {@link Session#flush() flushed} and closed
 		 * automatically, and the transaction committed automatically.
 		 *
-		 * @param work a function which accepts the session and returns
-		 *             the result of the work as a {@link Uni}.
+		 * @param work a function which accepts the session and transaction
+		 *             and returns the result of the work as a {@link Uni}.
 		 *
 		 * @see #withSession(Function)
 		 * @see Session#withTransaction(Function)
 		 */
 		<T> Uni<T> withTransaction(BiFunction<Session, Transaction, Uni<T>> work);
+
+		/**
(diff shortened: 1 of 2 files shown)


blafond (approved):
Pulled PR.  Ran a couple of existing tests that used `withTransaction( ( session, tx) -> ...` and removed the **tx**.  Both tests succeeded.

Looks good to me

DavideD:
Merged it.
But I've just realized that it doesn't do the same for the StatelessSession, I'm going to add it

DavideD:
And for the tenantId

gavinking:
I mean, my inclination would be to leave this as a convenience for the most common case, rather than filling up `SessionFactory` with a cartesian product if overloads.

DavideD:
I prefer not to have these kind of inconsistencies in these API.

If we believe i's something worth having for one method, we should have it for the other cases as well.
If it's too much hassle, maybe we should reconsider adding it (I mean, how hard is it for a user to create a utility method when they need something similar).

gavinking:
> If we believe it's something worth having for one method, we should have it for the other cases as well.

Well, I'm not quite sure that's true. The value of a "convenience" depends how often we think it's going to be used. I think it's pretty clear that the other two versions of `withTransaction()` are used *much* less often. 

And on the other hand, the convenience comes at a cost: every time I autocomplete `withTransaction()`, the IDE presents me with a list of all the overloads, and if there's many of them, I'm going to have trouble finding the one I'm looking for.

The issue here is that it's a Cartesian product of (stateful, stateless) ⨉ (no tenant id, tenant id) ⨉ (transaction, no transaction) ⨉ (who knows what else we'll add in the future). For now it might seem OK to have 8 overloads, though that already seems rather a lot to me, but consider that if in future we add one additional option, that blows out to 16, which is just too many!

(Well, that's one reason why people invent builder APIs, of course, and I know I've resisted that for other reasons, but actually here in this specific case, a builder wouldn't help much.)

DavideD:
Let's add it for the stateless session but not for the tenant id. I think it's a good trade off

gavinking:
Yes, that's reasonable.

PR #1057, opened 2021-11-05
Add StatelessSession#setBatchSize

### LINKED ISSUE #1056: Add setBatchSize to the stateless session

DavideD:
Fixes #1056

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -1027,12 +1027,13 @@ public interface Mutiny {
 		CacheMode getCacheMode();
 
 		/**
-		 * Set the session-level JDBC batch size, overriding the batch size set
+		 * Set the session-level batch size, overriding the batch size set
 		 * by the configuration property {@code hibernate.jdbc.batch_size}.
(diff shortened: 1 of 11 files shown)


gavinking:
You realize `StatelessSession` doesn't have any setter methods ... because it's supposed to be "stateless"....

I wonder if there's some other way we could handle this: like, implicitly turning on batching when you use `persistAll()`/`updateAll()`....

DavideD:
I didn't think the batchSize was considered part of the state. In Hibernate ORM Stateless has a setJdbcBatchSize(https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/SharedSessionContract.java#L194).

gavinking:
Sure, it's very arguable :-)

But even so, I guess I feel like it would fall more in line with the whole philosophy and feel of `StatelessSession` to make it an optional argument of the `xxxxxAll()` methods.

As a user I would find that more convenient, frankly.

gavinking:
The reason it's an attribute for regular sessions is that the user doesn't have fine control over updates. Here they do.

DavideD:
Alright, I will try to add it as an optional parameter

DavideD:
@gavinking I've applied the changes we've discussed

gavinking:
OK, great, LGTM

PR #1063, opened 2021-11-10
rework new doc section in a more narrative style

gavinking:


documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -1006,89 +1034,82 @@ currentContext.runOnContext( event -> {
...
+//[source, JAVA, indent=0]
+//----
+//Mutiny.SessionFactory sessionFactory = ...
+//VertxInstance vertxInstance = ((Implementor) sessionFactory)
+//					.getServiceRegistry()
+// 					.getService( VertxInstance.class );
+//Vertx vertx = vertxInstance.getVertx();
+//----
+//
+//The same will work for `Stage.SessionFactory`.
```
DavideD, line 1112: Are you planning to add this back somewhere? Maybe in a separate section about the service registry?

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -977,26 +977,54 @@ sessionFactory.withTransaction( (session, tx) -> session.persist(book) )
...
 
 [source, JAVA, indent=0]
 ----
 HR000068: This method should exclusively be invoked from a Vert.x EventLoop thread; ...
 ----
 
-One way to run the code in the correct Vert.x context is the following:
+On the other hand, if you use `openSession()`, you'll have to manage the
+association between sessions and contexts yourself. Now, that's in principle
+straightforward, but you'd be surprised how often people mess up.
```
DavideD, line 1015: I assure you, nobody is going to be surprised :-)

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -977,26 +977,54 @@ sessionFactory.withTransaction( (session, tx) -> session.persist(book) )
...
+
+IMPORTANT: The session is not thread-safe (or "stream-safe"), so using it across
+different threads (or reactive streams) may cause bugs that are _extremely_ hard
+to detect. Don't say we didn't warn you!
```
DavideD, line 1019: I think we should add a quick example of what it means to use the session on different reactive streams.
For example a loop that persist multiple entities:
```
List<CompletionStage> list = ...
for (Entity entity : entities) {
    list.add(session.persist(entity));
}
CompletableFuture.allOf( list ).thenCompose( session::flush );
```

Otherwise it's hard to wrap the head around when the warning applies

DavideD (approved):
Looks great

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Well I had two issues with this section:

1. It seemed to be describing some very technical interfaces that aren't of use to the casual user, and didn't seem to belong in an introductory guide.
2. It didn't explain why I might need it.

But, if you can give a good motivation for it being there, then I'll to add that motivation to the text and uncomment this passage.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: hahaha

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Yup, agreed. Though I'm not sure I completely understand the example you've given myself :-)

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: If I was being completely honest, I would have written:

> Now, that seems pretty straightforward *to me*, and I'm continually surprised how often *you lot* mess it up.

gavinking: With an eyeroll.gif

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Shouldn't it be:
```
CompletionStage<Void> stage = ...
for (Entity entity : entities) {
    stage = stage.thenCompose( v -> session.persist(entity));
}
stage.thenCompose( session::flush );
```
?

gavinking: I guess I never looked at `CompletableFuture.allOf()` :-)

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I guess is not useful for most users but accessing the vertx instance when it's created by Hibernate Reactive might be helpful.

I don't have a use case though and it's an advance feature, so I don't mind to omit it for now.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I've used `allOf` because it seems like something useful but it  isn't this wrong already?
```
    list.add(session.persist(entity));
```
The `persist` are not chained and they will start as soon as the CompletionStage is created.

gavinking: Right, OK, then it's a good example.

But perhaps we should have one for Mutiny?

gavinking: > but isn't this already wrong?

Yeah well that's actually what confused me about the example: there's kinda two errors in it, so I wasn't clear that it was demonstrating the particular error we were discussing in the text.

DavideD: OK

> But perhaps we should have one for Mutiny?

Sure, it was just the first thing I thought

gavinking: Just put a bit of thought into this, and frankly I'm not sure I understand how these APIs do concurrency well enough to be able to write a good example (i.e. one that isn't wrong in *multiple* ways). So I'm going to leave the example out for now, until I think I know WTF I'm talking about.

DavideD: Can't we just use the example I mentioned, explain the two issues with it and then replace it when we find something better?

Issue #1032, opened 2021-10-29
Add support for Oracle


DavideD:
Requires #1010 

https://vertx.io/docs/vertx-oracle-client/java/

DavideD:
I've created a branch upstream `wip/oracle` to keep track of the progress:
https://github.com/hibernate/hibernate-reactive/tree/wip/oracle

It's not clean but it allows testing with Oracle like we do for the other databases:
```
./gradlew testDbOracle -Pdocker
```
A lot more tests than I expected are already working. I haven't looked into the failures yet but most seems related to the usual: queries, query parameters syntax and converting the result type from queries.

Cheers

DavideD:
I've updated the branch and when using the latest snapshot we have 7 failures caused by the same 2 errors:
* `Error Msg = ORA-00932: inconsistent datatypes: expected - got BLOB `: I've already [created an issue for Vert.x](https://github.com/eclipse-vertx/vertx-sql-client/issues/1085).
* Inheritance and some other case where we have queries with `RETURNING ROWID INTO ?`. I will check this next week

tsegismont:
@DavideD 7 failures out of the entire test suite? Sounds like good news!

DavideD:
I think the other problem that it's missing is an issue with Vert.x: https://github.com/eclipse-vertx/vertx-sql-client/issues/1108

hernael:
Hi @DavideD, a question: Having added this issue to the next milestone, means that it is possible have Oracle Support in the next hibernate-reactive version?

gavinking:
Haaaay, that's a _really_ great question! I've lost track of where we're at on Oracle support.

blafond:
The `vertx-oracle-client` is in the build, which we developed against... probably 4.2.1?

Still waiting on [#1108](https://github.com/eclipse-vertx/vertx-sql-client/issues/1108) which right now is targeted at 4.3.0 vertx-sql-client

DavideD:
> Having added this issue to the next milestone, means that it is possible have Oracle Support in the next hibernate-reactive version

That's the plan but I need to rebase and review what we have so that I can document the limitations (see https://github.com/hibernate/hibernate-reactive/issues/1032#issuecomment-991155858).

PR #1064, opened 2021-11-11
Work on failed tests for Oracle support

blafond:
targeted at #1032 

Attempts at addressing issues in ResultSetAdapter and entity column definitions that use oracle key-word as column names.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -16,6 +16,7 @@ import org.hibernate.engine.jdbc.BlobProxy;
 import java.io.InputStream;
 import java.io.Reader;
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.net.URL;
 import java.sql.Array;
 import java.sql.Blob;
@@ -186,6 +187,16 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public boolean getBoolean(String columnLabel) {
+		final Object value = row.getValue( columnLabel );
+		wasNull = value == null;
+		if ( wasNull ) {
+			return false;
+		}
+
+		if( value instanceof BigInteger ) {
+			return ((BigInteger)value).intValue() > 0;
+		}
+
 		Boolean bool = row.getBoolean( columnLabel );
 		wasNull = bool == null;
 		return !wasNull && bool;
@@ -375,6 +386,7 @@ public class ResultSetAdaptor implements ResultSet {
 
 			@Override
 			public int getColumnType(int column) {
+//				if( rows.columnDescriptors().isEmpty() ) return 0;
 				ColumnDescriptor descriptor = rows.columnDescriptors().get(column-1);
 				return descriptor.isArray() ? Types.ARRAY : descriptor.jdbcType().getVendorTypeNumber();
 			}
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -197,8 +197,7 @@ public class ResultSetAdaptor implements ResultSet {
 			// get the int value and convert to expected boolean type for reactive
 			try {
 				int value = getInt( columnLabel );
-				boolean returnValue = BigDecimal.ZERO.intValue() == value ? false : true;
-				return returnValue;
+				return value!=0;
 			}
 			catch (Exception e) {
 				// ignore second exception and throw first cce
@@ -744,9 +743,15 @@ public class ResultSetAdaptor implements ResultSet {
 			return null;
 		}
 
-		byte[] bytes = value instanceof String ? ( (String) value ).getBytes()
-				: value instanceof byte[] ? (byte[]) value : ( (Buffer) value ).getBytes();
-		return BlobProxy.generateProxy( bytes );
+		if( value instanceof Buffer ) {
+			return BlobProxy.generateProxy( ( (Buffer) value ).getBytes() );
+		} else if( value instanceof String ) {
+			return BlobProxy.generateProxy( ( (String) value ).getBytes() );
+		} else if ( value instanceof byte[] ) {
+			return BlobProxy.generateProxy( (byte[]) value );
+		}
+
+		throw new IllegalArgumentException( "Unexpected type: " + value.getClass() );
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -16,7 +16,6 @@ import org.hibernate.engine.jdbc.BlobProxy;
 import java.io.InputStream;
 import java.io.Reader;
 import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.net.URL;
 import java.sql.Array;
 import java.sql.Blob;
@@ -187,19 +186,25 @@ public class ResultSetAdaptor implements ResultSet {
 
 	@Override
 	public boolean getBoolean(String columnLabel) {
-		final Object value = row.getValue( columnLabel );
-		wasNull = value == null;
-		if ( wasNull ) {
-			return false;
+		try {
+			Boolean bool = row.getBoolean( columnLabel );
+			wasNull = bool == null;
+			return !wasNull && bool;
 		}
-
-		if( value instanceof BigInteger ) {
-			return ((BigInteger)value).intValue() > 0;
+		catch (ClassCastException cce) {
+			// Since Oracle doesn't support an actual boolean/Boolean datatype
+			// Oracle8iDialect in ORM registers the BOOLEAN type as a 'number( 1, 0 )'
+			// get the int value and convert to expected boolean type for reactive
+			try {
+				int value = getInt( columnLabel );
+				boolean returnValue = BigDecimal.ZERO.intValue() == value ? false : true;
+				return returnValue;
+			}
+			catch (Exception e) {
+				// ignore second exception and throw first cce
+				throw cce;
+			}
 		}
-
-		Boolean bool = row.getBoolean( columnLabel );
-		wasNull = bool == null;
-		return !wasNull && bool;
 	}
 
 	@Override
@@ -739,9 +744,8 @@ public class ResultSetAdaptor implements ResultSet {
 			return null;
 		}
 
-		byte[] bytes = value instanceof String
-				? ( (String) value ).getBytes()
-				: ( (Buffer) value ).getBytes();
+		byte[] bytes = value instanceof String ? ( (String) value ).getBytes()
+				: value instanceof byte[] ? (byte[]) value : ( (Buffer) value ).getBytes();
 		return BlobProxy.generateProxy( bytes );
 	}
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
@@ -175,10 +175,19 @@ public class SqlClientConnection implements ReactiveConnection {
 			int i = 0;
 			RowSet<Row> resultNext = result;
 			if ( parametersBatch.size() > 0 ) {
-				do {
-					updateCounts[ i++ ] = resultNext.rowCount();
-					resultNext = resultNext.next();
-				} while ( resultNext != null && i < parametersBatch.size() );
+				final RowIterator<Row> iterator = resultNext.iterator();
+				if ( iterator.hasNext() ) {
+					while ( iterator.hasNext() ) {
+						updateCounts[i++] = iterator.next().getInteger( 0 );
+					}
+					resultNext = null;
+				}
+				else {
+					do {
+						updateCounts[i++] = result.rowCount();
+						resultNext = resultNext.next();
+					} while ( resultNext != null && i < parametersBatch.size() );
+				}
 			}
 
 			if ( resultNext != null || i != parametersBatch.size() ) {
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UUIDAsBinaryType.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UUIDAsBinaryType.java
@@ -23,6 +23,7 @@ import io.vertx.ext.unit.TestContext;
 import static java.util.Arrays.asList;
 import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.MARIA;
 import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.MYSQL;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.ORACLE;
 import static org.hibernate.reactive.testing.DatabaseSelectionRule.runOnlyFor;
 import static org.hibernate.reactive.testing.DatabaseSelectionRule.skipTestsFor;
 
@@ -105,7 +106,7 @@ public abstract class UUIDAsBinaryType extends BaseReactiveTest {
 	public static class ForOtherDbsTest extends UUIDAsBinaryType {
 
 		@Rule
-		public DatabaseSelectionRule rule = skipTestsFor( MYSQL, MARIA );
+		public DatabaseSelectionRule rule = skipTestsFor( MYSQL, MARIA, ORACLE );
 
 		@Override
 		protected Configuration constructConfiguration() {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        db: [ 'MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer' ]
+        db: [ 'MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer', 'Oracle' ]
     steps:
     - uses: actions/checkout@v2
     - name: Get year/month for cache key
--- a/.github/workflows/tracking-orm-5.build.yml
+++ b/.github/workflows/tracking-orm-5.build.yml
@@ -80,7 +80,7 @@ jobs:
     strategy:
       matrix:
         orm-version: [ '[5.6,5.7)' ]
-        db: ['MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer']
+        db: [ 'MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer', 'Oracle' ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK 1.8
--- a/.github/workflows/tracking-vertx-4.build.yml
+++ b/.github/workflows/tracking-vertx-4.build.yml
@@ -78,7 +78,7 @@ jobs:
     strategy:
       matrix:
         vertx-version: [ '[4.2,4.3)' ]
-        db: [ 'MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer' ]
+        db: [ 'MariaDB', 'MySQL', 'PostgreSQL', 'DB2', 'CockroachDB', 'MSSQLServer', 'Oracle' ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK 1.8
--- a/hibernate-reactive-core/build.gradle
+++ b/hibernate-reactive-core/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     testImplementation "io.vertx:vertx-mysql-client:${vertxVersion}"
     testImplementation "io.vertx:vertx-db2-client:${vertxVersion}"
     testImplementation "io.vertx:vertx-mssql-client:${vertxVersion}"
+	testImplementation "io.vertx:vertx-oracle-client:${vertxVersion}"
 
     // Optional dependency of vertx-pg-client, essential when connecting via SASL SCRAM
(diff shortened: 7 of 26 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -186,6 +187,16 @@ public String getString(String columnLabel) {
...
+		if ( wasNull ) {
+			return false;
+		}
+
+		if( value instanceof BigInteger ) {
+			return ((BigInteger)value).intValue() > 0;
+		}
+
 		Boolean bool = row.getBoolean( columnLabel );
 		wasNull = bool == null;
```
DavideD, line 201: You don't need to reassign the wasnull. If it wa null before for `getValue`, it is still null for `geBoolean`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -375,6 +386,7 @@ public int getColumnCount() {
 
 			@Override
 			public int getColumnType(int column) {
+//				if( rows.columnDescriptors().isEmpty() ) return 0;
```
DavideD, line 389: What's this comment?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -186,6 +187,16 @@ public String getString(String columnLabel) {
...
+		final Object value = row.getValue( columnLabel );
+		wasNull = value == null;
+		if ( wasNull ) {
+			return false;
+		}
+
+		if( value instanceof BigInteger ) {
+			return ((BigInteger)value).intValue() > 0;
+		}
+
```
gavinking, line 199: uuhhhhh what's going on here?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -186,6 +187,16 @@ public String getString(String columnLabel) {
...
 		Boolean bool = row.getBoolean( columnLabel );
```
DavideD, line 200: No need to reread the row, this should work: 
```
Boolean bool = (Boolean) value;
```

DavideD: Personally, I would just do:
```
final Object value = row.getValue( columnLabel );
Boolean bool = value != null && value instanceof BigInteger
                               ? BigInteger.ONE.equals( (BigInteger) value) // TRUE only when 1
                                : (Boolean) value;
```

DavideD: @Barry, how the field is mapped in the database? It seems weird to return a `BigInteger` for a boolean.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -186,6 +187,16 @@ public String getString(String columnLabel) {
 
 	@Override
 	public boolean getBoolean(String columnLabel) {
```
DavideD, line 189: Assuming that this change is correct, it should also be applied to `getBoolean(int colPos)`.
It's probably safer to collect the logic in a private method

DavideD (commented):
Nice, we are making progress!
Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: I mean, it seems like it would have to be a bug of some sort in the client....

DavideD: Yes, I think so, but it's hard to say without knowing which test fails without this change.

DavideD: To be honest, this seems a weird error for a client. It could also be that we are using the wrong type descriptor for a type in Oracle somewhere

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: 'twas an attempt to convert to boolean assuming the value is 1 or zero (true or false).  In the test value = "1"

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: Just tried something but commented it out for now.

I'm finding that Oracle's vertx client isn't setting column descriptors when there's an update with "remove" query. Bunch of these failures still there. 16 tests failures I think.
```
        Caused by:
        java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
            at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
            at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
            at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
            at java.base/java.util.Objects.checkIndex(Objects.java:372)
            at java.base/java.util.ArrayList.get(ArrayList.java:459)
            at org.hibernate.reactive.adaptor.impl.ResultSetAdaptor$1.getColumnType(ResultSetAdaptor.java:390)
            at org.hibernate.loader.custom.JdbcResultMetadata.getHibernateType(JdbcResultMetadata.java:68)
            at org.hibernate.loader.custom.ScalarResultColumnProcessor.performDiscovery(ScalarResultColumnProcessor.java:45)
            at org.hibernate.loader.custom.CustomLoader.autoDiscoverTypes(CustomLoader.java:494)
            at org.hibernate.reactive.loader.custom.impl.ReactiveCustomLoader.discoverTypes(ReactiveCustomLoader.java:123)
            at org.hibernate.reactive.loader.ReactiveLoader.lambda$doReactiveQueryAndInitializeNonLazyCollections$0(ReactiveLoader.java:76)
            at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1072)
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: OK, but ordinarily it's the job of the Hibernate `Type`s to handle this sort of translation. Where a `boolean` is mapped to a numeric column, `NumericBooleanType` deals with this.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: I haven't run any "get datatype" queries against Oracle yet... and yes it seems weird

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: good suggestion

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: @blafond what's the test that's failing?

DavideD: I gave a quick look and the table is created with this query ([testSubclassViaAssociation](https://github.com/hibernate/hibernate-reactive/blob/284920c659e91d6ecb11b91320bb18a2c627682f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SingleTableInheritanceTest.java#L85)):

```
create table BookST (DTYPE varchar2(31 char) not null, id number(10,0) not null, published date, title varchar2(255 char), forbidden number(1,0), primary key (id))
```

The column `forbidden` is a primitive boolean and it's mapped as `number(1,0)`.
So I don't think the client is wrong returning a `BigInteger` but there is something not quite right with our types.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: Ok, but why do you need it?

DavideD: Ah, wait, nevermind

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: ... but not sure how to set up a test to test `colPos`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
gavinking: Ah OK, thanks, now I see.

So now I recall that Oracle simply doesn't have a `boolean` type (which is nuts, but whatever).

So I guess this is simply another case where the Oracle JDBC driver would do this conversion automatically, but the Vert.x client doesn't.

gavinking: So @blafond, here's what I would do: I would try to call `getBoolean()` *first*, and if it throws, catch the exception and try calling `getInteger()` and converting the result. If that also fails, regrow the original exception.

blafond: k. thx

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: @tsegismon Does it make sense for `io.vertx.sqlclient.SqlResult#columnDescriptor()` to be empty?

`SqlResult#columnsNames()` return the expected result.
The query doesn't return any result (that's correct).

This is the query:
```
SELECT * FROM Person_phones where Person_id = ?
```

and  the table:
```
create table Person_phones (Person_id number(10,0) not null, "number" varchar2(255 char), "type" varchar2(255 char))
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: yep. that worked fine :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: Let's worry first about the tests that we have and are failing.

But I like how you think  :-)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
DavideD: I've created an issue for Vert.x SQL client: https://github.com/eclipse-vertx/vertx-sql-client/issues/1068

tsegismont:
👍

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
tsegismont: I'll take a look asap, thank you @DavideD

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -186,9 +186,25 @@ public String getString(String columnLabel) {
...
+			return !wasNull && bool;
+		}
+		catch (ClassCastException cce) {
+			// Since Oracle doesn't support an actual boolean/Boolean datatype
+			// Oracle8iDialect in ORM registers the BOOLEAN type as a 'number( 1, 0 )'
+			// get the int value and convert to expected boolean type for reactive
+			try {
+				int value = getInt( columnLabel );
+				boolean returnValue = BigDecimal.ZERO.intValue() == value ? false : true;
+				return returnValue;
```
gavinking, line 201: I would just go with `return value!=0` here. Its a primitive int.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -727,9 +744,8 @@ public Blob getBlob(String columnLabel) {
 			return null;
 		}
 
-		byte[] bytes = value instanceof String
-				? ( (String) value ).getBytes()
-				: ( (Buffer) value ).getBytes();
+		byte[] bytes = value instanceof String ? ( (String) value ).getBytes()
+				: value instanceof byte[] ? (byte[]) value : ( (Buffer) value ).getBytes();
 		return BlobProxy.generateProxy( bytes );
```
gavinking, line 733: I would prefer to see this written out as a chain of `if`s with a specific case to handle the possibility that it's none of the types you're expecting.

A CCE from the typecast to `Buffer` isn't a good way to report the unexpected type.

DavideD:
@tsegismont I suppose there is now way to reduce the log generated when we try to drop a table that doesn't exist, isn't it? We see a whole stack trace each time

tsegismont:
No. Oracle doesn't have the IF EXISTS support.

You can either query the system tables before dropping or create a stored
procedures that catches the specific exception.

DavideD:
But does it have to print the whole stacktrace? This is an example of the generated log:
```
java.sql.SQLSyntaxErrorException: ORA-00942: table or view does not exist

	at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:628)
	at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:562)
	at oracle.jdbc.driver.T4C8Oall.processError(T4C8Oall.java:1207)
	at oracle.jdbc.driver.T4CTTIfun.receive(T4CTTIfun.java:727)
	at oracle.jdbc.driver.T4CTTIfun.lambda$receiveRPCAsync$2(T4CTTIfun.java:426)
	at oracle.jdbc.internal.CompletionStageUtil.handleNormalCompletion(CompletionStageUtil.java:176)
	at oracle.jdbc.internal.CompletionStageUtil.lambda$normalCompletionHandler$1(CompletionStageUtil.java:227)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1072)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2073)
	at oracle.jdbc.driver.T4CTTIfun.lambda$receiveRPCAsync$0(T4CTTIfun.java:406)
	at oracle.jdbc.driver.RestrictedLock.lambda$runUnrestricted$0(RestrictedLock.java:253)
	at oracle.jdbc.driver.RestrictedLock.callUnrestricted(RestrictedLock.java:273)
	at oracle.jdbc.driver.RestrictedLock.runUnrestricted(RestrictedLock.java:252)
	at oracle.jdbc.driver.PhysicalConnection.lambda$createInternalCodeExecutor$7(PhysicalConnection.java:11647)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at oracle.jdbc.driver.PhysicalConnection.lambda$createInternalCodeExecutor$8(PhysicalConnection.java:11644)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
Caused by: Error : 942, Position : 11, Sql = drop table pig cascade constraints, OriginalSql = drop table pig cascade constraints, Error Msg = ORA-00942: table or view does not exist

	at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:632)
	... 22 more
```

It seems like only this part is interesting:
```
Error : 942, Position : 11, Sql = drop table pig cascade constraints, OriginalSql = drop table pig cascade constraints, Error Msg = ORA-00942: table or view does not exist
```

DavideD:
Maybe we can disable/reduce the log with some configuration

tsegismont:
> It seems like only this part is interesting:

@DavideD do you mean we could reduce the amount of information (as the other reactive clients do) by removing the stacktrace?

In this case, yes, I believe it makes sense and please file an issue.

DavideD:
@tsegismont: https://github.com/eclipse-vertx/vertx-sql-client/issues/1073

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UUIDAsBinaryType.java
```diff
@@ -105,7 +106,7 @@ public String toString() {
 	public static class ForOtherDbsTest extends UUIDAsBinaryType {
 
 		@Rule
-		public DatabaseSelectionRule rule = skipTestsFor( MYSQL, MARIA );
+		public DatabaseSelectionRule rule = skipTestsFor( MYSQL, MARIA, ORACLE );
```
DavideD, line 109: What's the problem here? I don't understand

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -175,10 +175,19 @@
...
+						updateCounts[i++] = iterator.next().getInteger( 0 );
+					}
+					resultNext = null;
+				}
+				else {
+					do {
+						updateCounts[i++] = result.rowCount();
+						resultNext = resultNext.next();
+					} while ( resultNext != null && i < parametersBatch.size() );
+				}
```
DavideD, line 190: @Blafond I cannot remember why we need to distinguish between the two cases, could you add some comments, please?

DavideD:
I've rebased and merged these changes. I've squashed some commits together and formatted the code.

I haven't merged this commit https://github.com/hibernate/hibernate-reactive/pull/1064/commits/fe6ec11e1ca055c2556b0ff0d631263fc3a8f796 about the RAW type as I want to see [if it's sometthing on the Vert.x side](https://github.com/eclipse-vertx/vertx-sql-client/issues/1066#issuecomment-978080400)

@blafond also, please, add some comments [about this change](https://github.com/hibernate/hibernate-reactive/pull/1064/files#r756304691). I can't remember why it was needed

I've also changed the branch `wip/oracle` so that it builds Oracle as default.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
blafond: @DavideD  For the *result* object using Oracle, the *next* value is NULL but the *list* is populated.
```
result = {RowSetImpl@8350} 
 list = {ArrayList@8354}  size = 20
 updated = 20
 columnNames = {Collections$EmptyList@8355}  size = 0
 columnDescriptors = {Collections$EmptyList@8355}  size = 0
 size = 20
 next = null
 properties = null
```
The fix checks *hasNext()* first and fills the updateCounts[] array

DavideD: Ok, but what's the test about? When does this happen?

blafond: Most of the tests in `BatchingConnectionTest.java` are failing, for instance

blafond:
```
        Caused by:
        java.lang.IllegalStateException: HR000059: Number of results is greater than number of batched parameters
            at org.hibernate.reactive.pool.impl.SqlClientConnection.lambda$updateBatch$4(SqlClientConnection.java:198)
            at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
coder-github578: Is this issue resolved. I am getting java.lang.ArrayIndexOutOfBoundsException: Index -2 out of bounds for length 46

sessionFactory.withTransaction((session, transaction) -> session.createNativeQuery("Select c.* from cars c", cars.class).getSingleResult())

PR #1065, opened 2021-11-11
move a workaround specific to schema export out of ResultSetAdaptor

gavinking:


hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
```diff
@@ -404,4 +395,1010 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+			try {
+				return delegate.getLong(columnLabel);
+			}
+			catch (ClassCastException ex) {
+				// Check if the value is a String that can be converted to a Long
+				final String aString = delegate.getString(columnLabel);
+				// aString won't be null; check just because...
+				if (aString == null) {
+					throw ex;
+				}
```
DavideD, line 540: Wouldn't `Long.parseLong` already check the null case? It will also throw a `NumberFormatException("null")` that seems a better error then ClassCastException.

DavideD (commented):
Looks good!
Just a comment about the null check

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
```diff
@@ -404,4 +395,1010 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+				try {
+					return Long.parseLong(aString);
+				}
+				catch (ClassCastException exNotAString) {
```
DavideD, line 544: Is this ever going to throw a `ClassCastException`?. Shouldn't it be `NumberFormatException`?

gavinking: I guess not. I mean, I just copy/pasted Gail's code here, without thinking too hard.

gavinking: OK, I fixed this. (It was a preexisting bug.)

DavideD:
Merged! Something is failing on CI but it doesn't seem related to this

Issue #1058, opened 2021-11-08
Replace deprecated `ExpectedException.none()` with `Assert.assertThrow`


DavideD:

ReyhanPatria:
Hi! Is this for every test? I'd like to work on this if you'd assign me to it. Thanks!

DavideD:
Thank you!
I did a quick search and I've found 6 tests:
```
hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTypeForCockroachDBTest.java
hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveMultitenantNoResolverTest.java
hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/ReactiveConnectionPoolTest.java
hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java
hibernate-reactive-core/src/test/java/org/hibernate/reactive/DefaultPortTest.java
hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveStatelessProxyUpdateTest.java
```

It might be more.

Please, make sure to use this code template: https://github.com/hibernate/hibernate-ide-codestyles

The one for Hibernate ORM.

Thanks again

ReyhanPatria:
@DavideD I'm having trouble setting up the development environment, is there a guide on this?

DavideD:
We don't have a guide because we haven't had many contributors and it should be possible to import the project without any particular configuration.

I cannot help you  if you don't give some more details: which IDE are you using and what's the issue?

DavideD:
> it should be possible to import the project without any particular configuration.

Maybe I'm a bit optimistic about  this. We use Intellij IDEA and it should work in it as long as you import it as a Gradle project.
Not sure about other IDE.

ReyhanPatria:
I'm using IntelliJ as well, I was having trouble runnings the tests. Do I
have to install every type of database?

ReyhanPatria:
I'm using Java 1.8, Gradle 6.9.1, and Docker 20.10.10

DavideD:
By default, it won't start the containers automatically. 
You need to set a system property: `docker=true`

Or you can change the value in this file: https://github.com/hibernate/hibernate-reactive/blob/main/gradle.properties#L25
I've just created an issue to use docker as the default approach: https://github.com/hibernate/hibernate-reactive/issues/1060

[...]

PR #1067, opened 2021-11-13
Issue #1058 - Replaced deprecated methods

ReyhanPatria:
#1058. 
Replaced deprecated ExpectedException.none() with Assert.assertThrows(). 
Other assertions such as exception message and cause also used the Assert library for consistency.


DavideD:
Thanks @ReyhanPatria
Overall it looks good but this approach doesn't seem to work for `org.hibernate.reactive.ReactiveMultitenantNoResolverTest`. The tests don't catch the exception.

DavideD:
Closing this for now

PR #1075, opened 2021-12-06
cleanup some nasty reflective code

gavinking:
unnecessary since https://github.com/eclipse-vertx/vertx-sql-client/issues/639


DavideD (approved):
Good idea!

PR #1103, opened 2021-12-21
Reactive value generation

### LINKED ISSUE #1099: Reactive ValueGenerator<> for use with @GeneratorType?
Hello! 
I have a column I need generated from a SQL snippet during session. It uses SELECT to grab a db specific global variable and stuff it into a column on insert using     
```java 
@GeneratorType( type = GrabGlobalDbVarGenerator.class, when = GenerationTime.INSERT)
    @Column(updatable = false, nullable = false)
    private long dbFetchedGlobal;
```
How can I do this in Hibernate Reactive?  `ValueGenerator<Long>`  doesn't use reactive Session.

I can `session.unwrap(ReactiveSession.class)` but there's no return of `CompletionStage<Long>` so I don't think it will work properly

Can you point me in right direction if this is supported?

DavideD:
Fixes #1099

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
@@ -39,6 +39,7 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.event.spi.EventSource;
+import org.hibernate.internal.SessionFactoryImpl;
 import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.jdbc.Expectation;
 import org.hibernate.loader.entity.UniqueEntityLoader;
@@ -51,16 +52,25 @@ import org.hibernate.reactive.loader.entity.impl.ReactiveDynamicBatchingEntityLo
 import org.hibernate.reactive.loader.entity.impl.ReactiveEntityLoader;
 import org.hibernate.reactive.logging.impl.Log;
 import org.hibernate.reactive.logging.impl.LoggerFactory;
+import org.hibernate.reactive.mutiny.Mutiny;
+import org.hibernate.reactive.mutiny.impl.MutinySessionFactoryImpl;
+import org.hibernate.reactive.mutiny.impl.MutinySessionImpl;
 import org.hibernate.reactive.pool.ReactiveConnection;
 import org.hibernate.reactive.pool.impl.Parameters;
 import org.hibernate.reactive.session.ReactiveConnectionSupplier;
 import org.hibernate.reactive.session.ReactiveSession;
+import org.hibernate.reactive.stage.Stage;
+import org.hibernate.reactive.stage.impl.StageSessionFactoryImpl;
+import org.hibernate.reactive.stage.impl.StageSessionImpl;
+import org.hibernate.reactive.tuple.MutinyValueGenerator;
+import org.hibernate.reactive.tuple.StageValueGenerator;
 import org.hibernate.sql.Delete;
 import org.hibernate.sql.SimpleSelect;
 import org.hibernate.sql.Update;
 import org.hibernate.tuple.GenerationTiming;
 import org.hibernate.tuple.InMemoryValueGenerationStrategy;
 import org.hibernate.tuple.NonIdentifierAttribute;
+import org.hibernate.tuple.ValueGenerator;
 import org.hibernate.type.EntityType;
 import org.hibernate.type.Type;
 import org.hibernate.type.VersionType;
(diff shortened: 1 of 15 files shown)


gavinking:
OK, this patch is bigger than expected, so can you give me until tomorrow evening to look it over?

DavideD:
Sure.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
```diff
@@ -201,102 +211,119 @@ default ReactiveConnection getReactiveConnection(SharedSessionContractImplemento
...
-					).thenApply( v -> id )
-			);
+	default CompletionStage<Void> reactivePreInsertInMemoryValueGeneration(Object[] fields, Object object, SharedSessionContractImplementor session) {
+		CompletionStage<Void> stage = voidFuture();
+		if ( getEntityMetamodel().hasPreInsertGeneratedValues() ) {
+			final InMemoryValueGenerationStrategy[] strategies = getEntityMetamodel().getInMemoryValueGenerationStrategies();
+			for ( int i = 0; i < strategies.length; i++ ) {
+				final int index = i;
+				final InMemoryValueGenerationStrategy strategy = strategies[i];
+				if ( strategy != null && strategy.getGenerationTiming().includesInsert() ) {
```
gavinking, line 275: What happens to in-memory (nonreactive) value generators?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/tuple/MutinyValueGenerator.java
```diff
@@ -0,0 +1,31 @@
...
+ * </p>
+ *
+ * @see StageValueGenerator
+ */
+public interface MutinyValueGenerator<T> extends ValueGenerator<T> {
+
+	@Override
+	default T generateValue(Session session, Object owner) {
+		throw new UnsupportedOperationException( "Use generateValue(Mutiny.Session, Object) instead" );
+	}
```
gavinking, line 28: Ugh. This isn't very nice.

gavinking:
So I think we need to find some way of removing the inheritance from `ValueGeneration`.

That's a problem I had not noticed.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java
DavideD: Not sure I understand the question. But I've just moved the existing code

DavideD:
Yeah, I agree, but I cannot think of a better solution right now.

gavinking:
Right, neither can I.

DavideD:
Maybe having abstract classes instead of interfaces is slightly better in this scenario: 2998ff0f5f3ff82d1d787fbb836a5ef8b27c4442

gavinking:
Because it lets you redeclare the method `final`, you mean?

DavideD:
Yes

DavideD:
I'm also going to add the `@Incubating` annotation to clarify that's a temporary solution.
This way people people have at least a workaround for this feature.

gavinking:
> I'm also going to add the `@Incubating` annotation to clarify that's a temporary solution.
> This way people have at least a workaround for this feature.

That's a good idea.

DavideD:
Done. I'm going to merge it as it is if there is no strong opposition

gavinking:
My only question was: does it break in-memory impls of `ValueGeneration`, or do they still work?

DavideD:
How can I test it? All our tests are passing ...

gavinking:
Well, the `@CurrentUser` annotation is our test for that.

gavinking:
Sorry, I mean `@GeneratorType(type = LoggedUserGenerator.class, when = GenerationTime.ALWAYS)`

DavideD:
I've moved the code around but I don't think I've changed anything of the original logic (not on purpose at least).
The core of the change is the method [ReactiveAbstractEntityPersister#generateValue](https://github.com/DavideD/hibernate-reactive/blob/b235ae9e3071c6bbd41a6046cab84fc53d1da622/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java#L753).

If a user will use the new reactive classes, Hibernate Reactive will call the reactive methods. Implementations of `ValueGenerator` [will still work](https://github.com/DavideD/hibernate-reactive/blob/b235ae9e3071c6bbd41a6046cab84fc53d1da622/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java#L772).

We test the new classes via `GeneratedProperty*Test` but we also have a test for `ValueGenerator` via `TimestampTest` because it tests the use of the annotations `@CreationTimestamp` and `@UpdateTimestamp` (they use a valuegenerator underneath). Maybe adding a direct test for ValueGenerator would be wise though.

The drawback of allowing an implementation of `ValueGenerator` is that a user can access the ReactiveSession, something that normally would be hidden.

Overall I think our tests cover most cases (I was pleasantly surprised when I've seen the test suite fails when I've started to work on this) but  if you have some other test in mind, let me know.

DavideD:
Merged.
This should be good enough for now.

PR #1115, opened 2022-01-10
Limit concurrent Github Actions builds

yrodiere:
See https://github.com/hibernate/hibernate-orm/pull/4615

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,15 @@ on:
   pull_request:
     branches: main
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -8,6 +8,10 @@ on:
   pull_request:
     branches: main
 
+concurrency:
```
DavideD, line 11: Could you add a few comments to describe what these changes do?

yrodiere: Done

DavideD (approved):

DavideD:
Fixes #1116 

Thanks

Issue #1073, opened 2021-12-03
java.lang.IllegalStateException: Session/EntityManager is closed


Jotschi:
I'm currently writing a [PoC for Hibernate Reactive + Vert.x](https://github.com/metaloom/poc-vertx-hibernate) and encountered an issue when loading data. I see `Session/EntityManager is closed` whenever I concurrently access the REST API.

I was also able to reproduce the issue in the boiler plate howto application from Vert.x.

## Steps to reproduce

* Import and run: https://github.com/vertx-howtos/hibernate-reactive-howto
* Create one product via 
 ```
POST http://localhost:8080/products
{
	"name": "test",
	"price": 0
}
```
* Run `wrk -d 10s -c 20 http://localhost:8080/products/1`
* Observe the errors in the server log

## Stacktrace 

* https://gist.github.com/Jotschi/d82e906ecde9db4b2f2232306cff5c26

## Notes

I have tried to [create reproducer tests](https://github.com/Jotschi/hibernate-reactive-howto/tree/reproducers/src/test/java/io/vertx/howtos/hr) for this which do not require the use of `wrk` but so far it has proven hard to reliably reproduce the fault with a test.

Bumping the hibernate reactive dependency in the howto project to 1.1.0.Final did not affect the issue.

The example uses `persist(product).call(session::flush)` the javadoc states to use `#map`. I'm not that familiar with Mutiny and not sure whether both equivalent.

DavideD:
I'm having a look but the fact that the session is closed makes me think that something hasn't been chain correctly in the stream.

DavideD:
>  persist(product).call(session::flush)

yes, this is correct

DavideD:
> I'm having a look but the fact that the session is closed makes me think that something hasn't been chain correctly in the stream.

Nevermind, this is something else

markusdlugi:
Hi @Jotschi, can you maybe try to set the HR version in your project to 1.0.0.CR9 and check if the problem still happens there?

We're seeing similar issues in our applications where HR will often throw such errors when things happen concurrently, but only in Quarkus 2.4.2.Final (which uses HR 1.0.1.Final). In Quarkus 2.3.1.Final (using HR 1.0.0.CR9) everything's working perfectly fine for us.

If the behavior's the same for @Jotschi, I guess some change in between those versions must be responsible for the problems, right, @DavideD? ORM is only one revision apart in both Quarkus versions (5.6.1.Final vs. 5.6.0.Final).

[...]

PR #1118, opened 2022-01-11
[Draft] to fix "java.lang.IllegalStateException: Session/EntityManager is closed"

Sanne:
Relates to https://github.com/quarkusio/quarkus/issues/22433 and #1073

@DavideD

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
@@ -35,7 +35,7 @@ public class VertxContext implements Context, ServiceRegistryAwareService {
     public <T> void put(Key<T> key, T instance) {
         final io.vertx.core.Context context = Vertx.currentContext();
         if ( context != null ) {
-            context.put( key, instance );
+            context.putLocal( key, instance );
         }
         else {
             throw LOG.notVertxContextActive();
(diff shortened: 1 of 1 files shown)


gavinking:
Oh, right, fine, this undoes the thing I just noticed.

I think we should definitely merge this. If i understand correctly, the change made in https://github.com/hibernate/hibernate-reactive/pull/948/files would cause the Hibernate session and transaction between _all_ threads in a verticle. If that's right, then no wonder users were running into concurrency problems.

DavideD:
@gavinking, Please, check with us before merging things there are in draft
and assign to somebody else

gavinking:
Um. No, I don't think that's a reasonable thing to yell at me about. The change to use `put()` and `get()` instead of `putLocal()` and `getLocal()` was _clearly wrong_, and certainly unintended. I know because this is code I originally wrote and I know what it's supposed to do.

And the change that broke it broke Hibernate Reactive _for essentially every user_. It's _extremely_ embarrassing that we have not noticed this issue since September, and have had multiple users report multiple bugs that I assume stem from this, and we have not done anything about it until now.

So, yeah, my fault for being swapped out, for sure, but right now we need to prioritize the fuck out of getting this fix out into a release, not keep sitting around fiddling and twiddling our fingers and worrying what process is followed when merging fixes to serious bugs!

gavinking:
Seriously, we should have been on top of this *way earlier* and we weren't.

We need to be feeling quite bad about this right now.

Sanne:
> We need to be feeling quite bad about this right now.

Don't worry, I felt quite bad when I realized my mistake.

But there's no reason to rush things as we won't be able to release until tomorrow the earliest.

PR #1119, opened 2022-01-11
Upgrade to Mutiny 1.3.0

jponge:


gavinking:
Nice, I missed that :)

jponge:
It's fresh from 10 minutes ago or so 😃

Sanne:
@jponge is the same upgrade expected to land in Quarkus 2.7 ?

We need to ensure compatibility with Quarkus 2.7, so unless that upgrade is likely we should hold

jponge:
@Sanne It's got merged 😃  

See https://github.com/quarkusio/quarkus/pull/22825

PR #1123, opened 2022-01-13
Add integration-tests module

DavideD:
With a test to check the creation of multiple sessions
from different http requests

This is a draft because it doesn't contain the fix already  in main because I want to see that it fails

### DIFF
--- a/integration-tests/verticle-postgres-it/LICENSE
+++ b/integration-tests/verticle-postgres-it/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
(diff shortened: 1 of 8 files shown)


gavinking:
> This is a draft because it doesn't contain the fix already in main because I want to see that it fails

Not sure if that's going to work. Doesn't the CI implicitly try to rebase it? At least that's what I infer happens for ORM.

DavideD:
> Not sure if that's going to work. Doesn't the CI implicitly try to rebase it? At least that's what I infer happens for ORM.

Damn I didn't know that. It makes sense though. I will temporarly revert those commits on this Draft

DavideD:
It seems to work as expected.

A few key points;
* It only runs for Postgres
* I've created a separate folder where to add future additional integration tests
* Right now it runs when the `test` task is called from the project base dir. It should be fine considering that's only one test and it's not too slow but the next step would be to make it work a bit more like the integration tests with maven (I don't think gradle 6 has a similar taslk)
* There is some duplication from the test classes in core. It would be nice to be able to reuse them but I'd prefer to treat it as a separate issue
* I will have to update the README
* This test really creates 30 entitties with parallels http requests and then reads them in the same way. It's enough to make the test fail without the patch for #1073. It's unlikely but everything could happen so fast and in the right order and the test wouldn't fail. It would be nice to have a simpler way to generate the error. This is a nice test to have anyway

gavinking:
That's excellent @DavideD, thanks so much.

DavideD:
Merged, thanks everybody

PR #1126, opened 2022-01-14
Upgrade to JBoss Logging 3.4.3.Final

Sanne:
Aligning with latest in Quarkus and ORM core.

We should probably declare a centralized "libraries" section, but won't do that today.


DavideD:
Thanks

> We should probably declare a centralized "libraries" section, but won't do that today.

It would be nice

Issue #1120, opened 2022-01-12
hibernate.default_schema don't work for sequences


guy1699:
When saving with `session.persist(entity)` I get an error saying the sequence (that set the serial PK for PostgreSQL DB) does not exists.

If I have this sequence in the public schema it does find it, so I know the issue is that selecting a sequence does not consider the schema configured by hibernate.default_schema.

If I update a record it will update correctly in the right schema.

```
@Entity
@Table(name = "tenants")
public class Tenant extends AbstractAdminOLTPEntity {

    @Id
    @SequenceGenerator(name="tenants_id_seq" , allocationSize=1)
    @GeneratedValue(strategy= GenerationType.SEQUENCE, generator="tenants_id_seq")
    @Column(name = "id")
    private Long id;
```

DavideD:
Can you show us how you've mapped the entity property that uses the sequence, please?

guy1699:
I don't know how but it is working after playing around with all kinds of things in the local PostgreSQL.
But the problem was there as I describe it.

```

@Entity
@Table(name = "tenants")
public class Tenant extends AbstractAdminOLTPEntity {

    @Id
    @SequenceGenerator(name="tenants_id_seq" , allocationSize=1)
    @GeneratedValue(strategy= GenerationType.SEQUENCE, generator="tenants_id_seq")
    @Column(name = "id")
    private Long id;
```

guy1699:
@DavideD Did you find a bug or did you just throw it to the next milestone and investigate later?
Do you have idea for workaround?

DavideD:
I've just label it and assigned it to myself so that I don't forget about it, but I haven't had time to look into this yet.
I will try to create a test this week.

As a workaround, you can try to set the attribute `schema` or `catalog` in `@SquenceGenerator`:

```
@SequenceGenerator(schema = "your_schema", ...)
```

gavinking:
Fixed.

yrodiere:
Hey @gavinking @DavideD, FYI we have a claim in Quarkus [here](https://github.com/quarkusio/quarkus/issues/31926) that this wasn't fixed. It would seem that calls to `nextval` do not mention the default schema in Quarkus.

[...]

PR #1130, opened 2022-01-16
fix hibernate.default_schema and hibernate.default_catalog

gavinking:
see #1120


gavinking:
Going to merge this. Even if it's not quite the _perfectly most elegant_ implementation, it's good enough for now.

PR #1142, opened 2022-01-19
Add log message about Vert.x at start up

### LINKED ISSUE #1134: Add log message for when we create a new Vert.x instance
Should we also have one for when we recognize that there is a Vert.x instance already?
Maybe one as info level and the other as debug.

DavideD:
Fixes #1134 
    
* If not detected:
  `[INFO] HR000002: Vert.x not detected, creating a new instance`
* if detected:
  `[DEBUG] Vert.x instance detected`

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
@@ -33,6 +33,10 @@ public interface Log extends BasicLogger {
 	void startHibernateReactive();
 
 	@LogMessage(level = INFO)
+	@Message(id = 2, value = "Vert.x not detected, creating a new instance")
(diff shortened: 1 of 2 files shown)


gavinking (approved):

gavinking:
Should we have codes for the debug messages?

DavideD:
As far as I know, anything from debug and under doesn't need to have code.
That said, I'm not against it in principle but this debug message doesn't seem super helpful and most people won't see it.
Not sure if a code in this case would be helpful

DavideD:
I'm happy to move it to INFO level so that we always print something

gavinking:
No no, just wanted to make sure we were being consistent. Don't recall what we do with other debug-level log messages.

DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/1142

PR #1143, opened 2022-01-19
improve Javadoc for VertxInstance stuff

gavinking:


hibernate-reactive-core/src/main/java/org/hibernate/reactive/vertx/impl/ProvidedVertxInstance.java
```diff
@@ -13,7 +13,7 @@
 
 /**
  * An implementation of {@link VertxInstance} which allows the client
- * to provide an instance of {@link Vertx} whose lifecyle is managed
+ * to provide an instance of {@link Vertx} whose lifecycle is managed
```
DavideD, line 16: It's not part of your changes, but a few lines after this it states:

> Hibernate will destroy the {@code Vertx} instance on shutdown.

This is not true though, isn't it?
a `ProvidedVertxInstance` won't do anything to the instance when the factory is closed

DavideD: I think it's copied from `DefaultVertxInstance`, since you are alredy updating this, could fix this too, please?

gavinking: Yah, looks wrong.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -1053,10 +1053,25 @@ Hibernate Reactive session associated with the context.
 === Vert.x instance service
 
 The `VertxInstance` service defines how Hibernate Reactive obtains an instance
-of Vert.x. The default implementation just creates one the first time it's
-needed. But if your program requires control over how the Vert.x instance is
-created, or how it's obtained, you can override the default implementation and
-provide your own `VertxInstance`. Let's consider this example:
+of Vert.x when there is no instance associated with the calling thread. The
```
DavideD, line 1056: > when there is no instance associated with the calling thread

Technically, it also defines how Hibernate Reactive obtains Vert.x even if there is  one associated to the Thread.
Isn't this a bit misleading?

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -1053,10 +1053,25 @@ Hibernate Reactive session associated with the context.
...
+default implementation just creates one the first time it's needed. But if your
```
DavideD, line 1057: Doesn't it create one when the service is started? So when the SessionFactory is started

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -1053,10 +1053,25 @@ Hibernate Reactive session associated with the context.
...
+program requires control over how the Vert.x instance is created, or how it's
+obtained, you can override the default implementation and provide your own
+`VertxInstance`.
+
```
DavideD, line 1061: I'm not sure about anything written after this point.

I think there are only two things to mention:
* Hibernate Reactive will bind the to the current Vertx instance if the factory is created inside an existing context (and will log a message about it)
* Hibernate Reactive will create a new instance if none is detected when the factory is created

The user needs to do something only if this is not the behaviour they want.
There are many way to create a factory inside a context without using `runOnContext`.

documentation/src/main/asciidoc/reference/introduction.adoc
```diff
@@ -1053,10 +1053,25 @@ Hibernate Reactive session associated with the context.
...
+IMPORTANT: If your program starts Vert.x externally to Hibernate Reactive, and
+you don't tell Hibernate Reactive how to obtain this instance of Vert.x by
+specifying an implementation of `VertxInstance`, you might end up with multiple
+instances of `Vertx` in your program, resulting in confusing error messages.
```
DavideD, line 1065: Isn't this a bug from our side? Shouldn't everything work anyway even if we start a different Vert.x instance?

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Ah sorry, I mean it logs the message when a new Vertx instance is created is created

DavideD: We should also mention that when Hibernate Reactive creates a new instance of Vert.x, it will also shut it down when the factory gets closed.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: Well,  I would expect it to work if one uses `.withSession`.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: Ummmmm. Hrrrrmmmm. Yeah, right. See there's a bit of a slightly confusing interplay between `VertxContext` and `VertxInstance` here. TBH I've never completely understood the significance of:

```java
public void execute(Runnable runnable) {
    io.vertx.core.Context context = vertxInstance.getVertx().getOrCreateContext();
    if ( Vertx.currentContext() == context ) {
        runnable.run();
    }
    else {
        context.runOnContext( x -> runnable.run() );
    }
}
```

Like .... why don't we just use the `Vertx.currentContext()` if there is one? Why do we insist on it being the one coming from "our" Vert.x?

I mean, I know why we do it that way: because that's what @cescoffier and @vietj told us to do, but I think I need to understand it better.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: True. So then it's sort-of a bigger problem than I thought.

But default we go creating an instance of Vert.x even if we don't know we're going to need it.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: See what I documented is how I actually want it to work.

But you're right: it doesn't work like that today.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: If the program already has a `Vertx`, and we go and create one, then that's simply wrong, I would have thought.

documentation/src/main/asciidoc/reference/introduction.adoc
gavinking: > how I actually want it to work.

Well, naively.

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: I remember reading that one could use these scenario if they want different event busses for different groups of Verticles.
I might have dreamt it.
But it, even if it's unlikely, I'm not sure it's wrong

documentation/src/main/asciidoc/reference/introduction.adoc
DavideD: 1. When I talk about "an instance of Vert.x", I mean `io.vertx.core.Vertx`.
2. When I talk about "a Vert.x context", I mean `io.vertx.core.Context`.

So, `DefaultVertxInstance` is responsible for `1`, `org.hibernate.reactive.context.impl.VertxContext` is responsible for `2`.
But the `VertxInstance` service is always called when the factory is started (with or without an existing Vert.x instance), that's why I said that the line `when there is no instance associated with the calling thread.` doesn't seem correct (it implies that it gets called only when there is no Vert.x instance).

DavideD: > Like .... why don't we just use the Vertx.currentContext() if there is one? Why do we insist on it being the one coming from "our" Vert.x?

Doesn't it makes sure that 1. we are running in a context (it will create a new one if one doesn't exist) and 2. an existing session run the operations always in the same context and that should imply the same thread (this is the part where I'm never sure).

DavideD: I think I need to think more about this

gavinking: > I think I need to think more about this

Right, me too.

And I think perhaps we should wait until @Sanne gets back and discuss it thoroughly.

DavideD: > Doesn't it makes sure that 1. we are running in a contex

I remember in the beginning we had a lot of issues with operations not running in the right context when using Quarkus.

PR #1153, opened 2022-01-24
attempt to tolerate the use of String for generated ids

gavinking:


gavinking:
@DavideD do you know what's up with the CI build?

DavideD:
Something is wrong with the Db2 docker image. It seems it cannot download it. I haven't tested it locally

gavinking:
OK, thanks.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTypeForCockroachDBTest.java
```diff
@@ -78,8 +78,8 @@ public void longIdentityType(TestContext context) {
 
 	@Test
 	public void integerIdentityType(TestContext context) {
-		thrown.expectMessage( "too big" );
-		thrown.expectMessage( "Integer" );
+//		thrown.expectMessage( "too big" );
+//		thrown.expectMessage( "Integer" );
```
gavinking, line 82: @DavideD what was the point of these tests? What were they trying to prove?

DavideD: I think cockroachDB generates ids that are too big to be saved in an Integer or Short.
So the test is really about checking that a meaningful exception is thrown.

We should probably add a comment about it

DavideD: Why would you uncomment this?

DavideD: I guess you've also fixed this "issue" for CockroachDB

DavideD: You can remove this class and update `IndentityGeneratorTypeTest`  so that the tests are the same for every db

DavideD: See https://github.com/hibernate/hibernate-reactive/blob/41f34afb2e46f2502899a5387229a939185d338f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/IdentityGeneratorTypeTest.java#L37

gavinking: > Why would you uncomment this?

Well because with this change the tests stopped failing.

DavideD: > Well because with this change the tests stopped failing.

Yes, I've sent other messages after this.

gavinking: Yeah but what I'm asking is: where those exceptions *necessary* for some reason. Have I broken things by "fixing" it?

DavideD: This is a mapping a user might try (or have tried, I can't remember) and it makes sure that the right exception get thrown.

Now you have inadvertently changed the behaviour when using CockroachDB.
I guess in this case you have changed it for the better, so there shouldn't be any negative consequence.

I don't think anybody is using `integer` or `short` on purpose with the identity type so that they can see the exceptions. At least we know we can remove some duplicated test classes.

gavinking: OK, fine, I will remove it then.

DavideD: Wait... now that I think about it, you have removed a validation check. I guess the identifier with CockroachDB will become pretty long pretty soon and we won't be able to set it to a `short`. What will happen in that case?

gavinking: Well, `throw LOG.generatedIdentifierTooBigForTheField()`, I suppose.

gavinking: Wait, suddenly the exception happens for me again.

gavinking: I dunno, something a bit fishy there, but anyway the tests are doing what they're supposed to do again now.

gavinking:
@mjakubowski84  can you please try this out and see if it solves your problem?

mjakubowski84:
@gavinking Thanks, I will try this today

mjakubowski84:
> @mjakubowski84 can you please try this out and see if it solves your problem?

@gavinking Tested. Works like a charm!

gavinking:
OK, let's apply this. I'm not super-happy about it, but it's fine, I suppose.

mjakubowski84:
Thanks, I really appreciate it.

PR #1156, opened 2022-01-25
Disable use of Nationalized chars for MariaDb

DavideD:
@gavinking Is this is correct or there is something I'm missing?


gavinking:
That seems pretty weird. Why do we need to disable the test especially for Maria?

DavideD:
Otherwise it fails

gavinking:
Yeah well I guessed that, but *why*?

DavideD:
I'm going to merge this because it fixes the build. I will create an issue so that we don't forget to check what's going on.

DavideD:
Well, it fixes the build for MariaDB, I still have to figure out what's going on with Db2

PR #1159, opened 2022-01-26
Fix UnicodeCharsTest

### LINKED ISSUE #1158: Enabling use of nationalize chars fails for MariaDB
See https://github.com/hibernate/hibernate-reactive/pull/1156

The test `UnicodeCharsTest` fails for MariaDB when `hibernate.use_nationalized_character_data` is set to `true`.

Test configuration: https://github.com/hibernate/hibernate-reactive/blob/6ad3d50fe195b8c75b0f7c372f7721283ad24731/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UnicodeCharsTest.java#L36

DavideD:
Fixes #1158

One of the chars used in the test is not compatible with the
collation used when setting the property to true for MySQL or
MariaDB.


gavinking:
@DavideD thanks for sorting this one out.

PR #1160, opened 2022-01-27
add SessionFactory.getStatistics()

gavinking:
Initial support for statistics.

Of course, all the JDBC-level statistics are missing, so I guess we need to mess up some of our nice clean code with trash like:

```java
final StatisticsImplementor statistics = sessionFactory.getStatistics();
if ( statistics.isStatisticsEnabled() ) {
	statistics.connect();
}
```


DavideD (approved):

gavinking:
Y'know, I think I might as well merge this in it's current form, and handle the vert.x-client level statistics as a separate phase.

DavideD:
Sounds good to me

PR #1164, opened 2022-01-30
Integration with Oracle

DavideD:
* Mapping requiring an insert and select identifier aren't going to work. Example: Join and Union inheritances 
* Filtering for UUID doesn't work
* The url needs to be in the form of `[jdbc:]oracle:thin:[username/password]@host:port/databasename`. I think we also need to also add support for `host:port:SID`


gavinking:
Awesome!

> Mapping requiring an insert and select identifier aren't going to work. Example: Join and Union inheritances

can you expand on this a little?

DavideD:
If I'm not mistaken, those type of mappings require a temporary table. When Hibernate Reactive insert an element in the table generates an id that it cannot read because the SQL doesn't work.

Enabling `JoinedSubclassInheritanceTest` or `UnionSubclassInheritanceTest` will show you the failures:
```
java.util.concurrent.CompletionException: io.vertx.core.VertxException: [SQL INCLUDES EXPRESSIONS ADDED BY THE ORACLE JDBC DRIVER],Error : 933, Position : 119, Sql = insert into ht_BookJS select joinedsubc0_.id as id from BookJS joinedsubc0_ where joinedsubc0_.title='Necronomicon II' RETURNING ROWID INTO :1 , OriginalSql = insert into ht_BookJS select joinedsubc0_.id as id from BookJS joinedsubc0_ where joinedsubc0_.title='Necronomicon II' RETURNING ROWID INTO ?, Error Msg = ORA-00933: SQL command not properly ended
```

gavinking:
> If I'm not mistaken, those type of mappings require a temporary table. 

Oh. Well, they only require a temp table for bulk `update` and `delete` queries.

DavideD:
This is the issue on Vert.x: https://github.com/eclipse-vertx/vertx-sql-client/issues/1108

DavideD:
There something wrong with it on CI. I will merge this once it builds correctly

PR #1166, opened 2022-01-31
Test with JDK 19-ea+7

yrodiere:


yrodiere:
It looks like Gradle cannot run a Java 19 toolchain for some reason... I'll need to dig a bit.

Hey @DavideD I see reactive is still using Gradle 6.9, which is kinda old (May 2021). Any plans to upgrade soon-ish? (see [releases](https://gradle.org/releases/))

DavideD:
I've sent a PR for Gradle 6.9.2:
https://github.com/hibernate/hibernate-reactive/pull/1170

yrodiere:
> I've sent a PR for Gradle 6.9.2: #1170

Ok, thanks. I guess upgrading to 7.x was too much work?

I rebased on main, let's see how it goes.

DavideD:
6.9.2 was fast and easy. I will check if we can upgrade to 7 soon

DavideD:
Merged via https://github.com/hibernate/hibernate-reactive/pull/1177

PR #1169, opened 2022-01-31
Trying to fix the build on CI

DavideD:


DavideD:
It seems that everything works now. Maybe something was wrong with GitHub?

gavinking:
Who knows?

DavideD:
I think I will merge this anyway. I find the code a bit nicer and the build of the pr passes

DavideD:
Merged

PR #1174, opened 2022-02-02
Fix issue persisting entity mapped with non-nullable association 

### LINKED ISSUE #1077: Persist of entity with non nullable association causes UnsupportedOperationException
Mapping example:
```
@JoinColumn(name = "device_id", nullable = false)
@ManyToOne(fetch = FetchType.LAZY, targetEntity = Device.class, optional = false)
private Device device;
```

Setting either `nullable = false` or `optional = false` will cause the following exception during a persist:
```
Caused by: java.lang.UnsupportedOperationException: reactive persistence context
	at org.hibernate.reactive.engine.impl.ReactivePersistenceContextAdapter.getDatabaseSnapshot(ReactivePersistenceContextAdapter.java:76)
	at org.hibernate.engine.internal.ForeignKeys.isTransient(ForeignKeys.java:306)
	at org.hibernate.engine.internal.ForeignKeys$Nullifier.isNullifiable(ForeignKeys.java:230)
	at org.hibernate.engine.internal.ForeignKeys$Nullifier.access$000(ForeignKeys.java:36)
	at org.hibernate.engine.internal.ForeignKeys.collectNonNullableTransientEntities(ForeignKeys.java:414)
	at org.hibernate.engine.internal.ForeignKeys.findNonNullableTransientEntities(ForeignKeys.java:385)
	at org.hibernate.action.internal.AbstractEntityInsertAction.findNonNullableTransientEntities(AbstractEntityInsertAction.java:91)
	at org.hibernate.reactive.engine.ReactiveActionQueue.addInsertAction(ReactiveActionQueue.java:346)
	at org.hibernate.reactive.engine.ReactiveActionQueue.addAction(ReactiveActionQueue.java:331)
```

blafond:
Fixes #1077

Requires minor change to [NonNullableTransientEntities](https://github.com/hibernate/hibernate-orm/blob/798f9b05184bd0b14bdf13841b127ddef912a62d/hibernate-core/src/main/java/org/hibernate/engine/internal/NonNullableTransientDependencies.java#L27).

Merge after [PR](https://github.com/hibernate/hibernate-orm/pull/4772) is merged and after newer ORM build

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
@@ -34,7 +34,6 @@ import org.hibernate.action.spi.AfterTransactionCompletionProcess;
 import org.hibernate.action.spi.BeforeTransactionCompletionProcess;
 import org.hibernate.action.spi.Executable;
 import org.hibernate.cache.CacheException;
-import org.hibernate.engine.internal.NonNullableTransientDependencies;
 import org.hibernate.engine.spi.ActionQueue;
 import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.ExecutableList;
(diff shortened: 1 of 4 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
```diff
@@ -326,12 +325,13 @@ public void clear() {
 		}
 	}
 
-	public CompletionStage<Void> addAction(ReactiveEntityRegularInsertAction action) {
+	public CompletionStage<Void> addAction(ReactiveEntityInsertAction action) {
```
DavideD, line 328: Why did you change the signature of this method?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+	@Entity(name = "Painting")
+	@Table(name = "painting")
+	public static class Painting {
+		@Id
+		public Long id;
+		String name;
+
+		@JoinColumn(name = "author_id", nullable = false)
+		@ManyToOne(fetch = FetchType.LAZY, targetEntity = Artist.class) //, optional = false)
```
DavideD, line 72: Maybe you could add another association that has `optional = false`. This way we are sure that both settings are working

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
```diff
@@ -0,0 +1,123 @@
...
+		test( context, deleteEntities( "Painting", "Artist" ) );
+	}
+
+	@Test
+	public void testNonNullableSuccess(TestContext context) {
+		test( context,
+				openMutinySession().chain( session -> session
+								.createQuery( "from Artist", Artist.class )
+								.getSingleResult() )
+						// We are checking `.getPaintings()` to verify success in persistence
```
DavideD, line 59: You are not checking anything, though. Not even that it's not null

DavideD (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/ReactiveActionQueue.java
blafond: mistake. fixed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
blafond: fixed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
blafond: fixed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
```diff
@@ -0,0 +1,154 @@
...
+							context.assertEquals( 1, paintings.size() );
+							context.assertEquals( "Mona Lisa", paintings.get( 0 ).name );
+						} )
+						.chain( () -> getMutinySessionFactory().withTransaction( s -> session
+										.createQuery( "from Dealer", Dealer.class )
+										.getSingleResult().chain( d -> session.fetch( d.paintings ) )
+										.invoke( paintings -> {
+											context.assertNotNull( paintings );
+											context.assertEquals( 1, paintings.size() );
+											context.assertEquals( "Mona Lisa", paintings.get( 0 ).name );
```
DavideD, line 74: This indentation is weird. It makes it look like the second `withTransaction` is inside the first one (even if it's not the case).
It's not critical but since we have to wait for the next ORM release anyway, could you align them correctly, please? (Or maybe create two different methods)

DavideD (approved):

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NonNullableManyToOneTest.java
DavideD: Actually, something is wrong in the stream.

blafond: fixed alone with correcting sessions in test.

PR #1181, opened 2022-02-04
Enable WrongCredentialsTest for MySQL

### LINKED ISSUE #511: UnsupportedOperationException on CI with MySql when credentials are wrong
I've updated the test WrongCredentialsTest so that it runs for every datastore. It seems to work fine locally but when it runs on CI  with MySQL and the latest ORM Snapshot it throws an UnsupportedOperationException instead of the expected one:
```
java.lang.UnsupportedOperationException: Unsupported authentication method: sha256_password
```

I cannot replicate this issue locally and I'm not sure what causing it on CI.
This is the test class: 
https://github.com/DavideD/hibernate-reactive/blob/d131bb2b38662b3bc64ee88faae42a9c223e5a70/hibernate-reactive-core/src/test/java/org/hibernate/reactive/WrongCredentialsTest.java

On CI it seems to happen with the latest ORM snapshot but I don't think it's related.
Here's the failed workflow: https://github.com/hibernate/hibernate-reactive/runs/1687501809?check_suite_focus=true#step:5:314

See also: https://github.com/eclipse-vertx/vertx-sql-client/issues/1132

DavideD:
Fixes #511 
I had to set the authentication policy at start up. 
I don't know why it solved the issue because it should be the default setting.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/WrongCredentialsTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/WrongCredentialsTest.java
@@ -5,18 +5,17 @@
  */
 package org.hibernate.reactive;
 
+import javax.persistence.Entity;
 import javax.persistence.Id;
 
 import org.hibernate.cfg.Configuration;
 import org.hibernate.reactive.provider.Settings;
-import org.hibernate.reactive.testing.DatabaseSelectionRule;
 
-import org.junit.Rule;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


gavinking:
> I don't know why it solved the issue because it should be the default setting.

Perhaps that's something you could report to the Vert.x client project?

DavideD:
Done: https://github.com/eclipse-vertx/vertx-sql-client/issues/1132

DavideD:
CLosing this pr because it doesn't actully fix the issue

PR #1185, opened 2022-02-08
Trying to figure out the failures in the Oracle build

DavideD:
For checking what's happening on CI with oracle


DavideD:
I can recreate the error locally but I haven't found a fix yet.
No need this pr

PR #1186, opened 2022-02-08
Add some logging in relation to reuse of open Mutiny.Session found in context

Sanne:
[trivial]
Adding the ability to debug when a contextual session is being reused


DavideD:
Sure, but could you also add it to the Stage API?

Sanne:
completed this: added the statelessession as well, and the multi-tenancy variants.
Then both Mutiny and Stages.

PR #1197, opened 2022-02-14
Enable SequenceGeneratorDefaultSchemaTest for Oracle

DavideD:
We don't have the permissions for the creation of a new schema with Oracle but one is created anyway when a user is created.
I think it's goog enough for testing for now


DavideD:
I'm going to merge this pr because the failures are not related to test I've enabled

PR #1217, opened 2022-02-25
Refactor UriPoolConfiguration

DavideD:
We use this class to test that we can override how we
get the connection options.

But now we don't need to update it every time we add
a new database.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UriPoolConfiguration.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UriPoolConfiguration.java
@@ -5,19 +5,12 @@
  */
 package org.hibernate.reactive;
 
-import io.vertx.db2client.DB2ConnectOptions;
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/UriPoolConfiguration.java
```diff
@@ -27,21 +20,6 @@ public PoolOptions poolOptions() {
...
-			case "db2":
-				return DB2ConnectOptions.fromUri( uri.toString() );
-			case "sqlserver":
-				return MSSQLConnectOptions.fromUri( uri.toString() );
-			case "oracle":
-				return OracleConnectOptions.fromUri( uri.toString() );
-			default:
-				throw new IllegalArgumentException( "Database not recognized: " + dbType() );
-		}
+		return SqlConnectOptions.fromUri( uri.toString() );
```
gavinking, line 23: Nice.

PR #1219, opened 2022-02-25
Jakarta EE version

Sanne:
N.B. Draft

This is not exactly how I envision we'll migrate to Jakarta EE; I'd actually aim to have these changes produced automatically by transforming the build during the release, like we did for Hibernate ORM core.

But I needed a prototype, sharing it here as it might be interesting for all to have a look. It compiles, but untested as it seems docker doesn't work today :)

Also worth noting: the first commit [bf93df4](https://github.com/hibernate/hibernate-reactive/pull/1219/commits/bf93df46140c29b5090329c0a73538a7e96c9f5a) includes only changes automatically generated by Eclipse Transformer.


DavideD:
Are you sure is docker the issue? It seems something is not quite right with the configuration of the persistence.xml we use in the examples and [the configuration of the cache](https://github.com/hibernate/hibernate-reactive/runs/5340317132?check_suite_focus=true#step:5:390) we use in CacheTest.

Sanne:
I had two mistakes:
 - a whitespace issue in the xmlns header of the example `persistence.xml` files
 - forgot to rename our provider service ([4874ee0](https://github.com/hibernate/hibernate-reactive/pull/1219/commits/4874ee016f70134356f42ed71e6f39c4a1922e83))

Should work better now

Sanne:
And yeah docker is still not working on my laptop - not sure why. Working on the other machine now.

Issue #333, opened 2020-08-27
simplify pool creation for Vert.x client 3.9.2


gavinking:
@aguibert did some work to automate the selection of the database driver when creating a `Pool`. Currently we still have a legacy workaround for 3.9.1. We need to clean that up.

gavinking:
https://github.com/eclipse-vertx/vertx-sql-client/issues/616

gavinking:
Hrrrrm. @aguibert I was looking at this fragment of code:

```java
		catch (ServiceConfigurationError e) {
			// Backup option if multiple drivers are on the classpath.
			// We will be able to remove this once Vertx 3.9.2 is available
			return findDriver( uri, e ).createPool( vertx, connectOptions, poolOptions );
		}
```

It seems to imply that we should now be able to delete the `findDriver()` method.

But actually it doesn't seem to be the case. On 3.9.2, it seems that the fallback stuff is still needed.

Can you clarify, please?

gavinking:
Hrrrrm, I've just noticed that `DB2ConnectOptions`, `MySQLConnectOptions`, and `PgConnectOptions` all have a `fromUri()` method which parses a connection string.

I wonder if it would be better to use these connection strings that the JDBC URLs we've been using up to now.

It would be ideal if Vert.x offered a `SqlConnectionOptions.forUri(String)` method, but even without that we could iterate over the loaded `Driver`s and find one which accepts the given connection string.

WDYT @aguibert @DavideD @Sanne?

gavinking:
> I wonder if it would be better to use these connection strings than the JDBC URLs we've been using up to now.

I've actually demonstrated what that would look like, via the new `SqlClientPoolConfiguration` (which I added for a different reason).

https://github.com/hibernate/hibernate-reactive/blob/42937a41097a0e974c06d134f956488df7a4fecc/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UriPoolConfiguration.java

gavinking:
I need https://github.com/eclipse-vertx/vertx-sql-client/issues/767 to make progress on this.

DavideD:
@gavinking Do you plan to work on this? Otherwise I think @blafond can start to have a first look

gavinking:
Nah, I tried a while ago and I didn't have what I needed, but perhaps we do now with the new stuff in latest Vert.x.

gavinking:
I mean to say no I'm not working on it :)

DavideD:
[...]

PR #1221, opened 2022-02-28
Only accept Vert.x style url

DavideD:
See #333


gavinking:
Not sure this is the kind of thing we should do in a point release.

DavideD:
Yes, I'm not sure either. But I wanted to see the impact on the code base.

Issue #1108, opened 2022-01-04
A way to pass SSL mode to the Vertx pgClient


guy1699:
Today there is no way to pass SSL mode by property or URL param when using Postgres.
Correct me if I'm wrong, the only client available is Vertx and it is using `io.vertx.pgclient.PgConnectOptions`.

The constructor of PgConnectOptions does not inherit the SSL mode from the SqlConnectOptions which Hibernate sends.
https://github.com/eclipse-vertx/vertx-sql-client/blob/352e8da205598ffd9730ef36891dbc28ff544e53/vertx-pg-client/src/main/java/io/vertx/pgclient/PgConnectOptions.java#L139

The DefaultSqlConnectOptions also does not seem to have an option for SSL to be configured, is that correct?

blafond:
@guy1699  I'm working on #1059 with the intent on capturing all valid properties on a URL, including for PostgreSQL. Vertx sql clients provide db-specific URL parsing ( see [PgConnectionUriParser](https://github.com/eclipse-vertx/vertx-sql-client/blob/352e8da205598ffd9730ef36891dbc28ff544e53/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionUriParser.java#L185) ).

By leveraging vertx (and tweaking their parsers for general properties) we'll should have what you need.

blafond:
Thx for the input. I think my fix will require a tweak to Vertx SQL client parsers. Created issue:  [#1115](https://github.com/eclipse-vertx/vertx-sql-client/issues/1115)

DavideD:
Isn't this a Vert.x issue?

In Hibernate Reactive we allow passing a custom configuration using `hibernate.vertx.pool.configuration_class` (like you have already did) or, for other advance use cases, pass a cusomt pool using `hibernate.vertx.pool.class` ([it's not described in the documentation yet though](https://github.com/hibernate/hibernate-reactive/issues/1112)).

@blafond Is there anything we need to do on the Hibernate Reactive front? Or this is going to be solved by https://github.com/hibernate/hibernate-reactive/issues/1059?

guy1699:
@DavideD, can you describe how to use a custom pool with `hibernate.vertx.pool.class` ?
Is this different than what I did?

DavideD:
It's different because you would have to implement `ReactiveConnectionPool` instead. What you have done seems fine.

DavideD:
[...]

PR #1225, opened 2022-03-09
Include connection parameters during the creation of SqlConnectOptions

### LINKED ISSUE #864: Pass the parameters in the url as properties when the connection is created
Each Vert.x Driver accept different additional parameter keys in the URL.
For example, in [PostgreSQL](https://vertx.io/docs/vertx-pg-client/java/#_connection_uri):
```
host
hostaddr
port
user
password
dbname
sslmode
properties including(application_name, fallback_application_name, search_path)
```

I think that when we create the `SqlClientOptions` in [DefaultSqlClientPoolConfiguration](https://github.com/hibernate/hibernate-reactive/blob/26e0aef99edaef45f29b3ee985dbe976b4e72b73/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L151) we don't keep track of them.

We should test this.

### LINKED ISSUE #1059: Test SqlConnectOptions creation from URL
See [JdbcUrlParserTest](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java#L43)

We have a simple test for Postgres but the logic changes a lot based on the database selected.
We should make sure to test that we create the right `SqlConnectOptions` for each database.

This is the method that creates the SqlConnectOptions object: https://github.com/hibernate/hibernate-reactive/blob/cd8d50c851a2a3a528e77555c214fca20693a7a8/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L92

Relates to #864

DavideD:
Fix #1059 
Fix #864
and maybe also #1108

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
@@ -7,6 +7,7 @@ package org.hibernate.reactive.pool.impl;
 
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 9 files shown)


DavideD:
I want to change something before merging this

PR #1226, opened 2022-03-10
 Include connection parameters during the creation of SqlConnectOptions #1225 

### LINKED ISSUE #864: Pass the parameters in the url as properties when the connection is created

### LINKED ISSUE #1059: Test SqlConnectOptions creation from URL

DavideD:
Fix #1059
Fix #864 

It now also uses the right default port for CockroachDB if none is defined

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPool.java
@@ -216,7 +216,7 @@ public class DefaultSqlClientPool extends SqlClientPool
 	 * @return the disambiguated {@link Driver}
 	 */
 	private Driver findDriver(URI uri, ServiceConfigurationError originalError) {
-		String scheme = uri.getScheme(); // "postgresql", "mysql", "db2", etc
+		String scheme = scheme( uri );
 		for ( Driver d : ServiceLoader.load( Driver.class ) ) {
 			String driverName = d.getClass().getCanonicalName();
 			LOG.detectedDriver( driverName );
@@ -224,7 +224,14 @@ public class DefaultSqlClientPool extends SqlClientPool
 				return d;
 			}
 		}
-		throw new ConfigurationException( "No suitable drivers found for URI scheme: " + uri.getScheme(), originalError );
+		throw new ConfigurationException( "No suitable drivers found for URI scheme: " + scheme, originalError );
+	}
+
+	private String scheme(URI uri) {
+		// For CockroachDB we use the same driver we sue for Postgres
+		return uri.getScheme().toLowerCase().startsWith( "cockroach" )
+				? "postgresql"
+				: uri.getScheme();
 	}
 
 	private boolean matchesScheme(String driverName, String scheme) {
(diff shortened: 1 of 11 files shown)


pendula95:
@DavideD 
I run on unexpected behavior while trying to make @ColumnTransformer that uses function pgp_sym_encrypt from added extension `pgcrypto`
I expected that correct search_path was propagated from hibernate.default_schema but looks like this is not the case.
Schema for this function is not propagated and it fails to find it as search_path is always wrong and not respecting the configuration param.
After checking with vertx pg client I see that it has a correct param search_path that can be set via params or via url.
So I try to do this buy specifying ?search_path on the end of my jdbc url. This does not work also and the  query I used to debug before `show search_path` again gives wrong value.
Then I checked how hibernate handles custom params and I see this really strange line that will check additional params only if username and password are null. 
https://github.com/hibernate/hibernate-reactive/blob/f17e8a1e80c70329baaf3d98fc7f229bf68b02e7/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L111-L117

Because I use hybrid approach where only host information is passed via jdbcurl and username/password/schema are passed via property values this additional values like search_path are totally ignored.

So from here I think there are 2 issues:

1) `hibernate.default_schema` is not propagated and used with fucntions that need to be looked up in correct schema.
2) Additional connection params are ignored when username and password are passed as properties.

After tweaking my tests that so that username is passed via param, password via url, rest of the params like (search_path) are respected and my @ColumnTransformer works as expected and this function is found.

In order to reproduce the issue:
create pg database
delete public schema
create some other schema
CREATE EXTENSION pgcrypto
try to use any of  pgcrypto functions

I have simple reproducers for all but just by looking at the code for parsing of params the issue 2) is simple to fix and reproduce.
Issue 1 I have not looked at code base so I am not sure If I am even right if schema should be propagated to search_path but I can open up another thread for this.

PR #1249, opened 2022-03-21
[DON'T MERGE] Testing publishing the docs on hibernate.org/staging

DavideD:
I was sure I've tested it but something doesn't work and the release fail.


DavideD:
I will create a temporary tag upstream. Sorry for the noise

PR #1251, opened 2022-03-22
Restore correct configuration for UTCTest

DavideD:
When I refactor the code I've delete the configuration without noticing that it sets some properties.
The test didn't fail for me because the time zone matches, I guess


blafond (commented):
works

PR #1255, opened 2022-03-29
Move to oracle-actions/setup-java to install newer JDKs

yrodiere:
See https://inside.java/2022/03/11/setup-java/

Major gain: we won't need to update the download links anymore :)


DavideD:
Awesome! Thanks

Issue #1256, opened 2022-03-30
Why ProxyConnection can not use batch?


cdmikechen:
I use `quarkus-hibernate-reactive` to insert some rows to table, but it seems like that quarkus does not support reactive batch mode at present. I have written some comments in quarkus issue https://github.com/quarkusio/quarkus/issues/16640 
So I start to try to start batch session manually.

I tried many ways, and finally succeeded in this way:
```java

    @Inject
    Mutiny.SessionFactory sessionFactory;

    @POST
    @Path("/batch")
    // @ReactiveTransactional
    public Uni<Void> batch() {
        List<Fruit> fruits = new ArrayList<>();
        for (long i = 0, row = 300; i < row; i++) {
            Fruit fruit = new Fruit();
            fruit.id = i + 1;
            fruit.name = "test";
            fruits.add(fruit);
        }

        return sessionFactory.withTransaction(session ->
                        session.setBatchSize(50).persistAll(fruits.toArray())
                );
//        return Fruit.getSession()
//                .chain(session -> session.persistAll(fruits.toArray()));
    }
```

Fruit is a PanacheEntityBase:
```java
@Entity
public class Fruit extends PanacheEntityBase {

    @Id
    @Column(name = "id")
    public Long id;

    @Column(length = 40, unique = true)
    public String name;

}
```

But, if I use Panache Session, session can not use batch. So I check hibernate-reactive codes and I found that Panache use `ProxyConnection`. in `ProxyConnection` allowBatching can not used and set to false. I try to change `false` to `allowbatching` and rebuild jar, I find that hibernate-reactive can perform batch insert correctly.

https://github.com/hibernate/hibernate-reactive/blob/dede54be9a38b4d904053624ed57922e5f6e2441/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ProxyConnection.java#L98-L105

So I want to know why `ProxyConnection` can not use batch?

DavideD:
I think it's a mistake and we should pass allowBatching instead

DavideD:
Please, feel free to send a PR

cdmikechen:
@DavideD 
Thanks for your reply, I've opened a PR https://github.com/hibernate/hibernate-reactive/pull/1257

DavideD:
Rebased and merged.
Thanks a lot @cdmikechen

PR #1257, opened 2022-03-30
ProxyConnection can support batch update

cdmikechen:
Issue https://github.com/hibernate/hibernate-reactive/issues/1256
The ProxyConnection update method forces `allowbatching` to `false`, which makes batch update impossible. ProxyConnection is still used in quarkus, so I think it may be recommended to support batch update.


DavideD:
Thanks a lot @cdmikechen 
I will merge this in a bit

DavideD:
Merged after a rebase

Issue #1235, opened 2022-03-15
Specify the repository where to download the docker images


DavideD:
Docker doesn't always use docker hub as the default repository. 
If the configuration contains multiple repositories, the build with reactive will fail because we use the short name of the image.

This issue is about specifying the long name of the image, for example:
```
docker.io/mariadb:10.7.1
```
Progress:
- [X] MariaDB
- [X] Oracle
- [X] MS SQL Server (Not necessary...probably)
- [X] MySQL
- [X] CockroachDB
- [X] PostgreSQL
- [X] Db2

DavideD:
@Sanne how come mariadb was the only one causing problems? It's because you already had the other images locally?

Sanne:
Yes. It's the only one that  (I didn't have locally) AND (can be fetched from multiple sources)

Sanne:
N.B. it needs to be `docker.io/mariadb:10.7.1` and not `docker.io/library/mariadb:10.7.1`. I had fixed the description but it seems you reverted it.

DavideD:
Fixed back

DavideD:
Fixed, thanks!

PR #1259, opened 2022-04-08
specified default repository for docker image downloads in test DB's

blafond:
see #1235 

Tweaked remaining test DB containers to default to `docker.io/` prefix

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/CockroachDBDatabase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/CockroachDBDatabase.java
@@ -9,6 +9,7 @@ import org.hibernate.HibernateException;
 
 import org.testcontainers.containers.CockroachContainer;
 import org.testcontainers.containers.Container;
+import org.testcontainers.utility.DockerImageName;
 
 import static org.testcontainers.shaded.org.apache.commons.lang.StringUtils.isNotBlank;
 
@@ -16,7 +17,8 @@ class CockroachDBDatabase extends PostgreSQLDatabase {
 
 	public static CockroachDBDatabase INSTANCE = new CockroachDBDatabase();
 
-	public final static String IMAGE_NAME = "cockroachdb/cockroach:v21.2.4";
+	public static final String IMAGE_NAME = "cockroachdb/cockroach";
+	public static final String IMAGE_VERSION = ":v21.2.4";
 
 	/**
 	 * Holds configuration for the CockroachDB database container. If the build is run with <code>-Pdocker</code> then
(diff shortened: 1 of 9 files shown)


DavideD:
It seems it doesn't work for Sql Server. Maybe it's not needed in that case because the docker image is already `mcr.microsoft.com/mssql/server:2019-latest`.

blafond:
Locally, I removed the docker image and re-ran the build with the CI's command and it passed.

What's different about the CI build/environment?

DavideD:
Did you run the build with sql server?

blafond:
`./gradlew clean build -PshowStandardOutput -Pdocker -Pdb=MSSQLServer`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/CockroachDBDatabase.java
```diff
@@ -25,7 +26,8 @@
 	 * TIP: To reuse the same containers across multiple runs, set `testcontainers.reuse.enable=true` in a file located
 	 * at `$HOME/.testcontainers.properties` (create the file if it does not exist).
 	 */
-	public static final CockroachContainer cockroachDb = new CockroachContainer( IMAGE_NAME )
+	public static final CockroachContainer cockroachDb = new CockroachContainer(
+			DockerImageName.parse( IMAGE_NAME ).asCompatibleSubstituteFor( "cockroachdb/cockroach:v21.2.4" ) )
```
DavideD, line 30: Could you use constant? So that we don't have to change the image and version in multiple different places?
Something like:
```
String DOCKER_REPOSITORY = "docker.io/";
public static final CockroachContainer cockroachDb = new CockroachContainer(DockerImageName.parse( DOCKER_REPOSITORY + IMAGE_NAME ).asCompatibleSubstituteFor( IMAGE_NAME  ) )
```

Maybe we can move `DOCKER_REPOSITORY` in `TestableDatabase`.

DavideD: Also, here you are keeping the version in the parameter for `asCompatibleSubstituteFor`. You are not replicating it for the other dbs. What's the difference?

DavideD (changes requested):
I've relaunched the build to see if the problem was caused by CI not downloading the image (it happens sometimes).

But I left a couple of additional comments

hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/CockroachDBDatabase.java
blafond: Added docker.io constant and re-worked the image name check for each db except for SQL Server which doesn't work with docker.io prefix.

DavideD:
Merged but I've updated the commit message.
Thanks!

PR #1260, opened 2022-04-11
1.1 jakarta

DavideD:
Draft because I this is going to be a separate branch.
@Sanne, I've noticed [you have changed the configuration in the example](https://github.com/hibernate/hibernate-reactive/commit/16051f66e03ca3c505d5650aade0abc9e5ae7030) so that the properties now have `jakarta.persistence.` as prefix (instead of `javax.persistence.`.

But the class `AvailableSettings` in Hibernate ORM still has all the configuration to `javax.persistence.`. Which one is the correct one? Do I need to update the documentation as well?


Sanne:
> I've noticed [you have changed the configuration in the example](https://github.com/hibernate/hibernate-reactive/commit/16051f66e03ca3c505d5650aade0abc9e5ae7030) so that the properties now have `jakarta.persistence.` as prefix

The purpose of the change was only to test that it would work when using jakarta based configurations - not necessarily to serve as an example.

> Do I need to update the documentation as well?

We don't need to change the configuration as long as the jakarta branch is an experiment.

Eventually we'll release something like "Hibernate Reactive 2.0" based on Jakarta and ORM 6 - at that point we'll need to update the documentation.

DavideD:
I've sent a new PR for a branch called `main-jakarta`: #1262

release/build.gradle
```diff
@@ -31,7 +31,7 @@ description = 'Release module'
 final String documentationDir = rootProject.project( 'documentation' ).buildDir
 
 // Relative path on the static website where the documentation is located
-final String docWebsiteRelativePath = "/reactive/documentation/${projectVersion.family}"
+final String docWebsiteRelativePath = "/reactive/documentation/${projectVersion.family}-jakarta"
```
Sanne, line 34: nice

PR #1267, opened 2022-04-14
Upgrade to ORM 5.6.8 and Vertx 4.2.7

blafond:
syncs with changes in main


DavideD:
Superseeded by https://github.com/hibernate/hibernate-reactive/pull/1272

The session-example still doesn't work, we need to look into it

PR #1269, opened 2022-04-15
Refactoring of the code that starts the containers for testing

DavideD:
@Sanne I've tested this code with a fresh Fedora installation and only using Podman. It seems to work fine for me, except Db2 that starts but doesn't seem to work with Podman. But I think this is a known issue.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/DockerImage.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/DockerImage.java
@@ -1,0 +1,31 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.containers;
+
+import org.testcontainers.utility.DockerImageName;
+
+/**
+ * A utility class with methods to generate {@link DockerImageName} for testcontainers.
+ * <p>
+ * Testcontainers might not work if the image required is available in multiple different registries (for example when
+ * using podman instead of docker).
+ * These methods make sure to pick a registry as default.
+ * </p>
+ */
+public final class DockerImage {
+
+	public static final String DEFAULT_REGISTRY = "docker.io";
+
+	public static DockerImageName imageName(String image, String version) {
+		return imageName( DEFAULT_REGISTRY, image, version );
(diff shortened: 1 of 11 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/DockerImage.java
```diff
@@ -0,0 +1,31 @@
...
+ * A utility class with methods to generate {@link DockerImageName} for testcontainers.
+ * <p>
+ * Testcontainers might not work if the image required is available in multiple different registries (for example when
+ * using podman instead of docker).
+ * These methods make sure to pick a registry as default.
+ * </p>
+ */
+public final class DockerImage {
+
+	public static final String DEFAULT_REGISTRY = "docker.io";
```
Sanne, line 20: :+1:

Sanne:
Just tried it after deleting all my containers and metadata; you can do the same invoking:
```
podman system reset
```
Also might be worth invoking the same as sudo, to delete all images owned by the superuser:
```
sudo podman system reset
```


Then I got:
`com.github.dockerjava.api.exception.BadRequestException: Status 400: {"cause":"short-name resolution enforced but cannot prompt without a TTY","message":"failed to resolve image name: short-name resolution enforced but cannot prompt without a TTY","`
Looks related? It seems to have been caused by CockroachDB

Sanne:
Also: make sure to delete `~/.cache/containers/short-name-aliases.conf` to ensure it didn't record mappings from short names to long names: that will certainly affect reproducibility of these issues.

(look into the file, it's interesting to see how it works)

DavideD:
> Looks related? It seems to have been caused by CockroachDB

I don't know, CockroachDB seems to work fine for me. Even after all the system reset

DavideD:
I don't have a `~/.cache/containers/short-name-aliases.conf`

DavideD:
Interesting, I haven't tested but it seems that this whole issue can be solved buy setting:
```
short-name-mode="permissive"
```
in `/etc/containers/registries.conf` (based on [this article](https://www.redhat.com/sysadmin/container-image-short-names)).
Although, it's not the default because of security concerned.

DavideD:
@Sanne I've added a commit to limit the number of CPUs for Oracle: https://github.com/hibernate/hibernate-reactive/pull/1269/commits/2b8ef4df26d36576f2e0b70af2cf1eb7652549b0

Issue #1015, opened 2021-10-21
AssertionException when using the Statistics service


DavideD:
This error occurs when I run [this test](4897fbd45f5f26463286d48739136a58c462f0e5):

```
    java.lang.AssertionError: Nested calls to partialFlushStart                                 
        at org.hibernate.engine.internal.StatisticalLoggingSessionEventListener.partialFlushStart(StatisticalLoggingSessionEventListener.java:241)                                                                   
        at org.hibernate.engine.internal.SessionEventListenerManagerImpl.partialFlushStart(SessionEventListenerManagerImpl.java:239)
        at org.hibernate.reactive.event.impl.DefaultReactiveAutoFlushEventListener.reactiveOnAutoFlush(DefaultReactiveAutoFlushEventListener.java:39)                                                                
        at org.hibernate.event.service.internal.EventListenerGroupImpl.lambda$fireEventOnEachListener$0(EventListenerGroupImpl.java:133)
        at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1106)                                                                                                             
        at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2235)
        at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:143)                                                                                                                  
        at org.hibernate.event.service.internal.EventListenerGroupImpl.fireEventOnEachListener(EventListenerGroupImpl.java:133)                      
```

This is the testcase for `CachedQueryResultsTest`:
```
	@Test
	public void testQueryPlanCacheHits(TestContext context) {
		test( context, criteriaFindAll()
				.call( CachedQueryResultsTest::criteriaFindAll )
				.call( CachedQueryResultsTest::criteriaFindAll )
				.invoke( () -> context.assertEquals( 2L, statistics().getQueryPlanCacheHitCount() ) )
		);
	}

	private static StatisticsImplementor statistics() {
		return ( (Implementor) getSessionFactory() ).getServiceRegistry().getService(
				StatisticsImplementor.class );
	}

	private static Uni<List<Fruit>> criteriaFindAll() {
		final Mutiny.SessionFactory sf = getMutinySessionFactory();
		return sf.withStatelessSession( s -> s.createQuery( criteriaQuery( sf.getCriteriaBuilder() ) ).getResultList()
		);
	}

[...]

PR #1271, opened 2022-04-18
Added test to verify query plan cache hits

blafond:
addresses [#1015]

After the upgrades to ORM, I revisited this issue.  Verified that the original failing test works (`testQueryPlanCacheHits()`) with `Settings.GENERATE_STATISTICS` to TRUE


DavideD:
Thanks @blafond, it looks good but could you test both use cases? With and without the statistics enabled.

blafond:
sure.  It fails as expected so I'll assert the exception

DavideD:
No... wait... what I mean is that you can add the additional test for when you enable the statistics and check that it works and skip it for when they are disabled. Or, you can test that the result is `0` ( I think) when the statistics are disabled.

No configuration should cause exceptions

blafond:
k. thx

PR #1276, opened 2022-04-23
Ensure that vert.x executor is used for CompletableFutures

encircled:
This fixes the problem, when hibernate randomly throws a "No Vert.x context active" during a query via `Stage.SessionFactory` API. The root cause is that StageSessionFactoryImpl uses `thenApply` and `thenCompose` methods, which does not preserve the executor and may run on main thread, fork join pool, or used-defined one. I believe this is unwanted and it should always run on vertx executor. This did not cause any runtime errors in prior to the change in `org.hibernate.reactive.context.impl.VertxContext`, which started to validate that vertx context is not null, but it is null when run on non-vertx executor.

This issue is hard to reproduce on fast hardware, but happens almost constantly on slower machines, like github actions (some settings of vertx pool perhaps, didn't check). I was able to reproduce it on older laptop as well.

This was asked here for instance - https://stackoverflow.com/questions/69900923/hibernate-reactive-no-vert-x-context-active-in-aws-rds. And I've faced this issue in this lib for reference - https://github.com/encircled/Joiner/runs/6119272633?check_suite_focus=true

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
@@ -82,30 +82,30 @@ public class StageSessionFactoryImpl implements Stage.SessionFactory, Implemento
 	public CompletionStage<Stage.Session> openSession() {
 		SessionCreationOptions options = options();
 		return stage( v -> connection( options.getTenantIdentifier() )
-				.thenCompose( connection -> create( connection, () -> new ReactiveSessionImpl( delegate, options, connection ) ) )
-				.thenApply( s -> new StageSessionImpl(s, this) ) );
+				.thenComposeAsync( connection -> create( connection, () -> new ReactiveSessionImpl( delegate, options, connection ) ), context )
+				.thenApplyAsync( s -> new StageSessionImpl(s, this), context ) );
 	}
 
 	@Override
 	public CompletionStage<Stage.Session> openSession(String tenantId) {
 		return stage( v -> connection( tenantId )
-				.thenCompose( connection -> create( connection, () -> new ReactiveSessionImpl( delegate, options( tenantId ), connection ) ) )
-				.thenApply( s -> new StageSessionImpl(s, this) ) );
+				.thenComposeAsync( connection -> create( connection, () -> new ReactiveSessionImpl( delegate, options( tenantId ), connection ) ), context )
+				.thenApplyAsync( s -> new StageSessionImpl(s, this), context ) );
 	}
 
 	@Override
 	public CompletionStage<Stage.StatelessSession> openStatelessSession() {
 		SessionCreationOptions options = options();
 		return stage( v -> connection( options.getTenantIdentifier() )
-				.thenCompose( connection -> create( connection, () -> new ReactiveStatelessSessionImpl( delegate, options, connection ) ) )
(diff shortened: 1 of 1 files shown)


encircled:
Seems that Mutiny has similar problem, I'm randomly getting a "No Vert.x context active" in a test class with 3 tests, each is doing just one persist query. Sometimes 1 or 2 of them fail.

DavideD:
Thanks,
is there a somewhere a test that I can run to see the error?

encircled:
https://github.com/encircled/hr-1276
On my old laptop it fails every 2nd run approx, screenshot https://github.com/encircled/hr-1276/blob/main/Untitled.png
Windows 10, cpu i5-6300. 

Still can't reproduce it on more powerful machine, setting workerPoolSize&eventLoopPoolSize to 1 did not help

DavideD:
Thanks @encircled, I can see the error with the test you've provided.

encircled:
@DavideD I can try to add unit tests as well, just not sure if it will be reproducible in CI

DavideD:
No, it's fine for now.
I'm just busy this week so I might be slower than usual, but I'm looking into this.

Thanks a lot, I will keep you posted.

Sanne:
Right, we can't delegate back to the Executor at that point as then while it's ensuring to be run on _a_ vert.x thread, it's not necessarily the same one. It really needs to be the same one for the contextual session to be propagated.

I'm trying to understand the issue better; it seems I misunderstood the `CompletionStage` semantics, in particular quoting

 - https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletionStage.html

> Additionally, execution of a new stage's computations may be arranged in any of three ways: default execution, default asynchronous execution (using methods with suffix async that employ the stage's default asynchronous execution facility), or custom (via a supplied [Executor](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/Executor.html)). The execution properties of default and async modes are specified by CompletionStage implementations, not this interface.

And the specific implementation:
 - https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletableFuture.html

> Actions supplied for dependent completions of non-async methods may be performed by the thread that completes the current CompletableFuture, **or by any other caller** of a completion method."

To make things worse I can never reproduce such issues on my machine. Let me ask @encircled , are you able to reproduce the problem even when using the Mutiny based API ?

encircled:
> are you able to reproduce the problem even when using the Mutiny based API

Will test it a bit later. Also, I think it should be possible to reproduce via github actions, will try that as well

@DavideD as for the root problem, I don't think there is a native support for something like 'context' in completable futures and you can't guarantee that the whole flow is executed in the same thread. Any chance to store such data in transaction/session?

encircled:
@Sanne https://github.com/encircled/hr-1276/actions/runs/2275054952 - reproduced on github action

DavideD:
I've added [the same tests with Mutiny](https://github.com/DavideD/hr-1276/blob/main/src/test/java/MutinyVertxContextTest.java) and they seem to work fine: https://github.com/DavideD/hr-1276/runs/6311253853?check_suite_focus=true#step:4:1156

Or at least, it hasn't failed for me yet.

DavideD:
I've adapted the test for Hibernate Reactive: https://github.com/hibernate/hibernate-reactive/pull/1281

Increasing the number of persist in the test will make it fail even on faster machines.
I'm closing this PR since we have a test for HR now.

PR #1278, opened 2022-05-04
Remove duplicated LICENSE files

DavideD:
We have a couple of LICENSE files in the submodules of the project


DavideD:
Fix #1279

PR #1281, opened 2022-05-06
A test for when we run without an initial Vert.x context

DavideD:
Follows https://github.com/hibernate/hibernate-reactive/pull/1276


DavideD:
It seems to fail as expected: https://github.com/hibernate/hibernate-reactive/runs/6325532355?check_suite_focus=true#step:5:1721

Sanne:
Thanks, I had to set an even higher number for `ENTRIES_NUM` but I can now finally reproduce on my machine :+1:

DavideD:
I've updated this PR with the latest changes from @Sanne: #1288 
I still need to clean it up and check why it fails for Oracle but I will do it tomorrow.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -48,9 +50,12 @@ public <T> void put(Key<T> key, T instance) {
 	public <T> T get(Key<T> key) {
 		final io.vertx.core.Context context = Vertx.currentContext();
 		if ( context != null ) {
-			return context.getLocal( key );
+			T local = context.getLocal( key );
+			LOG.tracef( "Getting value %2$s from context for key %1$s", key, local  );
+			return local;
 		}
 		else {
+			LOG.tracef( "Getting null from context for key %s", key );
```
Sanne, line 58: This message is a bit confusing as it states we "got" null, but it's the context that's null.
What about "Context get operation skipped for key %s as there is no current context"

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
```diff
@@ -59,22 +64,29 @@ public <T> T get(Key<T> key) {
 	public void remove(Key<?> key) {
 		final io.vertx.core.Context context = Vertx.currentContext();
 		if ( context != null ) {
-			context.removeLocal( key );
+			boolean removed = context.removeLocal( key );
+			LOG.tracef( "Key %s removed from context: %s", key, removed );
```
Sanne, line 68: are you sure about using "%s" for a boolean primitive?

Sanne, line 68: are you sure about using "%s" for a boolean primitive?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/OneOffDelegatingExecutor.java
```diff
@@ -13,14 +13,14 @@
  * via e.g. CompletableFuture.supplyAsync or .thenApplyAsync(task, Executor)
  * one needs to be careful that subsequent stages aren't guaranteed to be run
  * on the same executor, as execution is scheduled immediately, and
- * there's a change for the task to be completed before the chain
+ * there's a chance for the task to be completed before the chain
```
Sanne, line 16: thanks :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -149,13 +150,17 @@ private ReactiveConnection proxyConnection(String tenantId) {
...
 			LOG.debug( "Reusing existing open Stage.Session which was found in the current Vert.x context" );
 			return work.apply( current );
 		}
 		else {
 			LOG.debug( "No existing open Stage.Session was found in the current Vert.x context: opening a new instance" );
-			return withSession( openSession(), work, contextKeyForSession );
+			OneOffDelegatingExecutor oneOffExecutor = new OneOffDelegatingExecutor( context );
+			final CompletionStage<T> withSessionStage = voidFuture()
+					.thenComposeAsync( v -> withSession( openSession(), work, contextKeyForSession ), oneOffExecutor );
+			oneOffExecutor.runHeldTasks();
```
Sanne: :+1:

Sanne: But is it necessary now after the other changes? Feels like you might be able to remove the `OneOffDelegatingExecutor` class, since `openSession()` isn't using _thenComposeAsync_, so this completionstages series is fully defined when it's being passed to the Executor.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -171,6 +179,7 @@ public <T> CompletionStage<T> withSession(String tenantId, Function<Stage.Sessio
 		else {
 			LOG.debugf( "No existing open Stage.Session was found in the current Vert.x context for current tenant '%s': opening a new instance", tenantId );
 			return withSession( openSession( tenantId ), work, key );
+			return executeLazily( v -> withSession( openSession( tenantId ), work, key ) );
```
Sanne: Like my previous comment, I suspect these various `executeLazily` are not acually necessary. Same for the other instances.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/stage/impl/StageSessionFactoryImpl.java
```diff
@@ -201,10 +192,17 @@ public <T> CompletionStage<T> withStatelessSession(String tenantId, Function<Sta
...
-			return withSession( openStatelessSession( tenantId), work, key );
+			return oneOffExecution( v -> withSession( openStatelessSession( tenantId ), work, contextKeyForStatelessSession ) );
 		}
 	}
 
+	private <T> CompletionStage<T> oneOffExecution(Function<Void, CompletionStage<T>> fun) {
+		OneOffDelegatingExecutor oneOffExecutor = new OneOffDelegatingExecutor( context );
+		final CompletionStage<T> withSessionStage = voidFuture().thenComposeAsync( fun, oneOffExecutor );
+		oneOffExecutor.runHeldTasks();
+		return withSessionStage;
```
Sanne, line 203: Interesting, but if this is the final pattern then the `OneOffDelegatingExecutor` should not be necessary?
The function is completely defined at this point, so there are no additional stages that still need the Session as downstream events: you could run it with `thenComposeAsync` directly.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java
DavideD: Yes, you can use it with any type

DavideD:
Superseeded by https://github.com/hibernate/hibernate-reactive/pull/1292

PR #1284, opened 2022-05-11
Configuration option 'hibernate.order_updates' is not effective

Sanne:
We'll need to add a test, but leaving it here for now as I need testing in other projects.


DavideD:
Merged https://github.com/hibernate/hibernate-reactive/pull/1284

I've just added the issue code to the commit message

PR #1288, opened 2022-05-12
WIP Async fixes

Sanne:
This is another component to work towards a fix for #1276

It should be enough to fix the problem, BUT all users would be required to run every operation on the vertx thread. This is too unconvenient, so we'll also work on a follow up with the `OneOffDelegatingExecutor` (introduced here) to at least keep running a selected set of tasks in the deferred executor.

cc/ @DavideD


Sanne:
Sketch of how to use this:
 - https://gist.github.com/Sanne/42d9a79be578306ffd64e11af252dad6

DavideD:
CLosing this beacause I've already integrated in https://github.com/hibernate/hibernate-reactive/pull/1281

PR #1290, opened 2022-05-13
Remove some lambda expressions

geoand:
These are causing NoSuchMethodError on application startup
(this can be verified easily either via the debugger, or from JFR events - see
https://bugs.openjdk.java.net/browse/JDK-8161588 for more details
about why the error occurs) which can have a small impact on startup performance

N.B. I only replaced the code for the instances
that were causing the issue at the startup of a
Quarkus application that uses Hibernate Reactive


DavideD:
Thanks

geoand:
YW

PR #1292, opened 2022-05-13
Fix for no vertx context active message error

### LINKED ISSUE #1289: Ensure that we are using the right thread when HR is used without an initial context
See https://github.com/hibernate/hibernate-reactive/pull/1276

DavideD:
Fixes #1289


Sanne (approved):
Awesome, thanks

PR #1308, opened 2022-05-24
Update DB2 container image

Sanne:
also for consistency with other projects


DavideD:
The changes to the podman.md file are not correct thought.
Those instructions are for starting podman/docker containers without  using testcontainers. in that case one doesn't need to add the options `-Pdocker`

DavideD:
I've inlcuded the changes here: https://github.com/hibernate/hibernate-reactive/pull/1307

PR #1311, opened 2022-05-24
Upgrade Vert.x to 4.3.0

### LINKED ISSUE #1310: Upgrade to Vert.x 4.3.0

DavideD:
Fix #1310


DavideD:
@Sanne Any reason not to upgrade this?

Sanne:
Quarkus is still using Vert.x 4.2.7 - we can upgrade here but let's be ready to possibly need to release maintenance and bugfix versions for versions on Vert.x 4.2.7.

In other words, we probably want to bump the minor here and create a branch - but we can do that lazily.

PR #1316, opened 2022-05-25
Build using latest Vert.x snapshot plus Oracle

### LINKED ISSUE #1245: Re-enable build on CI with Oracle and the Vert.x snapshot
See https://github.com/hibernate/hibernate-reactive/pull/1244

Something has changed and makes that build fail with the following error:
```
Gradle Test Executor 6 > org.hibernate.reactive.BatchingConnectionTest > reactivePersist FAILED
    java.util.concurrent.CompletionException: io.vertx.oracleclient.OracleException: Error : 1453, Position : 0, Sql = SET TRANSACTION ISOLATION LEVEL READ COMMITTED, OriginalSql = SET TRANSACTION ISOLATION LEVEL READ COMMITTED, Error Msg = ORA-01453: SET TRANSACTION must be first statement of transaction
```

I think it's related to lock entities and I'm trying to create a reproducer for the Vert.x SQL client

DavideD:
Fix #1245

### DIFF
--- a/.github/workflows/tracking-vertx-4.build.yml
+++ b/.github/workflows/tracking-vertx-4.build.yml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         example: [ 'session-example', 'native-sql-example' ]
         db: [ 'MySQL', 'PostgreSQL' ]
-        vertx-version: [ '[4.2,4.3)' ]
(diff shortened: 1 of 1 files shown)


DavideD:
Seems to work fine, I will merge it and see if it fails in the future

PR #1327, opened 2022-06-09
Refactoring of a couple of tests

DavideD:


DavideD:
@blafond Wait a second to merge this one, I think we should make it work for all dbs

DavideD:
Ok, nevermind. Let's test this only on Postgres for now. Maybe in the future we will think of a more portable way to test this issue

PR #1337, opened 2022-06-13
Remove warnings

DavideD:
A collection of changes to remove warnings from IDEA (and cleaning the code in the process).
They should be safe but better if somebody else double check


gavinking:
LGTM, assuming you did what those comments say, and tested it on Quarkus....

DavideD:
Good point. Completely missed it. I'm going to revert that change

PR #1339, opened 2022-06-14
Validate identity generated id for CockroachDB

### LINKED ISSUE #1338: Throw better exception when using CockroachDB with Identity and the type is not long
When using identity generation with CockroachDB, users should use Long.
Identity generation of type Integer or Short, **usually** fail because the id generated by Cockroach are too big.

From time to time, it might happen that the id generated for an Integer is valid and this might cause random issues.

DavideD:
Fix #1338 

The error message now looks like:
```
org.hibernate.HibernateException: HR000073: java.lang.Short is an invalid identity type when using CockroachDB (entity org.hibernate.reactive.IdentityGeneratorTypeForCockroachDBTest$ShortTypeEntity) - CockroachDB might generates identifiers that are too big and won't always fit in a java.lang.Short. java.lang.Long is valid replacement
```

I don't think this was a critical issue before because we threw exceptions based on the size of the identifiers.
But this should give users more info about what's going on.

It would be nicer to validate this at startup


DavideD:
Thanks

Issue #1365, opened 2022-07-13
Add test case for ReactiveSession#clear


DavideD:
See #1362

Not critical, but at the moment we don't have an easy way to access the implementation and it makes writing unit tests for this class (and ReactiveActionQueue) hard.

meepown:
We found this bug because we also use the **graphql lib**. But it seems not work on the CDI scope, it's Singleton.  See [smallrye-graphql#1295](https://github.com/smallrye/smallrye-graphql/issues/1295).  So we have to add session.clear() when one request end and the bug found

meepown:
I can provide our useage.   we just add an interceptor, like:
```

public class ClearSessionInterceptor{

  @Inject Mutiny.Session session;

  @Override
  @AroundInvoke
  public Object intercept(InvocationContext context) throws Exception {
    if (null != session && session.isOpen()) {
      session.clear();
    }

    return context.proceed();
  }
}
```

meepown:
the graphql endpoint just like:

```
import org.eclipse.microprofile.graphql.GraphQLApi;
import org.eclipse.microprofile.graphql.Query;

@GraphQLApi
public class TestGraphQL {
    @Query
    @ClearSession
    public Uni<String> helloworld() {
         return Uni.createFrom().item("hello");
    }
}
```

meepown:
> OK, but when did you notice that the queue wasn't being cleared?

I notice that when  duplicate key error occurs with update acation.   here is the case: 
Pig.java:
```
import lombok.*;

import javax.persistence.*;

/**
 * @author yue.tang @Description Pig
 * @date 2022/7/14
 */
@Entity
@AllArgsConstructor
@NoArgsConstructor
@Data
@EqualsAndHashCode(callSuper = false)
@ToString
@Table(indexes = {@Index(columnList = "name", unique = true)})
public class Pig {
  @Id
  @GeneratedValue(strategy = GenerationType.IDENTITY)
  Long id;

  String name;
}
```

TestGraphql.java:
```
import io.smallrye.mutiny.Uni;
import org.eclipse.microprofile.graphql.GraphQLApi;
import org.eclipse.microprofile.graphql.Query;
import org.hibernate.reactive.mutiny.Mutiny;

import javax.inject.Inject;

/**
 * @author yue.tang @Description TestGraphql
 * @date 2022/7/14
 */
@GraphQLApi
public class TestGraphql {

  @Inject Mutiny.SessionFactory sessionFactory;

[...]

Issue #1362, opened 2022-07-13
Add ReactiveActionQueue.clear to ReactiveSessionImpl


meepown:
SessionImpl.clear() can clear activeQueue but ReactiveSessionImpl can't clear its ReactiveActionQueue. So once update/add event occurs error, the ReactiveActionQueue always keep these event. Whould ReactiveSessionImpl  override clear() method and clear its ReactiveActionQueue?

DavideD:
Merged, thanks!

I've remove some extra spaces and added the issue number to the commit.

Out of curiosity, how did you find this? Was there a use case where it was causing issues?

DavideD:
I'm asking because [it would be nice to have a test](https://github.com/hibernate/hibernate-reactive/issues/1365) for this

gavinking:
@DavideD I guess you could reproduce the issue using:

```java
session.persist(entity).invoke(()->session.clear()).chain(()->session.find(Entity.class, id))
```

DavideD:
I will try

DavideD:
> session.persist(entity).invoke(()->session.clear()).chain(()->session.find(Entity.class, id))

I've tried but it works even without the fix. I think there is a check somewhere related to the original session that makes everything works even if we don't  clear the reactive queue.

gavinking:
Wait, my code was wrong. You would need to add an explicit `flush()` in there!

DavideD:
Note that I've reverted the fix on that branch because I want to see the test fail

blafond:
@DavideD Running your test above and I'm not seeing an error

DavideD:
That particular test working. That's why I've asked for example from OP.

DavideD:
> ... I should have gotten an error if clear() wasn't working, right?

Not necessarly, it's possible that in that particular case everything still works for some reason.

DavideD:
We could add the test in the code base anyway (I don't think we have one  like this already).

DavideD:
This is the error I see when I add the test without the fix: https://github.com/hibernate/hibernate-reactive/runs/7453020650?check_suite_focus=true#step:6:1829

Note that only MutintySessionTest has failed.

PR #1366, opened 2022-07-20
Add test cases for ReactiveSession#clear

blafond:
Addresses session.clear() for #1365 and related to #1362

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
@@ -57,6 +57,25 @@ public class MutinySessionTest extends BaseReactiveTest {
 	}
 
 	@Test
+	public void sessionClear(TestContext context) {
+		final GuineaPig guineaPig = new GuineaPig( 81, "Perry" );
+		test(
+				context,
+				getMutinySessionFactory().withSession( session -> session
+						.persist( guineaPig )
+						.invoke( session::clear )
+						// If the previous clear doesn't work, this will cause a duplicated entity exception
+						.chain( () -> session.persist( guineaPig ) )
+						.call( session::flush )
+						.chain( () -> session.createQuery( "FROM GuineaPig", GuineaPig.class )
+								// By not using .find() we check that there is only one entity in the db with getSingleResult()
+								.getSingleResult() )
+						.invoke( result -> assertThatPigsAreEqual( context, guineaPig, result ) )
+				)
+		);
+	}
+
+	@Test
 	public void reactiveWithTransactionStatelessSession(TestContext context) {
 		final GuineaPig guineaPig = new GuineaPig( 61, "Mr. Peanutbutter" );
 		test( context, getMutinySessionFactory()
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
@@ -77,6 +77,25 @@ public class ReactiveSessionTest extends BaseReactiveTest {
 	}
 
 	@Test
+	public void sessionClear(TestContext context) {
+		final GuineaPig guineaPig = new GuineaPig( 81, "Perry" );
+		test(
+				context,
+				getSessionFactory().withSession( session -> session
+						.persist( guineaPig )
+						.thenAccept( v -> session.clear() )
+						// If the previous clear doesn't work, this will cause a duplicated entity exception
+						.thenCompose( v -> session.persist( guineaPig ) )
+						.thenCompose( v -> session.flush() )
+						.thenCompose( v -> session.createQuery( "FROM GuineaPig", GuineaPig.class )
+								// By not using .find() we check that there is only one entity in the db with getSingleResult()
+								.getSingleResult() )
+						.thenAccept( result -> assertThatPigsAreEqual( context, guineaPig, result ) )
+				)
+		);
+	}
+
+	@Test
 	public void reactiveWithTransactionSession(TestContext context) {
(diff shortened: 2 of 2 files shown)


DavideD:
Mmh... there something not quite right with this PR.
If I remove the fix, only `MutinySessionTest` fails. It also fails with an error that's not what the issue is about:
```
    javax.persistence.PersistenceException: org.hibernate.HibernateException: org.hibernate.AssertionFailure: possible non-threadsafe access to session                                                                                                                                                                                                                                                                                     
```

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -56,6 +56,32 @@ private static String nameFromResult(List<Object> rowSet) {
 		}
 	}
 
+	@Test
+	public void testClearSessionAndChangeName(TestContext context) {
+		final GuineaPig pig1 = new GuineaPig( 81, "pig1" );
+		final GuineaPig pig2 = new GuineaPig( 82, "pig2" );
```
DavideD, line 62: Why are you creating two entities. I think the original issue was about:

1. session.persist pig1
2. session.clear
3. session.persist pig1
4. session.flush

This steps, without the fix, would cause a Persistence exception because it looks like we are trying to insert the same entity twice. But now you are trying to do something different that I don't follow.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -76,6 +76,31 @@ public void reactiveFind(TestContext context) {
...
+	@Test
+	public void testClearSessionAndChangeName(TestContext context) {
+		final GuineaPig pig1 = new GuineaPig( 81, "pig1" );
+		final GuineaPig pig2 = new GuineaPig( 82, "pig2" );
+
+		test( context,
+				getSessionFactory().withSession( session -> session
+						.persist( pig1, pig2 )
+						.thenCompose( v -> getSessionFactory()
+								.withSession( s2 -> s2.clear().flush() ) )
```
DavideD, line 88: Why don't you `session` here?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ReactiveSessionTest.java
```diff
@@ -76,6 +76,31 @@ public void reactiveFind(TestContext context) {
...
+						.thenCompose( v -> {
+							// change name after clearing session and persist
+							pig1.setName( "pig2" );
+							return session.persist( pig1 );
+						} )
+						.thenCompose( v -> session.find( GuineaPig.class, pig1.getId() ) )
+						.thenAccept( result -> {
+							context.assertNotNull( result );
+							context.assertEquals( "pig2", result.name );
+							context.assertEquals( pig1.getId(), result.getId() );
```
DavideD, line 98: `GuineaPig` already implements equals, so you can `assertEquals(pig1, result)`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -56,6 +56,32 @@ private static String nameFromResult(List<Object> rowSet) {
...
+							// change name after clearing session and persist
+							pig1.setName( "pig2" );
+							return session.persist( pig1 );
+						} )
+						.call( session::flush )
+						.chain( () -> session.find( GuineaPig.class, pig1.getId() ) )
+						.invoke( result -> {
+							context.assertNotNull( result );
+							context.assertEquals( "pig2", result.name );
+							context.assertEquals( pig1.getId(), result.getId() );
```
DavideD, line 79: `GuineaPig` already implements equals, so you can `assertEquals(pig1, result)`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
```diff
@@ -56,6 +56,32 @@ private static String nameFromResult(List<Object> rowSet) {
...
+	@Test
+	public void testClearSessionAndChangeName(TestContext context) {
+		final GuineaPig pig1 = new GuineaPig( 81, "pig1" );
+		final GuineaPig pig2 = new GuineaPig( 82, "pig2" );
+
+		test( context,
+				getMutinySessionFactory().withSession( session -> session
+						.persistAll( pig1, pig2 )
+						.invoke( session::clear )
+						.invoke( session::flush )
```
DavideD, line 68: Not sure if it's intentional, but this is different from what the user was doing. The user first flush and then clear.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
DavideD: I see that the original user is actually doing an update of the entity. But can we simplify it and use a single entity instead?

blafond: yes, probably

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
DavideD: I've just noticed that flush should be  `.call(session::flush)`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutinySessionTest.java
blafond: re-worked the tests

DavideD:
Merged, thanks.

It works as expected now.

Issue #1287, opened 2022-05-12
Create a test for 'hibernate.order_updates'  option


DavideD:
We already have a fix we tested manually. We need to add JUnit test.
See #1284

blafond:
@DavideD I pushed a draft test to my repo. I don't think it's set up correctly yet so I have some questions

blafond:
@DavideD @Sanne this issue was logged stating the config property `hibernate.order_updates` was not effective.  I'm having trouble defining an applicable test.

Is there an example somewhere you recommend that would create the persist/update scenario that would trigger that property?

DavideD:
The one we discussed yesterday doesn't work?

blafond:
The test passes, but I can't reproduce any scenario where removing the 1 line fix creates an error of some sort.

Sanne:
It's indeed tricky to catch a failure; the reason I noticed that it wasn't doing the right thing is that when having multiple parallel requests, under load, is that occasionally there would be spikes in latency because of contention on the data in the database.
In extreme cases (but still quite likely under load) a deadlock would manifest.

These deadlocks would happen because we're locking N rows at random, in an unspecified order. But having set `order_updates` I would have expected all operations to be completed by locking the N rows in the same, consistent order - in which case deadlocks wouldn't happen.

N.B. I'm not suggesting to include a stress test or a load test - that would be overkill. Just verifying that update SQL statements are emitted in the expected order would be enough.

blafond:
@DavideD I pushed a couple of new commits and rewrote the test.  Wasn't a simple change to capture logged queries with actual parameter values. I had to tweak SqlClientConnection to trigger the process of generating additional SQL statements for the tracker to serve up to the tests.

I couldn't find a way around not moving `SqlStatementTracker` into `pool/impl` package though.  Wasn't my first choice.

And I still can't make the test fail by removing your one-line [commit](https://github.com/hibernate/hibernate-reactive/commit/d62630f68ce0170b377253bd2c0417462abd79e8).

DavideD:
I've created a draft for a test case here: https://github.com/DavideD/hibernate-reactive/commit/ebb7475ff4150854efddb4a6c7317f715a7f207a?diff=unified

You can see in the log that when the `hibernate.order_updates` property is false the SQL queries logged are:
```
[Hibernate] update Category set name=$1 where id=$2
[Hibernate] update Product set name=$1, price=$2 where id=$3
[...]

PR #1370, opened 2022-08-03
Added test for 'hibernate.order_updates' option

blafond:
Addresses #1287

- Checks order of logged sql statements for `order_updates` = true/false
- Checks that logged sql statements are collapsed when `order_updates` = true & `hibernate.jdbc.batch_size` is set

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
@@ -1,0 +1,442 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.provider.Settings;
+import org.hibernate.reactive.testing.DatabaseSelectionRule;
+import org.hibernate.reactive.testing.SqlStatementTracker;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import io.smallrye.mutiny.Uni;
+import io.vertx.ext.unit.TestContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.DB2;
+
+public abstract class UpdateOrderingTestBase extends BaseReactiveTest {
+	public enum QUERY_TYPE {
+		INSERT,
+		UPDATE
+	}
+
+	public static class UpdateOrderingTrueTest extends UpdateOrderingTestBase {
+
+		@Override
+		public String isUpdateOrdering() {
+			return "true";
+		}
+
+		@Override
+		public String isInsertOrdering() {
+			return "false";
+		}
+
+		@Override
+		public String getBatchSize() {
+			return "0";
+		}
+
+		@Override
+		public QUERY_TYPE queryType() {
+			return QUERY_TYPE.UPDATE;
+		}
+
+		@Override
+		public void assertLoggedQueries() {
+			assertThat( sqlTracker.getLoggedQueries() ).hasSize( 6 );
+			assertThat( sqlTracker.getLoggedQueries().get( 0 ) ).startsWith( "update Category set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 1 ) ).startsWith( "update Category set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 2 ) ).startsWith( "update Category set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 3 ) ).startsWith( "update Product set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 4 ) ).startsWith( "update Product set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 5 ) ).startsWith( "update Product set name=" );
+			sqlTracker.clear();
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+import io.smallrye.mutiny.Uni;
+import io.vertx.ext.unit.TestContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.POSTGRESQL;
+
+public abstract class UpdateOrderingTestBase extends BaseReactiveTest {
+
+	@Rule
+	public DatabaseSelectionRule rule = DatabaseSelectionRule.runOnlyFor( POSTGRESQL );
```
DavideD, line 34: I think this test will work for all databases, no need to limit it to Pg

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+		public void assertLoggedQueries() {
+			assertThat( sqlTracker.getLoggedQueries() ).hasSize( 30 );
+			assertThat( sqlTracker.getLoggedQueries().get( 0 ) ).startsWith( "update Category set name=" );
+			assertThat( sqlTracker.getLoggedQueries().get( 1 ) ).startsWith( "update Product set name=" );
+			sqlTracker.clear();
+		}
+	}
+
+	//	Test that when we set the Batch size only the expected number of queries run.
+	//	For example, setting hibernate.jdbc.batch_size to 50 will only log two update queries in this case (when order_inserts is true)
```
DavideD, line 83: I think you mean `when order_updates is true`, not order_inserts

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+												.invoke( category -> category.setName( "New " + category.getName() ) )
+												.chain( () -> session.find( Product.class, id ) )
+												.invoke( product -> {
+													product.setName( "New " + product.getName() );
+													product.setPrice( 2 * product.getPrice() );
+												} );
+									}
+									return loop;
+								} )
+								.invoke( () -> assertLoggedQueries() ) )
```
DavideD, line 184: you can change this to `.invoke(this::assertLoggedQueries)` (I think)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+		}
+
+		@Override
+		public String getBatchSize() {
+			return "0";
+		}
+
+		@Override
+		public void assertLoggedQueries() {
+			assertThat( sqlTracker.getLoggedQueries() ).hasSize( 30 );
```
DavideD, line 50: I think we can reduce the number of queries. No need to run 30 of them.
We can test with just 6 queries for example.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+	 * @return value of property {@link Settings#STATEMENT_BATCH_SIZE}
+	 */
+	public abstract String getBatchSize();
+
+	public SqlStatementTracker sqlTracker;
+
+	@Override
+	protected Configuration constructConfiguration() {
+		Configuration configuration = super.constructConfiguration();
+		configuration.setProperty( Settings.ORDER_INSERTS, "false" );
```
DavideD, line 125: Could you also test the inserts, please? We do the inserts in the `populatedb`,  so the test only needs to check the content of `sqlTracker`.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,307 @@
...
+		configuration.setProperty( Settings.ORDER_UPDATES, Boolean.toString( isUpdateOrdering() ) );
```
DavideD, line 126: Why don't you use a String instead of a boolean? You wouldn't have to convert the value.

DavideD (changes requested):
Thanks! Looks great overall.
I just have a few notes.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
blafond: done

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
blafond: done

hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpdateOrderingTestBase.java
```diff
@@ -0,0 +1,442 @@
...
+	private static boolean onlyUpdates(String s) {
+		return s.toLowerCase().startsWith( "update" );
+	}
+
+	private static boolean onlyInserts(String s) {
+		return s.toLowerCase().startsWith( "insert" );
+	}
+
+	private static boolean nothing(String s) {
+		return s.toLowerCase().startsWith( "nothing" );
```
DavideD, line 272: You could just return false

blafond: that's true :)

blafond: done

blafond:
Looks like the tests only succeed with PG, Cockroach & Oracle

DavideD:
See https://github.com/hibernate/hibernate-reactive/pull/1374

PR #1377, opened 2022-08-17
Upgrade to Vert.x 4.3.3

Sanne:
Experiment


Sanne:
Looks like straightforward. Should we upgrade @DavideD ?

PR #1387, opened 2022-09-15
Lazy field

### LINKED ISSUE #1386: ClassCastException when more than one field is lazy and bytecode enhancement is enabled
If we add a new field to [Book](https://github.com/hibernate/hibernate-reactive/blob/main/examples/session-example/src/main/java/org/hibernate/reactive/example/session/Book.java):
```
@Basic(fetch = LAZY)
public byte[] coverImage;
```

Hibernate Reactive throws `ClassCastException` because of a bug in the [loop](https://github.com/hibernate/hibernate-reactive/blob/42e0b2d949377e8d92ef3941cb6926a581dd5743/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveAbstractEntityPersister.java#L1276): it will always only load correctly one lazy field.

DavideD:
Fix #1386 
I've followed [the approach used by Hibernate ORM](https://github.com/hibernate/hibernate-orm/blob/1350c713a3c271d869d05fdf9e28fe2e57a4e25f/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java#L1267).

If I'm not mistaken, we are loading all the lazy properties as soon as we load the first one. Is this the correct behaviour?
Shouldn't we just load the lazy field required by the user?

### DIFF
--- a/examples/session-example/src/main/java/org/hibernate/reactive/example/session/Book.java
+++ b/examples/session-example/src/main/java/org/hibernate/reactive/example/session/Book.java
@@ -35,6 +35,9 @@ class Book {
 	@NotNull @Past
 	private LocalDate published;
 
+	@Basic(fetch = LAZY)
+	public byte[] coverImage;
+
 	@NotNull
 	@ManyToOne(fetch = LAZY)
 	private Author author;
@@ -44,6 +47,7 @@ class Book {
 		this.isbn = isbn;
 		this.author = author;
 		this.published = published;
+		this.coverImage = ("Cover image for '" + title + "'").getBytes();
 	}
 
 	Book() {}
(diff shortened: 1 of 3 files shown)


DavideD:
I also suspect there is a but in ORM, but I need to write a test first

gavinking:
The properties come in named "fetch groups". And all lady properties belonging to the same group should be fetched together.

DavideD:
> I also suspect there is a bug in ORM, but I need to write a test first

@blafond will check this

DavideD:
It seems that something has changed in the Vert.x snapshots and the workflows testing it is failing. I will check tomorrow what's going on. I've disabled that specific workflow for now.

DavideD:
> The properties come in named "fetch groups". And all lazy properties belonging to the same group should be fetched together.

OK, thanks.

gavinking (approved):
Looks reasonable to me.

DavideD:
Merged, thanks

PR #1405, opened 2022-10-18
Upgrade db versions for testing

### LINKED ISSUE #1406: Updgrade db versions

DavideD:
Fix #1406 

~~This is a draft because I'm waiting for the latest mysql connector to be available on Maven central~~
Nevermind, the problem was that the GAV for the connector have changed: https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-31.html

@Sanne Could you have a quick look and let me know if it's better if we test with different versions, please? Thanks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
           - 3306:3306
       postgres:
         # Docker Hub image
-        image: postgres:14.2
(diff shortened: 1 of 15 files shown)


DavideD:
I've merged this because it touches several files and I would prefer to avoid conflicts, but let me know if there's something to revert.

PR #1411, opened 2022-10-19
Upgrade Cockroach DB to 22.1.9

### LINKED ISSUE #1407: Build fails with CockroachDB 22.1.9
Here's an extract of [one of the stacktraces](https://github.com/hibernate/hibernate-reactive/actions/runs/3272511593/jobs/5383609666#step:6:2716):
```
Gradle Test Executor 7 > org.hibernate.reactive.schema.SchemaUpdateCockroachDBTestBase$IndividuallySchemaUpdateCockroachTestBase > testValidationSucceed PASSED                                                      
                                                                                                                                                                                                                     
Gradle Test Executor 7 > org.hibernate.reactive.schema.SchemaUpdateCockroachDBTestBase$IndividuallySchemaUpdateCockroachTestBase > testUpdate FAILED                                                                 
    java.lang.AssertionError: Not equals : CREATE INDEX i_asimple_avalue_astringvalue ON postgres.public.asimple USING btree (avalue ASC, astringvalue DESC) != CREATE UNIQUE INDEX asimple_pkey ON postgres.public.$
simple USING btree (id ASC)                                                      
```

A branch with Cockroach already updated is here: https://github.com/DavideD/hibernate-reactive/tree/update-cockroach

blafond:
Upgrade Cockroach DB version to 22.1.9 and fix update schema test error

fixes #1407

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateCockroachDBTestBase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateCockroachDBTestBase.java
@@ -31,6 +31,7 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import io.vertx.ext.unit.TestContext;
+import org.assertj.core.api.Assertions;
 
 import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.COCKROACHDB;
 import static org.hibernate.tool.schema.JdbcMetadaAccessStrategy.GROUPED;
@@ -103,6 +104,8 @@ public abstract class SchemaUpdateCockroachDBTestBase extends BaseReactiveTest {
 
 	@Test
 	public void testUpdate(TestContext context) {
+		// this test checks that the correct indexes and foreign constraints have been created
+		// by looking at the content of the pg_catalog schema using the following native queries.
 		final String indexDefinitionQuery =
 				"select indexdef from pg_indexes where schemaname = 'public' and tablename = ? order by indexname";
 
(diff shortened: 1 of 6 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateCockroachDBTestBase.java
```diff
@@ -155,21 +155,17 @@ public void testUpdate(TestContext context) {
...
-															"CREATE UNIQUE INDEX \"primary\" ON postgres.public.asimple USING btree (id ASC)",
-															list.get( 2 )
+													context.assertTrue( list.contains(
+															"CREATE UNIQUE INDEX asimple_pkey ON postgres.public.asimple USING btree (id ASC)")
 													);
-													context.assertEquals(
-															"CREATE UNIQUE INDEX u_asimple_astringvalue ON postgres.public.asimple USING btree (astringvalue ASC)",
-															list.get( 3 )
+													context.assertTrue( list.contains(
+															"CREATE UNIQUE INDEX u_asimple_astringvalue ON postgres.public.asimple USING btree (astringvalue ASC)")
```
DavideD, line 168: The problem with using `assertTrue` is that in case of error you we won't see the difference in the log about what we were expecting to find and what we have found.

It's not a big deal but I would replace this with:
```
.thenAccept( list -> assertThat( list ).containsExactlyInAnyOrder(
		"CREATE INDEX i_asimple_avalue_astringvalue ON postgres.public.asimple USING btree (avalue ASC, astringvalue DESC)",
		"CREATE INDEX i_asimple_avalue_data ON postgres.public.asimple USING btree (avalue DESC, data ASC)",
		"CREATE UNIQUE INDEX asimple_pkey ON postgres.public.asimple USING btree (id ASC)",
		"CREATE UNIQUE INDEX u_asimple_astringvalue ON postgres.public.asimple USING btree (astringvalue ASC)"
) )
```

DavideD (changes requested):
I think it's better if you squash the commits together for this PR.
If the changes are not committed together the build will fail during a `git bisect`:
if you update the version first, the build will fail because the queries don't match the expectation. The same is true if you update the expected queries first.

Thanks!

DavideD:
@blafond Maybe you can also add a comment on the test explaining that we are checking that the indexes and foreign constraints haven been created by looking at the content of the `pg_catalog` schema with a native query.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateCockroachDBTestBase.java
blafond: Hadn't used that contains method before.  I'll make the changes.

DavideD:
Merged!
Thanks

PR #1430, opened 2022-11-29
[#1425] Fix URL parsing error

### LINKED ISSUE #1425: Wrong support for URI
Hi Hibernate ! 

I worked with Hibernate Reactive to introduce it in my client's spring boot environment. But the OPS team in my client use an api to generate URL DB. Theses url contains _  (underscore). Using _ will never be a problem with "standard" hibernate, I use these url today without any issue but you choose to use URI instead of URL and URI doesn't support underscore.

The problem is  [here : DefaultSqlClientPoolConfiguration](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L93) 

Example value (anything with _) : jdbc:postgresql://my_db:5432.com/my_schema

URI doesn't throw any error but the value Host & port stay null and authority is fill instead. but DefaultSqlClientPoolConfiguration (L 111) : 

[link](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L109)

`
 String host = scheme.equals("oracle")
                ? oracleHost(uri)
                : uri.getHost();
`

[link](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java#L113)
`
        int port = scheme.equals("oracle")
                ? oraclePort(uri)
                : uri.getPort();
`

Try to read only getHost() and getPort()

This problem doesn't occurs with URL, but URL doesn't support all scheme like URI. 
But URI must support _ (https://www.rfc-editor.org/rfc/rfc3986#section-2.3), I'll try to replace URI with %5F but still broken. 

Replace by ipv4 adresse == no problem

Problem follow here in OpenJDK : https://bugs.openjdk.org/browse/JDK-8291591

But today is a regression comparing to Hibernate  "standard". And Without dns name I can't go to production with Hibernate Reactive today :'( 

For more in formation : 

Hibernate version : 
org.hibernate.reactive:hibernate-reactive-core:1.1.9.Final

blafond:
Relaxed host value character constraints and added test to verify.

fixes #1425


blafond:
There are some differences in the options for some DB's.  Little more work to do.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
```diff
@@ -305,6 +305,23 @@ private String oraclePath(URI uri) {
 		return string.substring( start, end );
 	}
 
+	// if the host contains invalid characters
+	// {@link java.net.URI} parseAuthority(int start, int n) will end up setting host to NULL
```
DavideD, line 309: There's no point to use `{@link ...`} if it's not a javadoc.
Maybe in this case a javadoc would be better even if it's a private method.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
```diff
@@ -305,6 +305,23 @@ private String oraclePath(URI uri) {
...
+	//  when a {@link java.net.URISyntaxException} is thrown
+	//  this method will reparse and set the host value
+	private String findHost( URI uri ) {
+		if( uri.getHost() == null ) {
```
DavideD, line 313: Missing a white space after if

hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java
```diff
@@ -108,11 +108,31 @@ public void testDatabaseAsProperty() {
...
 	 * and assert that's correct.
 	 */
 	private void assertOptions(String url, String expectedDbName, Map<String, String> parameters) {
+		assertOptions( url, expectedDbName, "localhost", parameters);
+	}
+
+		/**
+		 * Create the default {@link SqlConnectOptions} with the given extra properties
+		 * and assert that's correct.
+		 */
```
DavideD, line 134: Indentation seems wrong here

DavideD (changes requested):
Looks good, thanks.
I would have merged it but it doesn't seem to work with SQL Server.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java
blafond: fixed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
blafond: fixed

DavideD (changes requested):
I don't think it's going to work if a user does not specify a port.
Could you check this use case, please?
```

	@Test
	public void testInvalidHostWithoutPort() {
		Map<String, String> params = new HashMap<>();
		params.put( "user", "hello" );

		// Port -1 so it won't be added to the url
		String url = createJdbcUrl( "local_host", -1, "my_db", params );
		assertOptions( url, "my_db", "local_host", params );
	}
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
```diff
@@ -305,6 +295,45 @@ private String oraclePath(URI uri) {
...
+			char endOfHostPortChar = ( scheme.equals( "sqlserver" )) ? ';' : '/';
+			String hostPortString = remainder.substring( 0, remainder.indexOf( endOfHostPortChar ) );
+			int startOfPort = hostPortString.indexOf( ':' );
+			if ( startOfPort == -1 ) {
+				return -1;
+			}
+			return Integer.valueOf( hostPortString.substring( startOfPort+1 ) );
+		}
+
+		return uri.getPort();
```
DavideD, line 334: I don't understand all this extra logic. 
What's going on here?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java
```diff
@@ -125,7 +155,7 @@ private void assertOptions(String url, String expectedDbName, Map<String, String
 		assertThat( options.getUser() ).as( "URL: " + url ).isEqualTo( username );
 		assertThat( options.getPassword() ).as( "URL: " + url ).isEqualTo( password );
 		assertThat( options.getDatabase() ).as( "URL: " + url ).isEqualTo( expectedDbName );
-		assertThat( options.getHost() ).as( "URL: " + url ).isEqualTo( "localhost" );
+		assertThat( options.getHost() ).as( "URL: " + url ).isEqualTo( expectedHost );
 		assertThat( options.getPort() ).as( "URL: " + url ).isEqualTo( dbType().getDefaultPort() );
```
DavideD, line 159: We need to add some tests for checking that we read the correct port if set.

DavideD (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPoolConfiguration.java
DavideD: Sorry about it, I see that there was some extra logic even before.
I think we should collect all these cases in a  unique `findPort` method.
Something like:
```
int findPort(...) {
    int port = -1;
    if ("oracle".equals(scheme) ) {
        port = oraclePort(...);
    }
    else  if ("sqlServer".equals(scheme) ) {
        port = sqlServerPort(...);
    } else {
        port = ...
    }
    if ( port == -1) {
       return defaultPort();
    }
     return port;
}
```

It makes it easier to refactor and understand what's going on

blafond: agreed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/DefaultPortTest.java
```diff
@@ -32,7 +32,11 @@ public class DefaultPortTest {
 	public void testDefaultPortIsSet() throws URISyntaxException {
 		DefaultSqlClientPoolConfiguration configuration = new DefaultSqlClientPoolConfiguration();
 		configuration.configure( requiredProperties() );
-		SqlConnectOptions sqlConnectOptions = configuration.connectOptions( new URI( scheme() + "://localhost/database" ) );
+		// all supported DB URLs except for sqlserver use the '/' delimiter separating host/port and user/db info
+		// sqlserver URL uses ';' as the delimeter so added a check for sqlserver to insure connectOptions() call succeeds
+		// with valid URL syntax
+		String partialValidUrl = scheme().equalsIgnoreCase( "sqlserver" ) ? "://localhost;database=my_db" : "://localhost/database=my_db";
+		SqlConnectOptions sqlConnectOptions = configuration.connectOptions( new URI( scheme() + partialValidUrl ) );
```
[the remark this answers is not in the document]
blafond: Done

hibernate-reactive-core/src/test/java/org/hibernate/reactive/configuration/JdbcUrlParserTest.java
blafond: Done

DavideD:
Merged, thanks

PR #1438, opened 2023-01-18
Test Hibernate Reactive against JDK 21 EA, JDK 19 GA

yrodiere:
JDK 18 EOL'd on 2022-09-20
See https://endoflife.date/java


DavideD:
Merged, thanks

PR #1443, opened 2023-02-09
Fix MultilineImportsTest

geoand:
The fix is to use MultiLineSqlScriptExtractor instead of MultipleLinesSqlCommandExtractor (which was deprecated anyway and now removed in ORM6)


DavideD:
Thanks

PR #1455, opened 2023-02-14
Fix MutinyExceptionsTest

geoand:
Hibernate ORM no longer wraps `HibernateException`
in a `PersistenceException`, as per [HHH-15797](https://hibernate.atlassian.net/browse/HHH-15797)


DavideD:
Nice

PR #1456, opened 2023-02-15
Fix LockOnUpdateTest

geoand:
This is done by following same pattern that we have in main: essentially make a reactive copy of the CacheEntityLoaderHelper (which exists in main but was deleted to make the project compile against ORM 6)

With this change `org.hibernate.reactive.LockOnUpdateTest` passes and there are no new failures

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/ReactiveCacheEntityLoaderHelper.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/ReactiveCacheEntityLoaderHelper.java
@@ -1,0 +1,148 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.loader.entity;
+
+import static org.hibernate.loader.ast.internal.LoaderHelper.upgradeLock;
+import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.LockOptions;
+import org.hibernate.ObjectDeletedException;
+import org.hibernate.cache.spi.access.EntityDataAccess;
+import org.hibernate.cache.spi.access.SoftLock;
+import org.hibernate.engine.spi.EntityEntry;
+import org.hibernate.engine.spi.EntityKey;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.engine.spi.Status;
+import org.hibernate.event.spi.EventSource;
+import org.hibernate.event.spi.LoadEvent;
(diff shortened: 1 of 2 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/loader/entity/ReactiveCacheEntityLoaderHelper.java
```diff
@@ -0,0 +1,148 @@
...
+				}
+
+				if ( persister.isVersioned() && requestedLockMode == LockMode.PESSIMISTIC_FORCE_INCREMENT  ) {
+					// todo : should we check the current isolation mode explicitly?
+					Object nextVersion = persister.forceVersionIncrement( entry.getId(), entry.getVersion(), source );
+					entry.forceLocked( object, nextVersion );
+				}
+				else {
+					( (ReactiveEntityPersister) persister )
+							.reactiveLock( entry.getId(), entry.getVersion(), object, lockOptions, source );
```
gavinking, line 134: Ufffff, this isn't right.

Did I really write that?! Terrible.

geoand: Just for my own knowledge, what's wrong about it?

gavinking: It just throws away the `CompletionStage`. It's never actually going to hit the database and lock the row.

gavinking: It's like the easiest and most basic error to make in writing reactive code.

geoand: Totally

geoand: Actually, it will hit the database, it's a `CompletionStage`, not a `Uni`, but it obviously won't be done in the proper sequence

DavideD: @geoand Are you still working on this? Otherwise we can merge it as it is and create a separate issue to fix this (we have the same bug in main anyway).

geoand: Personally I'm in favor of getting parity with `main` as the failing test count is still high 🙂

DavideD:
I've created https://github.com/hibernate/hibernate-reactive/issues/1458

geoand:
👌

PR #1465, opened 2023-02-20
Fix compilation errors with ORM snapshot

DavideD:
@geoand, this should be with the snapshot now. Would you like to try it?


DavideD:
I've merged it, shouldn be worse than the previous one anyway

DavideD:
Np, sorry for the previous mistake

geoand:
No worries :)

PR #1467, opened 2023-02-22
Introduce batch single table batch support

geoand:
This change leverages the fact that when batching
is enabled, a BatchingConnection is provided
under the hood that takes care of all the necessary plumbing.

Before the change, **122** tests where failing, after this change the number drops to **87**


DavideD:
Awesome!

geoand:
🤟🏼

PR #1473, opened 2023-02-27
Fix initialization of EmulatedSequenceReactiveIdentifierGenerator

Sanne:
This avoid EmulatedSequenceReactiveIdentifierGenerator to fail to initialize: `renderedTableName` is already correctly initialized by the call to `super#configure`


DavideD:
Merged, thanks

PR #1476, opened 2023-02-28
Fix FetchedAssociationTest

geoand:
The assertion of the test needed to updated to the ORM 6 generated queries (the queries as essentially the same, but the sequences and names of variables have changed).

Furthermore, the custom cleanup is done because the actual test itself passes (with the updated assertion), but the DB clearing fails because of the association loading issue.


DavideD:
Nice, thanks

PR #1484, opened 2023-03-01
Work around lazy eager loading issue in clean DB for EagerOneToOneAssociationTest

geoand:


DavideD:
It's good to have a workaround, but this issue is going to be fixed by the reactive initializers.

geoand:
👌

Issue #1384, opened 2022-09-06
UnexpectedAccessToTheDatabase when querying entity with IdClass with association


markusdlugi:
When you have an entity with a composite identifier that has an association to another entity (e.g., like in `Example 134. IdClass with @ManyToOne` in the [Hibernate User Guide](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#identifiers-composite-nonaggregated)) then any attempt to query such entities will result in an exception.

So given for example the following entities:

```java
@Entity
public class GroceryList implements Serializable {

    @Id @GeneratedValue
    private Long id;

    @OneToMany(mappedBy = "groceryList", fetch = FetchType.LAZY, cascade = CascadeType.ALL, orphanRemoval = true)
    private List<ShoppingItem> shoppingItems;

}
```

```java
@Entity
@IdClass(ShoppingItem.ShoppingItemId.class)
public class ShoppingItem implements Serializable {

    @Id
    @ManyToOne
    @JoinColumn(name = "grocerylistid")
    private GroceryList groceryList;

    @Id
    private String itemName;

    private int itemCount;

    public static class ShoppingItemId implements Serializable {

        private String itemName;
        private GroceryList groceryList;
    }
}
```

When you try to run a simple query like this:

```hql
FROM ShoppingItem WHERE grocerylistid = :id
```

Then you will receive the following exception:

```
2022-09-06 14:32:55,518 ERROR [org.hib.rea.errors] (vert.x-eventloop-thread-2) HR000057: Failed to execute statement [select shoppingit0_.grocerylistid as groceryl3_1_, shoppingit0_.itemName as itemname1_1_, shoppingit0_.itemCount as itemcoun2_1_ from ShoppingItem shoppingit0_ where grocerylistid=$1]: could not execute query: java.util.concurrent.CompletionException: org.hibernate.reactive.event.impl.UnexpectedAccessToTheDatabase: Unexpected access to the database
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1081)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
[...]

PR #1485, opened 2023-03-01
Work around lazy eager loading issue in clean DB for CompositeIdManyToOneTest

geoand:
The point of this test is to test that persisting composite id works, but without this change the test actually fails when loading the entity (due to the eager loading issue that is still unresolved).

This change makes sure that the test only fails if the persist functionality does not work properly

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CompositeIdManyToOneTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CompositeIdManyToOneTest.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import org.junit.After;
 import org.junit.Test;
 
 import io.vertx.ext.unit.TestContext;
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/CompositeIdManyToOneTest.java
```diff
@@ -71,7 +80,7 @@ public static class ShoppingItem implements Serializable {
 //        @Id Long groceryListId;
         @Id private String itemName;
 
-        @Id @ManyToOne
+        @Id @ManyToOne(fetch = FetchType.LAZY)
```
DavideD, line 83: I get that this fixes the test, but I don't think we should change the mapping for it.

We should at least replace the test with an equivalent one that uses the default or create an issue to make sure that we test the original use case.

I think that what I'm doing now will fix this test. Maybe we can wait merging this fix for now

geoand: Sure, no problem

DavideD:
Closing this PR because we already keep track of this in this issue: https://github.com/hibernate/hibernate-reactive/issues/1384

PR #1494, opened 2023-03-10
Ensure we apply the exception converter consistently

Sanne:
Thrown exceptions were being properly routed through the "convertException" utilities, but when it's wrapped in a CompletionException we'd fail to recognize it.


Sanne:
@DavideD FYI ^
I'll merge this already as it's trivial

PR #1495, opened 2023-03-11
Sub select fetch

DavideD:
@gavinking Was this test correct?
It seems that Hibernate Reactive was loading both collections instead of the one required. 

This is the commit with the fix: https://github.com/hibernate/hibernate-reactive/commit/a25646a50b91cf79b93592086ea8b7ca25b1ac50

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
@@ -74,7 +74,7 @@ public class SubselectFetchTest extends BaseReactiveTest {
 							return s.fetch( n1.getElements() ).thenAccept( elements -> {
 								context.assertTrue( Hibernate.isInitialized( elements ) );
 								context.assertTrue( Hibernate.isInitialized( n1.getElements() ) );
-								context.assertTrue( Hibernate.isInitialized( n2.getElements() ) );
+								context.assertFalse( Hibernate.isInitialized( n2.getElements() ) );
 							} );
 						} )
 				)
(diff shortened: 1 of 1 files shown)


gavinking:
> Was this test correct?

I believe so, why, what precisely is it that looks wrong to you?

> It seems that Hibernate Reactive was loading both collections instead of the one required.

You mean "both collections" as in `n1.getElements()` and `n2.getElements()`?

Well yeah, that's the whole idea of subselect fetching. It's a more powerful sort of batch fetching.

(The test looks good to me, but I might have missed something.)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
```diff
@@ -60,24 +60,24 @@ public void testQuery(TestContext context) {
...
+						.thenCompose( list -> {
+							context.assertEquals( list.size(), 2 );
+							Node n1 = list.get( 0 );
+							Node n2 = list.get( 1 );
+							context.assertFalse( Hibernate.isInitialized( n1.getElements() ) );
+							context.assertFalse( Hibernate.isInitialized( n2.getElements() ) );
+							return s.fetch( n1.getElements() ).thenAccept( elements -> {
+								context.assertTrue( Hibernate.isInitialized( elements ) );
+								context.assertTrue( Hibernate.isInitialized( n1.getElements() ) );
+								context.assertFalse( Hibernate.isInitialized( n2.getElements() ) );
```
gavinking, line 77: With subselect fetching we're hoping that both the associations get loaded at once, in one query.

DavideD:
> You mean "both collections" as in n1.getElements() and n2.getElements()?

Yes, that's what I was wondering. It seems that now it only loads `n1.getElements()`,and that's why `n2.getElements` is not initialized.

I've just found it weird that `.fetch(n1.getElements())` also load `n2.getElements()`. Just wanted to double-check.

DavideD:
Ok, the test was correct. I'm going to close this PR and create a new issue about this

PR #1497, opened 2023-03-11
Merge reference

DavideD:
This is an unusual use case where the association is with an entity
column instead of the identifier. It's a specific case and the current
mapping, as it is, fails using regular Hibernate ORM anyway.

I've added a test with the persist that seems to work as expected.


DavideD:
Closing this for now

gavinking:
> This is an unusual use case where the association is with an entity column instead of the identifier.

Well, it's unusual, yes, but it's something we're definitely supposed to support.

> the current mapping, as it is, fails using regular Hibernate ORM anyway.

That surprises me. What is the error. I believe it should work and if it doesn't, then it would be a bug in H6.

DavideD:
Let me rewrite the test for ORM

DavideD:
Here's the test for ORM: https://github.com/DavideD/hibernate-orm/commit/a4b444845b39bf67f6626be43b24b1c736d3fed8

Error:
```
org.hibernate.AssertionFailure: non-transient entity has a null id: org.hibernate.orm.test.embeddable.MergeWithReferenceTest$Bar$HibernateProxy$2WSQf5sQ
```

Same as the Hibernate Reactive one

PR #1498, opened 2023-03-11
Merge and persist with reference

DavideD:
This is an unusual use case where the association is with an entity
column instead of the identifier. It's a specific case and the current
mapping, as it is, fails using regular Hibernate ORM anyway.

I've added a test with the persist that seems to work as expected.


DavideD:
Replace the previous PR: https://github.com/hibernate/hibernate-reactive/pull/1497

Here's the test for ORM: https://github.com/DavideD/hibernate-orm/commit/a4b444845b39bf67f6626be43b24b1c736d3fed8

Test fails with error:
```
org.hibernate.AssertionFailure: non-transient entity has a null id: org.hibernate.orm.test.embeddable.MergeWithReferenceTest$Bar$HibernateProxy$2WSQf5sQ
```

Same as the Hibernate Reactive one

DavideD:
I've created an issue: https://github.com/hibernate/hibernate-reactive/issues/1504

PR #1510, opened 2023-03-13
Prep for ORM 6.2

sebersole:
Accounts for changes from 

- https://hibernate.atlassian.net/browse/HHH-16284
- https://hibernate.atlassian.net/browse/HHH-16290

Note - many tests fail for me with these changes, but I was unable to run tests without them so ...


DavideD:
Thanks, it passes all the expected tests

PR #1511, opened 2023-03-14
Enable ci tests using latest Hibernate snapshots

DavideD:
This also changes to sonatype link to: https://oss.sonatype.org/content/repositories/snapshots/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,7 +99,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


DavideD:
@blafond @Sanne 
Now `main` is using the ORM snapshot from the remote repository.
If you want to test a local copy of ORM, remember to enable the use of local maven repository [here](https://github.com/hibernate/hibernate-reactive/blob/main/gradle.properties#L38).

The build uses the latest stable version of the ORM Gradle plugin because I couldn't figure out how to download the snapshot from a remote repository.

Issue #1512, opened 2023-03-14
Auto-flush not triggered


Sanne:
It would seem that when we have pending objects to be persisted and a query is triggered on the same type, auto-flush is not being triggered.
Attaching a test soon in a related PR.

PR #1513, opened 2023-03-14
Test for auto-flush and ordering of executions

Sanne:
Test for issue #1512 

This introduces a test for the missing auto-flush capababilities in Hibernate Reactive;
the test is currently passing as there's an explicit flush() : this should be unnecessary.

Included an ORM test as well for comparison.


Sanne:
Superseded by #1519

PR #1519, opened 2023-03-14
Fixing auto-flush on mutation queries

### LINKED ISSUE #1512: Auto-flush not triggered

Sanne:
Fixes #1512


DavideD:
Awesome!

PR #1552, opened 2023-04-03
Ignore `hibernate.jdbc.use_get_generated_keys`

DavideD:
This property might make sense with JDBC. But the Vert.x client is less flexible about the type of queries it needs to read the id.


gavinking:
Also, deleting 140 lines of code is nice :)

PR #1555, opened 2023-04-04
Add Db2

DavideD:
Fix https://github.com/hibernate/hibernate-reactive/issues/1550

I had to disable some tests because of https://github.com/eclipse-vertx/vertx-sql-client/issues/1303 and https://github.com/eclipse-vertx/vertx-sql-client/issues/899
@tsegismont, do you have time to have a look at these? ^

Even with the errors, I think we can merge this.
I will keep track of the disabled tests with separate issues.


tsegismont:
> @tsegismont, do you have time to have a look at these? ^

I'll take a look, it might be something which does not require deep knowledge of the protocol

PR #1573, opened 2023-04-07
Timezone tests

gavinking:


gavinking:
So the tests are failing on Java 17+ due to a change in precision of the Java clock. This is a pain.

gavinking:
@DavideD is there an easy way to exclude the new tests from certain jobs?

gavinking:
The failure on SQL Server is the result of #1577.

DavideD:
> @DavideD is there an easy way to exclude the new tests from certain jobs?

What's the right approach to use in this case?
Can't we just compare the values truncated to the MILLIS and call it a day?

Something like this: https://github.com/hibernate/hibernate-reactive/pull/1581/commits/c60928d7a5c5df227d010b521cffc498e6127840

gavinking:
@DavideD Fine by me.

DavideD:
Superseeded by https://github.com/hibernate/hibernate-reactive/pull/1581

PR #1575, opened 2023-04-07
Autodetect the dialect version

### LINKED ISSUE #1564: pass the DatabaseVersion to the Dialect constructor
In Hibernate 6, we are supposed to pass a `DatabaseVersion` to the constructor of the `Dialect`. This is really important, because if we don't do it, the dialect defaults to an old version of the database.

Vert.x exposes this info via `DatabaseMetadata`.

It's critical we get this fixed before the release.

DavideD:
Fix #1564 

Requires https://github.com/hibernate/hibernate-orm/pull/6371


DavideD:
@gavinking (maybe @Sanne , if interested) I think this is now ready:

* For SQL Server we are going to use the dialect version 10 because the latest one uses a query that doesn't work with Vert.x. This will print an additional line of log in the console to warn users.
* We need a custom dialect for CockroachDB, because the one in ORM will run a query. I think we can avoid it by changing something in ORM, but I haven't looked into it yet (I wanted a workaround that doesn't require changes in ORM).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
```diff
@@ -278,6 +285,10 @@ public interface Log extends BasicLogger {
 	@Message(id = 353, value = "Could not release a cache lock : %s" )
 	void unableToReleaseCacheLock(CacheException ce);
 
+	@LogMessage(level = WARN)
+	@Message(id = 400, value= "Changing dialect \"%1$s\" to \"%2$s\"")
```
gavinking, line 289: Very minor thing: I would prefer single quotes here.

gavinking:
OK, great! Looks good!

PR #1576, opened 2023-04-07
test demoing bug #1577 on SQL Server

gavinking:


gavinking:
This is actually a bug in ORM core, and will be anyway resolved by setting the correct version of the MySQL dialect.

PR #1590, opened 2023-04-13
Re-simplify query APIs

gavinking:
The redesign resulted in 80-method APIs, plenty of which were simply not relevant to the reactive usecase. I've rolled back the API to something much more similar to what we had before.


gavinking:
Only done for the `Stage` flavor of the API. `Mutiny` is still todo.

gavinking:
`Mutiny` done now.

PR #1610, opened 2023-04-24
Lock-free BlockingIdentifierGenerator LoHi

franz1981:
This is a version that pay a single allocation cost per each  `blockSize` to save any synchronized to happen.
It can be made gc-free by playing some bit tricks packing both lo/hi into a `long` (and assuming there are no negative values for both, likely) - or with some more complex coordination mechanism (yet to be designed).

I don't know few things yet before un-drafting:
1. If many concurrent `state.next(hi)` happen do we need to enforce validation of (hi,lo) based on existing values? new hi > old hi (or diff by blockSize?) && old low >= blockSize. 
2. If many concurrent `state.next(hi)` happen do we need to ensure some ordering of updates?  eg concurrent `next(10), next(20),next(30)` updates in this pr are not yet guaranteed to be set in any specific order and  I cannot see any ordering enforced in the original code, but better be safe then sorry with concurrency 
2. blockSize is a constant? I'm assuming it is...


hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
```diff
@@ -56,22 +57,55 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
...
+					return hi + nextLo;
+				}
+				return -1;
+			}
+		}
+
+		private volatile LoHi loHi;
+
+		public long hi(long hi) {
+			loHi = new LoHi(hi);
```
franz1981, line 87: @Sanne 
This could be made much more strict and based on the previous value while failing if it happens concurrently too (by using a `compareAndSet`)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
```diff
@@ -56,22 +57,52 @@ public abstract class BlockingIdentifierGenerator implements ReactiveIdentifierG
...
 	//supposed to work on.
 	private static class GeneratorState {
-		private int loValue;
-		private long hiValue;
+
+		private static final class LoHi {
+
+			private static final AtomicIntegerFieldUpdater<LoHi> LO_UPDATER = AtomicIntegerFieldUpdater.newUpdater(LoHi.class, "lo");
+			private final long hi;
+			private volatile long lo;
```
Sanne, line 65: should this be an `int` ? Especially as you're using a `AtomicIntegerFieldUpdater` ?

Sanne:
I'd challenge you in ever noticing the current `synchronized` on profiling data :)

franz1981:
>  if we met later in the week as I'm facing 200+ emails and dms this morning 🥴

eheh I know, but it won't cost anything to make it lock free like this I think, I'm more concerned instead by this -> https://github.com/hibernate/hibernate-reactive/pull/1610#issuecomment-1519467511

franz1981:
To quote myself of the future (and @Sanne ) re https://github.com/hibernate/hibernate-reactive/pull/1610#issuecomment-1519467511

this is not a "real" problem leading to break anything, but just an efficiency one; in short, if more then one concurrently try to increase the sequence due to exhaustion and succeed, but update `hi/lo` in the "wrong" order (ie by creating gaps or "jumping forward" before previously existing ones will be "consumed") it can just cause the consumers to request to the DB more sequences then necessary (again) because in both cases (missing new or "old" created batches) would still cause someone to be left unsatisfied.
The existing tests are not verifying this hence there was no reason to noticed that.

franz1981:
@Sanne undrafting, but feel free to pick what you like for next works on this bud ;)

PR #1616, opened 2023-05-03
Remove legacy ProxyConnection stuff

gavinking:


DavideD:
Did you also check if it's still used in Quarkus?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -1721,7 +1721,7 @@ public ReactiveConnection getReactiveConnection() {
 
 	@Override
 	public void close() throws HibernateException {
-		throw new UnsupportedOperationException( "Non reactive close method called. Use 'reactiveClose()' instead." );
+		throw new UnsupportedOperationException( "Non-reactive 'close()' method called. Use 'reactiveClose()' instead." );
```
DavideD, line 1724: Actually, we now have a specific error in the Log now: https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java#L243

gavinking:
> Did you also check if it's still used in Quarkus?

Yes, @Sanne and I looked at this yesterday. Quarkus no longer uses it. `MutinyImplementor` *does* occur in one annotation, but that occurrence can just be deleted.

DavideD:
Nice

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
gavinking: OK, but we can fix that up in a subsequent commit.

PR #1635, opened 2023-05-24
Upgrade test suite to JUnit 5

DavideD:


blafond:
I pulled your branch and built locally.  I'm only getting one failure and that's with MSSQL/SQLServer... a timeout in `MultithreadedInsertionTest.testIdentityGenerator()`

Increasing the class's @Timeout won't fix it for me.  I had to add back the pool timeout:
`configuration.setProperty( POOL_CONNECT_TIMEOUT, Integer.toString( 100000 ) );`

PR #1669, opened 2023-06-13
Test against JDK 22 EA

yrodiere:


yrodiere:
Looks like we might need a Gradle upgrade for this one... It seems Gradle doesn't detect JDK 22 correctly.

@DavideD any plans to upgrade to Gradle v8.1.1?

DavideD:
Yes, we have an issue and I have a branch with a prototype. But I haven't finished yet.

PR #1683, opened 2023-06-28
Experimental solution for HREACT-34

Sanne:
https://hibernate.atlassian.net/browse/HREACT-34


DavideD:
We've decided to go with the [minimal solution](https://github.com/hibernate/hibernate-reactive/pull/1684).
We will work on adding assertions and more advance checks as a separate issue.

PR #1685, opened 2023-06-28
[#1504] Merge with reference fails if entity is associated to a column (non id)

### LINKED ISSUE #1504: Merge with reference fails if entity is associated to a column (non id)
See the following tests:
* https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/LazyUniqueKeyTest.java#L71
* https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/EagerUniqueKeyTest.java#L71

I've created the following testcase for ORM: https://github.com/hibernate/hibernate-orm/pull/6239
Waiting for ORM to fix https://hibernate.atlassian.net/browse/HHH-16298

(Follows https://github.com/hibernate/hibernate-reactive/pull/1498)

blafond:
fixes #1504 

The two tests were failing because the `Bar` entity column was still a `HibernateProxy` object in `EntityTypes` methods so the identifier `Bar.key` value was always null.

The fixes I implemented check for the proxy and force a load via `reactiveImmediateLoad()`

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/EntityTypes.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/EntityTypes.java
@@ -10,27 +10,37 @@ import java.util.concurrent.CompletionStage;
 
 import org.hibernate.AssertionFailure;
 import org.hibernate.HibernateException;
+import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
 import org.hibernate.engine.spi.EntityKey;
 import org.hibernate.engine.spi.EntityUniqueKey;
 import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.engine.spi.PersistentAttributeInterceptor;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.persister.entity.EntityPersister;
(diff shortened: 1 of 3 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/EntityTypes.java
```diff
@@ -124,13 +134,29 @@ static CompletionStage<Object> loadByUniqueKey(
...
+						// The unique key may be a HibernateProxy object and needs to be loaded before
+						// adding to persistenceContext
+						if ( loadedUniqueKey instanceof HibernateProxy ) {
+							try {
+								CompletionStage<Object> resultObject = loadHibernateProxyEntity(
+										(HibernateProxy) loadedUniqueKey,
+										session
+								);
+								loadedUniqueKey = ((CompletableFuture)resultObject).get();
+							}
```
DavideD, line 148: You should not use a `.get()` because it's a blocking method.
 Assuming the logic is correct (I haven't cheked yet), the code should look something like:
```
.thenCompose ukResult -> {
      ...
      if ( loadedUniqueKey instanceof HibernateProxy ) {
         return   loadHibernateProxyEntity( ... )
                            .thenApply(loadedUniqueKey -> { .... } );
      }  
      ... 
}
```

DavideD:
Thanks, looks good.

I've just decided to keep the original tests and took the opportunity to clean up a bit the code and remove some warnings in EntityTypes: https://github.com/hibernate/hibernate-reactive/pull/1696

PR #1690, opened 2023-06-30
Fix race condition during updates

### LINKED ISSUE #1686: Add back integration tests with Vert.x
We used to have an[ integration test to check how Hibernate Reactive can be used with Vert.x](https://github.com/hibernate/hibernate-reactive/tree/1.1/integration-tests/verticle-postgres-it).

We should add it back. Aside from testing functionality, it's also helpful for quick benchmarks and to see what happens under load.

On the same note, we should add a test that emulates the [TechEmpower benchmark](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java/quarkus). Not to use for performances, but to prevent and debug errors related to the test.

### LINKED ISSUE #1687: Fix race condition during the update of multiple entites
Under load, the CompletionStage responsible for signaling a successful update operation might not complete. This causes the user's application  to block without apparent reason.

The bug is in [ReactiveUpdateStandardCoordinator](https://github.com/DavideD/hibernate-reactive/blob/7fdc676c72252860445bfabfc05b6a65613588d1/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/mutation/ReactiveUpdateCoordinatorStandard.java#L148).

DavideD:
Fix #1687 
Fix #1686

### DIFF
--- a/examples/native-sql-example/build.gradle
+++ b/examples/native-sql-example/build.gradle
@@ -1,7 +1,25 @@
+// Optional: We configure the repositories to ease development and CI builds
+buildscript {
+    repositories {
+        // Optional: Enables the maven local repository
+        // Example: ./gradlew build -PenableMavenLocalRepo
+        if ( project.hasProperty( 'enableMavenLocalRepo' ) ) {
+            // Useful for local development, it should be disabled otherwise
+            mavenLocal()
+        }
+        mavenCentral()
+    }
+}
+
+plugins {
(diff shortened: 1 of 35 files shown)


Sanne (approved):
Excellent!

@FroMage @franz1981 FYI we now have some variations of the Techempower stress tests included as integration tests within the Hibernate Reactive codebase, so at least the `/updates` endpoint logic is being tested upstream.
N.B. in this case we don't use Quarkus but vertx-web

PR #1701, opened 2023-07-11
various fixes to Session API

gavinking:
1. add `Session.getFactory()` and `StatelessSession.getFactory()`
2. fix the signature of `mergeAll()`
3. fix the broken signature of `createSelectionQuery()` oops!
4. #1703 add `enableFetchProfile()` to Query
5. deprecate `Session.createQuery()` to align with deprecation in H 6.3
6. add `createSelectionQuery()`/`createMutationQuery()` to `StatelessSession`

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -347,6 +347,12 @@ public interface Mutiny {
 		 */
 		SelectionQuery<R> setPlan(EntityGraph<R> entityGraph);
 
+		/**
+		 * Enable a {@linkplain org.hibernate.annotations.FetchProfile fetch
+		 * profile} which will be in effect during execution of this query.
+		 */
+		SelectionQuery<R> enableFetchProfile(String profileName);
+
 		@Override
 		SelectionQuery<R> setParameter(int parameter, Object argument);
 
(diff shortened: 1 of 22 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1464,6 +1486,30 @@ default <T> Uni<T> get(Class<T> entityClass, Object id, LockModeType lockModeTyp
 		 */
 		<R> SelectionQuery<R> createQuery(String queryString, Class<R> resultType);
```
DavideD, line 1487: Shouldn't this also have a `@Deprecated` (with link to createSelectionQuery)?

gavinking: No that one is OK. It's the ones without a `resultType` that are the problems.

gavinking: (I mean, it's _redundant_, but no need to deprecate it, IMO.)

DavideD:
We will merge this one in HR 2.1

DavideD:
Thanks a lot!
I've rebased this to the latest changes and merged it into https://github.com/hibernate/hibernate-reactive/pull/1726

Issue #1673, opened 2023-06-16
Check that we actually run the custom queries in CustomStoredProcedureSqlTest


DavideD:
In [CustomStoredProcedureSqlTest](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CustomStoredProcedureSqlTest.java), we use custom queries with stored procedures to test the support for the annotation `@SqlInsert`, `@SqlUpdate`, and `@SqlDelete`.

It works fine, but we are not actually checking that we are running the custom queries for the CRUD operations.

We usually check the log using our custom [SqlStatementTracker](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/testing/SqlStatementTracker.java) class

Nesrinasan:
Hi, I want to examine this issue. Please assign me. And Would I get some more information about the issue? @DavideD

DavideD:
Hi,
the assertions in this test check the values in the entities.

But I would also like to add some assertions on the generated SQL, and make sure that we are running the SQL specified by the annotations.
For example, for the insert, we should see in the log `SELECT PROC_INSERT_RECORD( $1, $2 );`,  instead of `INSERT INTO ...`.

An existing test that does something similar is [OrderQueriesTestBase](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderQueriesTestBase.java).

We already know that this feature works, so we only need to add the new assertions to the existing ones.

I haven't added a contribution guide for this project yet, but you can find the templates for the code on the [hibernate/hibernate-ide-templates](https://github.com/hibernate/hibernate-ide-codestyles) (the `hibernate_orm` ones).
And start the commit message with the issue code, please. For example: `[#1673] Check the generated SQL in  CustomStoreProcedureSQLTest `.

Thanks a lot, and feel free to ask more questions.

Nesrinasan:
Hi Davide. I started to write the test. while I write test, I am using @SqlInsert, @SqlUpdate, and @SqlDelete. 
for example:  I will use these annotations(@SqlInsert, @SqlUpdate, and @SqlDelete. ) while writing the insert, delete, update test between one-to-one related tables. 

Did I understand correctly? @DavideD

DavideD:
Sounds correct.

[...]

PR #1710, opened 2023-07-18
[#1673] Check the generated SQL in  CustomOneToOneStoredProcedureSqlT

Nesrinasan:
add test storeProsedure using hibernate @SqlInsert and @SqlDelete annotation


DavideD:
Thanks, but I don't understand the difference between this test and [the one we already have](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CustomStoredProcedureSqlTest.java). Am I missing something?

DavideD:
Merged, with a change to the commit message

DavideD:
Thanks a lot!

PR #1712, opened 2023-07-20
Exploring the idea of potentially move away from internal use of JDK's CompletionStage and CompletableFuture

Sanne:
This is very rough and not fully compiling; needs some changes in Hibernate ORM core to be able to be finalized.

Leaving it here for the moment so we can discuss the approach before going too far.

The general idea would be that JDK's CompletionStage system is very feature rich, but so doing exposes several methods which are actively dangerous for our purposes, and it consumes more memory than a lighter version could; by having our own contract we can restrict what we use.
Also, we don't need (I thought? there's some evidence to the contrary) all the multi-threading capabilities the standard types offer: we just need to model a linear chain of events, with error propagation. The multi-threading safeguards are a performance drag, which we don't really need as long as we can verify correct design.

The additional benefit of using a custom interface would be that integration tests could replace the underlying implementation with something which more extensively traces the lifecycle of the events; for example we could verify that:
 - each chain is strictly tied to the expected threads exclusively (so to verify we really don't need the multi-threading capabilities)
 -  all stages are actually being completed, as we had cases for critical bugs in which some completion stage would get "lost" in the chain of operations.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -32,15 +32,15 @@ org.gradle.java.installations.auto-download=false
 #enableSonatypeOpenSourceSnapshotsRep = true
 
 # Enable the maven local repository (for local development when needed) when present (value ignored)
-#enableMavenLocalRepo = true
+enableMavenLocalRepo = true
 
 # Override default Hibernate ORM version
-#hibernateOrmVersion = 6.2.3.Final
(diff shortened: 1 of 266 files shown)


DavideD:
Closing this because we decided for a different approach

PR #1719, opened 2023-07-24
Upgrade to Hibernate ORM 6.2.7.Final

DavideD:
[I had to change one of the test](https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:ORM-6.2.7?expand=1#diff-3e90f461f2350c2e82be9fad23b31da034848a9ed9eb2cf43a501c8c520e44fdR486) because now Hibernate ORM throw an exception if an associated entity has  `null` version (before it was treated as a transient entity).

I've added a fetch before removing the entity.
@gavinking, do you think this is acceptable or should we recognize the  use case and  do something smarter?

Basically, before the user was able to delete the entity without fetching it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ version = projectVersion
 // ./gradlew clean build -PhibernateOrmVersion=5.6.15-SNAPSHOT
 ext {
     if ( !project.hasProperty('hibernateOrmVersion') ) {
-        hibernateOrmVersion = '6.2.6.Final'
+        hibernateOrmVersion = '6.2.7.Final'
     }
     if ( !project.hasProperty( 'hibernateOrmGradlePluginVersion' ) ) {
         // Same as ORM as default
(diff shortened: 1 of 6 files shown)


gavinking:
So in principle that change was not supposed to have affected entities with initialized ids. (I reviewed this.)

DavideD:
Ok, I will check why the version is null in that particular case

gavinking:
@DavideD I don't follow what is going on here at all. I don't even see where there are detached entities in that test.

But what is *supposed* to happen here is that that error should occur *only* if the detached entity has an initialized id but an uninitialized version.

If that's _not_ what's happening, then there's a bug somewhere (in ORM core, or in reactive).

DavideD:
> But what is supposed to happen here is that that error should occur only if the detached entity has an initialized id but an uninitialized version.

This is what's happening, there's an id but not a version (it's null before the fetch).
I'm not sure if this is an error in ORM.
OK, I will have a better look and close the PR for now, Thanks for checking

Issue #1442, opened 2023-02-09
Test TimeZoneStorage annotation


DavideD:
See https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#_timezonestorage

After we have finished the migration to Hibernate ORM 6.2

PR #1722, opened 2023-07-26
[#1442] Test TimeZoneStorage

blafond:
fixes [#1442]

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/timezones/TimeZoneStorageMappingTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/timezones/TimeZoneStorageMappingTest.java
@@ -1,0 +1,301 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.timezones;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.List;
+
+import org.hibernate.annotations.TimeZoneColumn;
+import org.hibernate.annotations.TimeZoneStorage;
+import org.hibernate.annotations.TimeZoneStorageType;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.BaseReactiveTest;
+import org.hibernate.reactive.testing.DBSelectionExtension;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import jakarta.persistence.Tuple;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.cfg.AvailableSettings.TIMEZONE_DEFAULT_STORAGE;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.SQLSERVER;
+
+/**
+ * Adapted from org.hibernate.orm.test.mapping.basic.TimeZoneStorageMappingTests
+ *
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/timezones/TimeZoneStorageMappingTests.java
```diff
@@ -0,0 +1,308 @@
...
+						assertThat( resultList.get( 0 ).get( 4, OffsetDateTime.class ).toInstant()).isEqualTo( OFFSET_DATE_TIME.toInstant() );
+						assertThat( resultList.get( 0 ).get( 5, ZonedDateTime.class ).toInstant()).isEqualTo( ZONED_DATE_TIME.toInstant() );
+					}
+				)
+			)
+		);
+	}
+
+	// This converted test fails during the getResultList()
+	// Unsure how to enable the "@RequiresDialectFeature(....)" annotation which is an ORM JUnit construct
```
DavideD, line 205: How does the annotation check if the dialect is valid or not?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/timezones/TimeZoneStorageMappingTests.java
```diff
@@ -0,0 +1,308 @@
...
+							"select " +
+									"e.offsetTimeNormalized, " +
+									"e.offsetDateTimeNormalized, " +
+									"e.zonedDateTimeNormalized, " +
+									"e.offsetTimeNormalizedUtc, " +
+									"e.offsetDateTimeNormalizedUtc, " +
+									"e.zonedDateTimeNormalizedUtc " +
+									"from TimeZoneStorageEntity e",
+							Tuple.class
+					).getResultList()
```
DavideD, line 188: if you use `getSingleResult()`, you can avoid having `resultList.get(0)` each time

blafond: right.. simplify ;)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/timezones/TimeZoneStorageMappingTests.java
blafond: [DialectFeatureChecks](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-testing/src/main/java/org/hibernate/testing/orm/junit/DialectFeatureChecks.java) but it's over in their `hibernate-testing` package

This class is coded with options/features for all the dialects.  The @Disable'd test can't access their testing src.

DavideD: JUnit 5 has the annotation `@DisableIf`. If you check the condition, we can figure out to add it to our test.
I don't think we need something as sophisticated as the one in Hibernate ORM (for now at least).

DavideD: We should be able to get the dialect from the factory

blafond: The existing `@DisabledIf` annotation use text-based scripting language...  accessing system properties.

I did find an [example](https://javabydeveloper.com/junit-5-conditional-test-execution-with-examples/) of writing custom `ExecutionCondition` classes that can use java code, which currently resides in specific Dialects and `DialectFeatureChecks`.  Seems like a lot of work right now.

blafond: .. check that

I created a `doNotRunWithDialect()` method in the test using ORM's logic for their checks.

`@DisabledIf( "doNotRunWithDialect" )`

DavideD: I will have to look into it, but please, try to use better naming.
It's not clear at all what `doNotRunWithDialect` means.


In this case, this might make more sense:
```
@EnableIf("supportsTimeZoneStorage")
```

Note that I'm suggesting some ideas, I haven't look into this issue in depth.
We could also create an extension if it's easier.

blafond:
The design looks clean but Just discovered that that @EnabledIf("doIfTrue") annotation get's run via reflection before our our hibernate config is set-up, so `sessionFactory` hasn't started and  `dialect` is null, so getDialect() fails.... so only the non-annotated tests will actually run.

Going to take more work.

blafond:
@DavideD Pushed a change that works for all DB's except **SQLSERVER**.  Fixed my local **DB2** & **ORACLE** failures by adding the missing `@Timeout(value = 10, timeUnit = MINUTES)' annotation

I'm good with this version.

DavideD:
Thanks, I will have a look

PR #1723, opened 2023-08-01
[#1604] remove cleanDb() override and fixed error

### LINKED ISSUE #1604: Check why cleaning the db fails for JoinedSubclassInheritanceWithManyToOneTest
[I've disabled it for now](https://github.com/hibernate/hibernate-reactive/blob/d3698dc8755992faf6a266a00e4205138f2fad62/hibernate-reactive-core/src/test/java/org/hibernate/reactive/issue/JoinedSubclassInheritanceWithManyToOneTest.java#L40) because it's not related to the issue I was trying to fix (I think). But we should check why it fails on CI when cleaning the db is enabled.

It seems to fail because something is removed in the wrong order and a constraint fails.

blafond:
fixes #1604 

- removed cleanDb() override
- changed Set.of(....) to List.of(....)

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/issue/JoinedSubclassInheritanceWithManyToOneTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/issue/JoinedSubclassInheritanceWithManyToOneTest.java
@@ -6,9 +6,8 @@
 package org.hibernate.reactive.issue;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Objects;
-import java.util.Set;
-import java.util.concurrent.CompletionStage;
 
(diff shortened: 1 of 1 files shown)


blafond:
Note it was only failing locally with Oracle

DavideD:
I mean, what's the error?

blafond:
The Set isn't ordered so the values order wasn't guaranteed in the deleteEntities() call.   For some reason the Oracle test failed and the other DB's didn't.

DavideD:
Good catch.

> The Set isn't ordered so the values order wasn't guaranteed in the deleteEntities() call.

Could you add  this to message in the commit, please?
Something like:
> Replaced the `Set` with a `List` because entity deletions must happen in the specified order

Issue #1501, opened 2023-03-13
BatchFetchTest#testQuery fails


DavideD:
See https://github.com/hibernate/hibernate-reactive/blob/a0117ad42af25357ed622dcfe2196c05d239a80f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BatchFetchTest.java#L55

DavideD:
This test has been re-enabled here: [582f971](https://github.com/hibernate/hibernate-reactive/commit/582f971)

Should work since Hibernate Reactive 2.1.0.Final

PR #1729, opened 2023-08-14
[#1501] Fix SubselectFetchTest

### LINKED ISSUE #1502: SubSelectFetchTest fails
It only load the fetched collection and not all the associations that are part of the entity.

See https://github.com/hibernate/hibernate-reactive/blob/21e2c2ad7e90599a1eb8968593b6fdc7a4ea032c/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java#L40

blafond:
fixes #1502 

- Follows ORM's [SubselectFetchTest.testSubselectFetchHql(...)](https://github.com/hibernate/hibernate-orm/blob/761ec951455eff85cd801473a97c7edb4976f717/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/fetch/subselect/SubselectFetchTest.java#L100C22-L100C22)
- initialization on a collection is forced via a call to a collections `size()`

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
@@ -14,7 +14,6 @@ import org.hibernate.annotations.Fetch;
 import org.hibernate.annotations.FetchMode;
 
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import io.vertx.junit5.Timeout;
@@ -43,7 +42,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@Disabled // see https://github.com/hibernate/hibernate-reactive/issues/1502
 @Timeout(value = 10, timeUnit = MINUTES)
 public class SubselectFetchTest extends BaseReactiveTest {
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/results/graph/entity/ReactiveAbstractEntityInitializer.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/results/graph/entity/ReactiveAbstractEntityInitializer.java
@@ -26,10 +26,12 @@ import org.hibernate.sql.results.graph.AssemblerCreationState;
 import org.hibernate.sql.results.graph.DomainResult;
 import org.hibernate.sql.results.graph.DomainResultAssembler;
 import org.hibernate.sql.results.graph.Fetch;
+import org.hibernate.sql.results.graph.basic.BasicFetch;
 import org.hibernate.sql.results.graph.entity.AbstractEntityInitializer;
 import org.hibernate.sql.results.graph.entity.EntityLoadingLogging;
 import org.hibernate.sql.results.graph.entity.EntityResultGraphNode;
 import org.hibernate.sql.results.graph.entity.LoadingEntityEntry;
+import org.hibernate.sql.results.graph.entity.internal.EntityResultInitializer;
 import org.hibernate.sql.results.jdbc.spi.RowProcessingState;
 import org.hibernate.stat.spi.StatisticsImplementor;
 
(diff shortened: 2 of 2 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java
```diff
@@ -80,6 +82,11 @@ public void testQuery(VertxTestContext context) {
 							Node n2 = list.get( 1 );
 							assertFalse( isInitialized( n1.getElements() ), "'n1.elements' should not be initialized" );
 							assertFalse( isInitialized( n2.getElements() ), "'n2.elements' should not be initialized" );
+
+							// .size() triggers loading of elements
+							assertTrue( n1.getElements().size() == 3 );
+							assertTrue( n2.getElements().size() == 2 );
+
```
gavinking, line 89: Wait, we don't want to do this.

It's the call to `s.fetch( n1.getElements() )` that is supposed to trigger initialization of _both_ collections.

Also, you can't trigger loading by just calling `size()` in reactive.

blafond: I should have worded it that `.size()` triggers initialization.

Does ORM now initialize both collections or only reactive code? (i.e. looking for a better example than `SubselectFetchTest`)

gavinking: The best thing is to read my explanation of subselect fetching in the new Guide to Hibernate 6, so you understand what it is we're trying to test here. (On phone right now so I don't have a link.)

But the idea is: the whole point of subselect fetching is that access to one collection triggers fetching of all of them.

gavinking: The link is https://docs.jboss.org/hibernate/orm/6.3/introduction/html_single/Hibernate_Introduction.html#batch-subselect-fetch.

By the way, to my eyes, the test looks correct as it was written, and shouldn't be changed.

blafond: thx.  ... the changes I pushed actually worked locally, but I should have restarted workspace. It's still failing ...  as it should now.

blafond:
`ReactiveEntityResultInitializer` was extending `ReactiveAbstractEntityInitializer` but `ReactiveAbstractEntityInitializer` was not extending `EntityResultInitializer` (was extending `AbstractEntityInitiailzer`)

Replaced the extended class of  `ReactiveAbstractEntityInitializer` with  `EntityResultInitializer` so the proper methods were called during the loading of associated collection

blafond:
Note that this change/fix did not fix #1501.  `BatchFetchTest`   `testQuery()` and `testBatchLoad()` are failing

DavideD (changes requested):
I don't understand why this fixes the issue, and I don't think it's correct even if it does.
That's said, I think you are on the right track.

The reason I don't think this is correct is that now `ReactiveEntityJoinedFetchInitializer` becomes a subclass of `EntityResultInitializer`. This is different from what's happening in ORM where `EntityResultInitializer` and `EntityJoinedFetchInitializer` are both subclasses of `AbstractEntityInitializer`. That's why in Hibernate Reactive we have `ReactiveAbstractEntityInitializer` mimicking `AbstractEntityInitializer`

Also, I don't understand why this change makes the test pass. 
The only difference between `ReactiveEntityResultInitializer` and `EntityResultInitializer` is the `CONCRETE_NAME`. Maybe that's what's causing the test to fail?

blafond:
ORM's 6.2 version of [SubselectFetch](https://github.com/hibernate/hibernate-orm/blob/d78681eaa27439d480f5cedb6bb890cefe861266/hibernate-core/src/main/java/org/hibernate/engine/spi/SubselectFetch.java#L176) contains a check on `EntityResultInitializer` which explains why my _fix_ works.

In ORM 6.3, they've cleaned up `SubselectFetch` and that check is gone.

So I can merge my commit if we need to release another HR version off of 6.2, or push this issue to our 6.3 version?

DavideD:
We cannot merge this commit.

We could avoid extending `ReactiveAbstractEntityInitializer` and convert it to a delegator similar to what we do in `ReactiveSingleTableEntityPersister`

blafond:
k... I'll check that out. thx

DavideD:
Note that this issue was solved by [a change in Hibernate ORM](https://github.com/hibernate/hibernate-orm/pull/7232).
From this PR we only need[ the commit that re-enable the test](https://github.com/hibernate/hibernate-reactive/pull/1729/commits/4ba8cf8412cdce7f35a19d711fa712ae564e75cb) after the next Hibernate ORM upgrade.

DavideD:
@blafond I've just tested this with 6.3 and it doesn't seem to work. Could you double check?
I've updated the 2.1 branch to include 6.3.0.Final already

blafond:
sure

DavideD:
Superseded by https://github.com/hibernate/hibernate-reactive/pull/1746

PR #1736, opened 2023-08-29
properly parse query string during native select query

dhermanson:
This fixes a bug in `ReactiveNativeSelectQueryPlanImpl` where the sql query string was not getting processed by its `SQLQueryParser`. Before adding this code, any select query which had `{h-schema}`, `{h-domain}`, or `{h-catalog}` in the query string would throw an invalid sql grammar exception.


DavideD:
The fix looks correct, but we need to update the tests to make the build work

dhermanson:
Thank you for the quick responses. Is there anything I can do to help fix the tests? This issue is currently blocking my team from upgrading `hibernate-reactive` from `1.x` to `2.x`.

DavideD:
Never mind, it seems it was a problem with CI.
I should be able to merge this today

DavideD:
Superseded by https://github.com/hibernate/hibernate-reactive/issues/1737

Thanks!

PR #1747, opened 2023-09-01
add new methods from H6.3 to control batch/subselect fetching

gavinking:
@DavideD what is the status of support for batch fetching and subselect fetching? We do support them both now, is that correct?


DavideD:
Subselect seems to work in Hibernate Reactive 2.0.5.Final (with ORM 6.2.8.Final).
But, it doesn't after the upgrade to Hibernate ORM 6.3.0.Final.
@blafond will look into it soon

The test for batch fetching fails, but this week I'm on PTO. I will try to have look at it when I'm back if the issue is still open.

gavinking:
OK thanks, so we can put this on ice until the underlying functionality is working, I guess.

blafond:
Found the issue with reactive subselect fetch:  https://github.com/hibernate/hibernate-reactive/pull/1753

DavideD:
Batch should work now, thanks

Issue #1732, opened 2023-08-28
Postgresql 15 Array Type Support


nichts12:
error:
```
2023-08-28 18:51:12,984 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-0) HTTP Request to /request failed, error id: 1dea0f53-6add-4099-b03c-df768780ea1d-1: org.hibernate.HibernateException: Unable to extract JDBC value for position `4`
	at org.hibernate.reactive.sql.exec.spi.ReactiveValuesResultSet.lambda$readCurrentRowValues$2(ReactiveValuesResultSet.java:148)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:684)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662)
	at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2168)
	at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:144)
	at org.hibernate.reactive.sql.exec.spi.ReactiveValuesResultSet.readCurrentRowValues(ReactiveValuesResultSet.java:139)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1150)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2147)
	at io.vertx.core.Future.lambda$toCompletionStage$3(Future.java:536)
	at io.vertx.core.impl.future.FutureImpl$3.onSuccess(FutureImpl.java:141)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:60)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:211)
	at io.vertx.core.impl.future.PromiseImpl.tryComplete(PromiseImpl.java:23)
	at io.vertx.sqlclient.impl.QueryResultBuilder.tryComplete(QueryResultBuilder.java:88)
	at io.vertx.sqlclient.impl.QueryResultBuilder.tryComplete(QueryResultBuilder.java:32)
	at io.vertx.core.Promise.complete(Promise.java:66)
	at io.vertx.core.Promise.handle(Promise.java:51)
[...]

PR #1809, opened 2023-11-29
Enable Array type support

blafond:
addresses [#1732]
-   Initial tests work for PostrgreSQL and CockroachDB

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
@@ -35,6 +35,7 @@ import java.util.function.Function;
 
 import org.hibernate.engine.jdbc.BlobProxy;
 import org.hibernate.engine.jdbc.ClobProxy;
+import org.hibernate.type.descriptor.jdbc.JdbcType;
 
 import io.vertx.core.buffer.Buffer;
 import io.vertx.sqlclient.Row;
@@ -746,6 +747,15 @@ public class ResultSetAdaptor implements ResultSet {
 	@Override
 	public Array getArray(int columnIndex) {
 		throw new UnsupportedOperationException();
+	}
+
+	public Array getArray(int columnIndex, JdbcType elementJdbcType) {
+		Object[] objects = (Object[]) row.getValue( columnIndex - 1 );
+		wasNull = objects == null;
+		if ( objects == null ) {
+			return null;
+		}
+		return new ArrayAdapter( elementJdbcType, objects );
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
@@ -1,0 +1,304 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.types;
+
+import java.io.Serializable;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.Month;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Consumer;
+
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.reactive.BaseReactiveTest;
+
+import org.hibernate.reactive.testing.DBSelectionExtension;
+import org.hibernate.type.SqlTypes;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.ORACLE;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@Timeout(value = 10, timeUnit = MINUTES)
+
+public class JavaTypesArrayTest extends BaseReactiveTest {
+
+	@RegisterExtension
+	public final DBSelectionExtension skip = DBSelectionExtension.skipTestsFor( ORACLE );
+
+	@Override
+	protected Set<Class<?>> annotatedEntities() {
+		return Set.of( Basic.class );
+	}
+
+	private void testField( VertxTestContext context, Basic
+			original, Consumer<Basic> consumer) {
+		test( context, getSessionFactory()
+				.withTransaction( (s, t) -> s.persist( original ) )
+				.thenCompose( v -> getSessionFactory().withSession( s -> s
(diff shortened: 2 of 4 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -748,6 +750,72 @@ public Array getArray(int columnIndex) {
...
+				return null;
+			}
+
+			@Override
+			public Object getArray(long l, int i) throws SQLException {
+				return null;
+			}
+
+			@Override
+			public Object getArray(long l, int i, Map<String, Class<?>> map) throws SQLException {
```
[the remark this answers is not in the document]
DavideD: The PostgreSQL JDBC driver throws an exception when the map is not null: https://github.com/pgjdbc/pgjdbc/blob/6f11a2854ef67aec16dd413e2fa88bd8512eaf3d/pgjdbc/src/main/java/org/postgresql/jdbc/PgArray.java#L140

I think we can do the same (`UnsupportedOperationException` should be fine)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -748,6 +750,72 @@ public Array getArray(int columnIndex) {
...
+			}
+
+			@Override
+			public ResultSet getResultSet(long l, int i, Map<String, Class<?>> map) throws SQLException {
+				return null;
+			}
+
+			@Override
+			public void free() throws SQLException {
+
```
DavideD, line 814: I don't think it matters in our case but, if you create a separate class, you can then implement this as `stringArray = null`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
```diff
@@ -0,0 +1,319 @@
...
+		};
+		basic.localTimeArray = dataArray;
+
+		testField( context, basic, found -> assertArrayEquals( dataArray, found.localTimeArray ) );
+	}
+
+	@Test
+	public void testLocalDateTimeArrayType(VertxTestContext context) {
+		Basic basic = new Basic();
+		LocalDateTime date1 = LocalDateTime.now();
```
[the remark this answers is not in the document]
DavideD: Unused variable

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
```diff
@@ -0,0 +1,319 @@
...
+		Long[] longArray = new Long[0];
+		Float[] floatArray = new Float[0];
+		Double[] doubleArray = new Double[0];
+		UUID[] uuidArray = new UUID[0];
+		AnEnum[] enumArray;
+		@JdbcTypeCode(SqlTypes.ARRAY)
+		BigInteger[] bigIntegerArray = new BigInteger[0];
+		@JdbcTypeCode(SqlTypes.ARRAY)
+		BigDecimal[] bigDecimalArray = new BigDecimal[0];
+		@Column
```
DavideD, line 290: `@Column` is not needed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
```diff
@@ -0,0 +1,319 @@
...
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.COCKROACHDB;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.POSTGRESQL;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@Timeout(value = 10, timeUnit = MINUTES)
+
+public class JavaTypesArrayTest extends BaseReactiveTest {
+	@RegisterExtension
+	public final DBSelectionExtension skip = DBSelectionExtension.runOnlyFor( POSTGRESQL, COCKROACHDB );
```
DavideD, line 44: A note on the reason this only work for Postgres would be nice.
Don't the other databases  have an array type? is it a problem with the Vert.x client?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -748,6 +750,72 @@ public Array getArray(int columnIndex) {
...
+			}
+
+			@Override
+			public Object getArray() throws SQLException {
+				return stringArray;
+			}
+
+			@Override
+			public Object getArray(Map<String, Class<?>> map) throws SQLException {
+				return null;
```
DavideD, line 779: Unless it causes other issues, it would be better to throw an exception instead of returning `null`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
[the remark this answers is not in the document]
blafond: fixed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
blafond: Working through the changes and making progress on other DB types.  JdbcTypeCode, for some reason is less restrictive with PG. Other DB's returned different sql types, so removing/changing the annotations at the moment.  Making progress, so looks like there will be less impact for other DB's.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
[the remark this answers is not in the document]
DavideD: If you push the change, the PR conversation on GH will be updated accordingly

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -748,6 +750,72 @@ public Array getArray(int columnIndex) {
 		throw new UnsupportedOperationException();
 	}
 
+	public Array getArray(int columnIndex, JdbcType elementJdbcType) {
+		// TODO:  check object type and implement interface methods to convert
+		//
```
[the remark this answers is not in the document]
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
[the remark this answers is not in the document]
blafond: Done

hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/JavaTypesArrayTest.java
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: done

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
blafond: fixed

PR #1813, opened 2023-12-08
order support

### LINKED ISSUE #1735: Add support for `Order`
Related to https://github.com/hibernate/hibernate-reactive/issues/1706

See the new class in ORM: [org.hibernate.query.Order](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/query/Order.java)

blafond:
fixes #1735
 - implements and tests reactive session's `setOrder(...)`

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
@@ -1,0 +1,327 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+
+import org.hibernate.metamodel.model.domain.EntityDomainType;
+import org.hibernate.metamodel.model.domain.internal.MappingMetamodelImpl;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.metamodel.SingularAttribute;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.stream.Collectors.toList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.query.Order.asc;
+import static org.hibernate.query.Order.desc;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Timeout(value = 10, timeUnit = MINUTES)
+public class OrderTest extends BaseReactiveTest {
+	final Book book1 = new Book("9781932394153", "Hibernate in Action");
+	final Book book2 = new Book("9781617290459", "Java Persistence with Hibernate");
+
+	@Override
+	protected Collection<Class<?>> annotatedEntities() {
+		return List.of( Book.class );
+	}
+
+	@BeforeEach
+	public void populateDB(VertxTestContext context) {
+		test( context, getSessionFactory()
+				.withTransaction( session -> session.persist( book1, book2 ) )
+		);
+	}
+
+	@Test
+	public void testOrder(VertxTestContext context) {
+		final SingularAttribute<? super Book, ?> isbn = getIsbnAttribute();
+		final SingularAttribute<? super Book, ?> title = getTitleAttribute();
+
+		test(
+				context,
+				getSessionFactory().withSession( session -> session
+						.createSelectionQuery( "from Book", Book.class )
+						.setOrder( asc( title ) ).getResultList()
+						.thenAccept( books -> assertThat( books ).contains( book1, book2 ) )
+						.thenCompose( v -> session
+								.createSelectionQuery( "from Book", Book.class )
+								.setOrder( desc( title ) ).getResultList()
+						).thenAccept( books -> assertThat( books ).containsExactly( book2, book1 ) )
+						.thenCompose( v -> session
+								.createSelectionQuery( "from Book", Book.class )
+								.setOrder( asc( isbn ) )
+								.getResultList()
+						).thenAccept( books -> assertThat( books ).containsExactly( book2, book1 ) )
+						.thenCompose( v -> session
+								.createSelectionQuery( "from Book", Book.class )
+								.setOrder( desc( isbn ) )
+								.getResultList()
+						).thenAccept( books -> assertThat( books ).containsExactly( book1, book2 ) )
+				)
+		);
+	}
+
+	@Test
+	public void testOrderMutiny(VertxTestContext context) {
+		final SingularAttribute<? super Book, ?> isbn = getIsbnAttribute();
+		final SingularAttribute<? super Book, ?> title = getTitleAttribute();
+
+		test(
+				context,
+				getMutinySessionFactory().withSession( session -> session
+						.createSelectionQuery( "select title from Book", String.class )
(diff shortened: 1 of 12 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+		return List.of(
+				Book.class
+		);
+	}
+
+	static boolean skipDBs() {
+		// MySQL & MariaDB will fail because the simple 'String isbn' element will prevent book from being created correctly
+		// errorMessage=Table 'hreact.Book' doesn't exist, errorCode=1146
+		// NOTE: ORM's OrderTest will also fail with these DB's
+		return DatabaseConfiguration.dbType() == MYSQL ||
```
DavideD, line 51: If I run the test with MySQL it works fine. There was probably something wrong on your machine.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+	private CompletionStage<Void> populateDB() {
+		return getSessionFactory()
+				.withTransaction( session -> session.persist( book1, book2 ) );
+	}
+
+	private Uni<Void> populateDBMunity() {
+		return getMutinySessionFactory()
+				.withSession( session -> session.persistAll( book1, book2 )
+						.chain( session::flush ) );
+	}
```
DavideD, line 64: There's no reason to have two separate methods for adding entries in the database.
We are testing the order, so you can just inert the elements once ([like we do everywhere else](https://github.com/hibernate/hibernate-reactive/blob/11950394f8cf943a28e00d2cfd99c21cc58e31b4/hibernate-reactive-core/src/test/java/org/hibernate/reactive/HQLQueryTest.java#L42)).

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+	@Test
+	public void testOrder(VertxTestContext context) {
+		test(
+				context,
+				populateDB().thenCompose( v -> getSessionFactory()
+						.withTransaction( (session, tx) -> session
+								.createSelectionQuery( "select title from Book", String.class )
+								.getResultList()
+								.thenAccept( list -> assertTrue( list.size() == 2 ) )
```
DavideD, line 74: Why did you add this block?
The purpose of the test class is to check the `setOrder` method, and you are not using it here.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+											.findEntityType( Book.class );
+									SingularAttribute<? super Book, ?> title = bookType.findSingularAttribute( "title" );
+									return session
+											.createSelectionQuery( "from Book", Book.class )
+											.setOrder( Order.asc( title ) )
+											.getResultList();
+								} ).thenAccept( books -> {
+									assertEquals( "Hibernate in Action", books.get( 0 ).title );
+									assertEquals( "Java Persistence with Hibernate", books.get( 1 ).title );
+								} )
```
DavideD, line 87: with AssertJ, you can replace this assertion with:
```
assertThat( books ).containsExactly(book1, book2);
```

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+								.thenCompose( vv -> {
+									MappingMetamodelImpl metamodel = (MappingMetamodelImpl) getSessionFactory().getMetamodel();
+									EntityDomainType<Book> bookType = metamodel.getJpaMetamodel()
+											.findEntityType( Book.class );
+									SingularAttribute<? super Book, ?> title = bookType.findSingularAttribute( "title" );
```
DavideD, line 92: There's no reason to repeat these three lines everywhere: you can declare them once at the begging of the test

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
DavideD: There are other places where you can do the same

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+								} ).thenAccept( books -> {
+									assertEquals( "Hibernate in Action", books.get( 0 ).title );
+									assertEquals( "Java Persistence with Hibernate", books.get( 1 ).title );
+								} )
+						) )
+		);
+	}
+
+	@Test
+	public void testAscendingDescendingBySelectElement(VertxTestContext context) {
```
DavideD, line 132: Maybe it's better to test this with Mutiny as well

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
```diff
@@ -0,0 +1,242 @@
...
+									SingularAttribute<? super Book, ?> isbn = bookType.findSingularAttribute( "isbn" );
+									return session
+											.createSelectionQuery( "from Book", Book.class )
+											.setOrder( Order.desc( isbn ) )
+											.getResultList();
+								} ).invoke( books -> {
+									assertEquals( "Hibernate in Action", books.get( 0 ).title );
+									assertEquals( "Java Persistence with Hibernate", books.get( 1 ).title );
+								} )
+				) )
```
DavideD, line 224: If you create a single test method for each block, the whole class becomes much more readable and easier to understand (same is true for the other tests).

DavideD (changes requested):
Looks fine overall.

Maybe I've missed it, but I don't think there's a test for `setOrder(List<Order>)`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
blafond: FYI... yes.. clearing my docker containers and rebooting... all DB's work

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
DavideD: On second thought, nevermind. Ignore that

blafond:
I'll have an update in the morning to the test.  Took your suggestions and it's looking cleaner and added a couple of tests, including the List<> case.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
blafond: removed

hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
blafond: Can we mix Mutiny and Stage methods within the `test(...)` method?

DavideD: No, but you can use the `@BeforeEach` to run the inserts before running the actual tests. Like in the test class I linked in the previous comment

blafond: done

DavideD:
Looks good.
 @blafond, could you check if [these changes](https://github.com/DavideD/hibernate-reactive/commit/32ce11e1ad138c9ba22004c94a28c55a71a606d5) make sense to you?

DavideD:
Superseded by https://github.com/hibernate/hibernate-reactive/pull/1815

PR #1815, opened 2023-12-13
Support setOrder

### LINKED ISSUE #1735: Add support for `Order`

DavideD:
fix #1735 
Supersedes #1813

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
@@ -1,0 +1,343 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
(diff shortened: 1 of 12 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/OrderTest.java
blafond, line 1: this import isn't used

`import static org.junit.jupiter.api.Assertions.assertEquals;`

blafond (approved):
Just one change... the test has an unused import:

import static org.junit.jupiter.api.Assertions.assertEquals;

DavideD:
Thanks, there also something else that doesn't work. I'm looking into it

blafond:
I tested locally and there are some failed tests :(

DavideD:
Should be fixed now, let's see what CI think about it

PR #1841, opened 2024-01-19
Upgrade to Mutiny 2.5.4

jponge:


jponge:
@DavideD That's an interesting one, especially as the change was made a long time ago. I'm going to have a look.

jponge:
Confirmed, there's a race-condition here 😄

DavideD:
Thanks @jponge

PR #1828, opened 2024-01-10
Replace DBSelectionExtension with annotations

blafond:
fixes #1827

PR #1842, opened 2024-01-22
Disable or enable test for database types using annotations

### LINKED ISSUE #1827: Replace DBSelectionExtension with annotations
The current DBSelectionExtension only provides class level management of what DB's to run or skip. Some test methods have been failing in a test, but options only include using **RunOnlyFor( dbs...)** or **SkipTestsFor** for all test cases.

Currently there are some tests that use the **@Disabled** annotation, but the tests are disabled for ALL DBTypes.

Create annotations that will allow running tests that can skip or run with individual test methods as well all tests in the test class.

DavideD:
Fix #1827 
Supersedes https://github.com/hibernate/hibernate-reactive/pull/1828


blafond (approved):
I like the annotation changes.. yes, it's simpler and cleans up their tests.

I walked through the real test changes and checked that the comment messages were converted to the annotation "reason =".

DavideD:
Thanks,
merged!

PR #1868, opened 2024-02-29
Tests for @SoftDelete

### LINKED ISSUE #1787: Add support for soft deletes
Hibernate ORM 6.4 introduces soft deletes, we should check if the new feature also works in Hibernate Reactive

Hibernate ORM docs: https://docs.jboss.org/hibernate/orm/6.4/userguide/html_single/Hibernate_User_Guide.html#soft-delete

blafond:
Fixes #1787

Covers the primary annotation value options.


DavideD:
I will send a new PR

PR #1924, opened 2024-05-27
Don't open multiple connections on schema update

### LINKED ISSUE #1909: Acquisition of multiple connections for schema management
Hibernate Reactive seems to acquire several connections for schema management, which can lead to deadlocks in some extreme situations (small connection pool, multiple schema management running concurrently on the same pool).

See https://github.com/quarkusio/quarkus/issues/39930#issuecomment-2090882772

> We could fix it by avoiding the use of multiple concurrent transactions in schema update script generation in Hibernate ORM and/or Hibernate Reactive.
> For Hibernate Reactive, weirdly it seems to [acquire a connection](https://github.com/hibernate/hibernate-reactive/blob/8ca1c9f528e5cd0b8f7398e55c06e28144d238ad/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java#L77) to then [immediately ignore it and acquire a second one from the pool in selectJdbcOutsideTransaction](https://github.com/hibernate/hibernate-reactive/blob/8ca1c9f528e5cd0b8f7398e55c06e28144d238ad/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java#L109). That's probably a bug @DavideD , isn't it? At the very least I'd expect the first connection not to be acquired.

DavideD:
Fix #1909

This should prevent opening multiple connections during schema migration.
I need to figure out how to test it though (I've  only checked manually).

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
@@ -36,6 +36,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.hibernate.boot.model.relational.SqlStringGenerationContext;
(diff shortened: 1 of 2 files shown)


DavideD:
I'm not going to merge this for now, I think there's something to change related to the API that and I need to think more about it

PR #1928, opened 2024-05-29
Avoid creation of multiple connections during schema migration

### LINKED ISSUE #1909: Acquisition of multiple connections for schema management

DavideD:
Fix #1909

I'm not sure if it strictly necessary anymore to run the queries outside of the current transaction, but I think this way the code is simpler and consistent to what we were doing before. And the code is cleaner.

Basically, I moved the method we used to run the query in the pool.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
@@ -60,23 +60,6 @@ public interface ReactiveConnection {
 
 	CompletionStage<ResultSet> selectJdbc(String sql, Object[] paramValues);
 
-	/**
-	 * This method is intended to be used only for queries returning
-	 * a ResultSet that must be executed outside of any "current"
-	 * transaction (i.e with autocommit=true).
-	 * <p/>
-	 * For example, it would be appropriate to use this method when
-	 * performing queries on information_schema or system tables in
-	 * order to obtain metadata information about catalogs, schemas,
-	 * tables, etc.
-	 *
-	 * @param sql - the query to execute outside of a transaction
-	 * @param paramValues - a non-null array of parameter values
-	 *
-	 * @return the CompletionStage<ResultSet> from executing the query.
-	 */
-	CompletionStage<ResultSet> selectJdbcOutsideTransaction(String sql, Object[] paramValues);
-
 	<T> CompletionStage<T> insertAndSelectIdentifier(String sql, Object[] paramValues, Class<T> idClass, String idColumnName);
 	CompletionStage<ResultSet> insertAndSelectIdentifierAsResultSet(String sql, Object[] paramValues, Class<?> idClass, String idColumnName);
 
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ExternalSqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ExternalSqlClientPool.java
@@ -5,15 +5,26 @@
  */
 package org.hibernate.reactive.pool.impl;
 
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 
+import org.hibernate.engine.jdbc.internal.FormatStyle;
 import org.hibernate.engine.jdbc.spi.SqlExceptionHelper;
 import org.hibernate.engine.jdbc.spi.SqlStatementLogger;
+import org.hibernate.reactive.adaptor.impl.ResultSetAdaptor;
 import org.hibernate.reactive.mutiny.Mutiny;
 import org.hibernate.reactive.stage.Stage;
 import org.hibernate.reactive.util.impl.CompletionStages;
 
+import io.vertx.sqlclient.DatabaseException;
 import io.vertx.sqlclient.Pool;
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowSet;
+import io.vertx.sqlclient.Tuple;
+
+import static org.hibernate.reactive.util.impl.CompletionStages.rethrow;
 
 /**
  * A pool of reactive connections backed by a Vert.x {@link Pool}.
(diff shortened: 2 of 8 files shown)


gavinking:
> I'm not sure if it strictly necessary anymore to run the queries outside of the current transaction

WDYM?

DavideD:
I was referring to this comment: https://github.com/quarkusio/quarkus/issues/39930#issuecomment-2111727890

It made me think we don't actually need to do it. But now that I re-read it, it seems  that we need to do it but it's a bit unclear why.

Can you shed some light about this?

gavinking:
I mean there are various reasons we might want to do that.

For example, increment on a table generator should happen outside the current tx.

For DDL execution, well, IIRC that's something we've always done non-transactionally. I know there are some databases (Postgres, IIRC) which do support transactional DDL, but I doubt that this is something we need/want.

DavideD:
Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
```diff
@@ -60,23 +60,6 @@ interface Expectation {
...
-	 * order to obtain metadata information about catalogs, schemas,
-	 * tables, etc.
-	 *
-	 * @param sql - the query to execute outside of a transaction
-	 * @param paramValues - a non-null array of parameter values
-	 *
-	 * @return the CompletionStage<ResultSet> from executing the query.
-	 */
-	CompletionStage<ResultSet> selectJdbcOutsideTransaction(String sql, Object[] paramValues);
-
```
yrodiere, line 79: This looks like a breaking change in APIs... ? Should it be documented somewhere?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ExternalSqlClientPool.java
```diff
@@ -82,4 +93,41 @@ public SqlExceptionHelper getSqlExceptionHelper() {
...
+		return preparedQueryOutsideTransaction( sql, Tuple.wrap( paramValues ) )
+				.thenApply( ResultSetAdaptor::new );
+	}
+
+	public CompletionStage<RowSet<Row>> preparedQueryOutsideTransaction(String sql, Tuple parameters) {
+		feedback( sql );
+		return getPool().preparedQuery( sql ).execute( parameters ).toCompletionStage()
+				.handle( (rows, throwable) -> convertException( rows, sql, throwable ) );
+	}
+
```
yrodiere, line 109: Is there any reason for callers to call this instead of just... getting a connection from the pool and using it?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
```diff
@@ -65,54 +62,42 @@ public ReactiveImprovedExtractionContextImpl(
...
-				getJdbcEnvironment().getDialect()
-		);
+		final Parameters parametersDialectSpecific = Parameters.instance( getJdbcEnvironment().getDialect() );
 		final String queryToUse = parametersDialectSpecific.process( queryString, parametersToUse.length );
-		return connectionStage.thenCompose( c -> c.selectJdbcOutsideTransaction( queryToUse, parametersToUse ) )
+		return connectionPool
+				// DDL needs to run outside the current transaction. For example:
+				// - increment on a table-based id generator should happen outside the current tx.
+				// - not all databases support transactional DDL
+				.selectJdbcOutsideTransaction( queryToUse, parametersToUse )
```
yrodiere, line 88: This makes sense, but then:

1. I don't think this code is ever invoked within a transaction... is it?
2. Why wouldn't this code just retrieve a connection from the pool and use it directly? I.e. instead of removing the retrieval of a connection (`service.getConnection()`), you'd remove the call to `selectJdbcOutsideTransaction`, and would just use the connection to perform DDL. I think this would achieve the same results, without needing a specific method in the pool... ?

yrodiere (commented):
Thanks.

As far as I understand this would indeed solve the problem, though I do not quite understand why you chose to keep `selectJdbcOutsideTransaction` instead of simply retrieving and using the connection the way it's done everywhere else.

yrodiere:
> As far as I understand this would indeed solve the problem [...]

So maybe let's merge and improve later if necessary? Assuming the breaking change gets documented somewhere.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveConnection.java
DavideD: This API is marked as `@Incubating` and I doubt somebody else is implementing, using it. But, I guess I can mark the method as deprecated if we prefer to play it safe.

yrodiere: > This API is marked as `@Incubating`

Ah, I didn't see that. Please ignore me then :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/ExternalSqlClientPool.java
DavideD: I think that would work.

When I wrote the patch I felt it had the following benefit:
* I don't have to handle an open connection and make sure that it gets closed correctly in `ReactiveImprovedExtractionContextImpl`
* It's consistent with what we had before
* It's consistent with the Vert.x API, without exposing it. Not very important, but it helps us to use that existing method in Vert.x
* It's sending a stronger message about running in a different transaction (after sleeping on this, I don't know if I believe this any more. Still mentioning it for reference)

The only drawback I can see right now is a small amount of simple duplicated code and maybe a single method in the pool API that could be removed. If you feel strong about this or you have some concerned I haven't noticed, I can change the code.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/ReactiveImprovedExtractionContextImpl.java
DavideD: I've answered in the other comments bu, shortly:
1.  I like that it's more explicit about what's going on for future reference (I have the memory of a gold fish)
2. I think the code is simpler this way, but I'm open about changing it. See https://github.com/hibernate/hibernate-reactive/pull/1928#discussion_r1624264056

yrodiere:
Alright then, let's merge. Thanks @DavideD!

Issue #1938, opened 2024-06-11
Add jdk 24 to the build on CI


marko-bekhta:
Related to https://hibernate.zulipchat.com/#narrow/stream/132095-hibernate-infra/topic/JDK.2Ftools.20on.20CI/near/443732491

PR #1939, opened 2024-06-11
Add jdk 24 to the build on CI

marko-bekhta:
fixes https://github.com/hibernate/hibernate-reactive/issues/1938

Hey @DavideD 😃 
I thought while I was at it (adding JDK builds to other projects) I'd create one for Reactive as well 😃 

22 is now GA, so I've removed the -ea from the version


DavideD:
Thanks

Issue #1929, opened 2024-05-29
Enable reproduceable archives


DavideD:
See https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives and https://reproducible-builds.org/

PR #1981, opened 2024-08-29
Enable reproducible archives

marko-bekhta:
fixes https://github.com/hibernate/hibernate-reactive/issues/1929

Hey @DavideD  😃,
I thought while I was at it for ORM, I might as well see about Reactive....
It seems that the only thing for Reactive was to enable those two bool flags as nothing else was detected by the check. Now as for the check to make sure we don't break something along the way ... for Search/Validator and ORM I've added a check in the nightly builds, not sure about Reactive as most of the things are done with GH actions here... How would you want to handle that? (A Validator example https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator-nightly/detail/main/46/pipeline)

Also... to double check... Reactive produces only one jar (`hibernate-reactive-core`), right? Here's the result I get from the check:

```
ok=4
ko=0
okFiles="hibernate-reactive-core-2.4.1-SNAPSHOT.jar hibernate-reactive-core-2.4.1-SNAPSHOT.module hibernate-reactive-core-2.4.1-SNAPSHOT.pom hibernate-reactive-core-2.4.1-SNAPSHOT-sources.jar"
koFiles=""
```

### DIFF
--- a/ci/compare-build-results.sh
+++ b/ci/compare-build-results.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+# This is a simple script to check if builds are reproducible. The steps are:
+# 1. Build Hibernate Reactive with `./gradlew --no-daemon clean publishToMavenLocal --no-build-cache -Dmaven.repo.local=some-path/out/build1`
+# 2. Build Hibernate Reactive with `./gradlew --no-daemon clean publishToMavenLocal --no-build-cache -Dmaven.repo.local=some-path/out/build2` second time pointing to a different local maven repository to publish
+# 3. Compare the build results with sh ./ci/compare-build-results.sh some-path/out/build1 some-path/out/build2
+# 4. The generated .buildcompare file will also contain the diffscope commands to see/compare the problematic build artifacts
+
+outputDir1=$1
+outputDir2=$2
+outputDir1=${outputDir1%/}
+outputDir2=${outputDir2%/}
+
+ok=()
+okFiles=()
+ko=()
+koFiles=()
+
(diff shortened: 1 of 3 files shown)


DavideD:
Thanks. I'm a bit late on this but I'm going to have a look at it soon.
By the way, you left `ORM` in the comments, but I'm assuming it should be `Reactive

marko-bekhta:
right ... fixed the comments 🙈 sorry 
Let me know if you want the check to run on Jenkins or as GH actions ... 😃

marko-bekhta:
it's been a while 😄 I noticed this one while going through the list of my PRs 🙈 
I've added a Jenkins job similar to the one we have for Search/Validator/ORM to test the build reproducibility in a nightly build. A test run against my repo for this job is here https://ci.hibernate.org/view/Personal%20jobs/job/hibernate-reactive-personal-marko/job/build%252Freprod-test/2/console

marko-bekhta:
> Any particular reason for this to be still a draft?

Besides me not noticing it's still a draft, no, no other reason 😃 🙈 

> Hibernate Reactive also releases snapshots from CI now here: https://ci.hibernate.org/view/Reactive/job/hibernate-reactive-publish-snapshot/

nice 😃👍🏻, so adding another one to CI seems like a way to go.

DavideD:
> nice 😃👍🏻, so adding another one to CI seems like a way to go.

Are you planning on doing it? Or should I merge this now and do it later?
I'm assuming it will need this file updated: https://github.com/hibernate/hibernate-reactive/blob/main/ci/snapshot-publish.Jenkinsfile

DavideD:
Merged, thanks a lot

PR #2061, opened 2025-01-14
Use specific Jenkins nodes for releases

yrodiere:
This should be safer as these nodes are only used once.

### DIFF
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -58,7 +58,7 @@ def checkoutReleaseScripts() {
 
 pipeline {
 	agent {
-		label 'Worker&&Containers'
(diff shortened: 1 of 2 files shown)


DavideD:
Merged, thanks.

I changed the commit message to include an issue code: 8178bf82da4063c8766c30f516bd5138240c5e54

yrodiere:
Thanks.

> I changed the commit message to include an issue code: [8178bf8](https://github.com/hibernate/hibernate-reactive/commit/8178bf82da4063c8766c30f516bd5138240c5e54)

FWIW on Hibernate ORM, Search and Validator we consider issues optional for changes that only affect the build/CI.

PR #2062, opened 2025-01-14
[2.4] Use specific Jenkins nodes for releases

yrodiere:
This should be safer as these nodes are only used once.

Backport of #2061


DavideD:
Merged, thanks.

I changed the commit message to include an issue code: 91fc13658b7d75675cdd6ece6766ea0d95c5e82c

Issue #2026, opened 2024-12-09
UnexpectedAccessToTheDatabase error when updating (session.merge()) an existing entity with a ManyToMany relation, which uses a JoinTable with composite id.


BertSuffys:
**Allow me to clarify;**

My system has Users and roles. A user can have many roles, and each role can belong to multiple users. A User has a list of roles, mapped as ManyToMany, fetched eagerly. It is also using an intermediary table with a composite id. 

Consider the code below;

```
public void updateUser(RoutingContext routingContext) {
    Map<String, String> postBody = getRequestBody(routingContext);
    if (postBody != null) {
      String id = postBody.get("id");
      if (id != null) {
        userRepository.getUserById(Long.decode(id)).onComplete(cmsUserGet -> {
          if (cmsUserGet.succeeded()) {

            // language
            String urlLang = routingContext.request().getParam("lang");
            Lang lang = urlLang != null ? Lang.createLang(urlLang) : LangUtil.getConfigDefaultLang();
            CmsUser cmsUser = cmsUserGet.result();
            cmsUser.setLang(lang);

            this.roleRepository.getAllCmsRoles().onComplete(roles -> {

              // Retrieve fields
              String firstname = postBody.get("firstname");
              String lastname = postBody.get("lastname");
              String email = postBody.get("email");
              String title = postBody.get("title");
              String password = AuthenticationUtil.bCrypt(postBody.get("password"));
              List<Long> roleIds = Arrays.stream(postBody.get("roles").split(",")).map(Long::parseLong).toList();
              List<CmsRole> selectedRoles = roles.result().stream().filter(obj -> roleIds.contains(obj.getId())).toList();

              // Update fields
              cmsUser.getRoles().clear();
              cmsUser.getRoles().addAll(selectedRoles);
              cmsUser.setFirstname(firstname);
              cmsUser.setLastname(lastname);
              cmsUser.setEmail(email);
              cmsUser.setTitle(title);
              if(password != null){
                cmsUser.setPassword(password);
              }

              // Update
              userRepository.updateUser(cmsUser).onComplete(cmsUserUpdate -> {
                if (cmsUserUpdate.succeeded()) {
                  ResponseUtil.ok(routingContext, new CmsUserDTO(cmsUser, lang).serializeJson().encode());
                } else {
                  ResponseUtil.internal_server_error(routingContext, "Something went wrong updating the CMSRole");
                }
              });
            });
          } else {
[...]

PR #2065, opened 2025-01-17
UnexpectedAccessToTheDatabase error when merging a detached entity with a ToMany association

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/2026
Related Jira https://issues.redhat.com/projects/HREACT/issues/HREACT-42

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
@@ -1,0 +1,516 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.engine.impl;
+
+import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.Hibernate;
+import org.hibernate.HibernateException;
+import org.hibernate.collection.spi.AbstractPersistentCollection;
+import org.hibernate.collection.spi.PersistentArrayHolder;
+import org.hibernate.collection.spi.PersistentCollection;
+import org.hibernate.engine.spi.CollectionEntry;
+import org.hibernate.engine.spi.PersistenceContext;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.metamodel.mapping.PluralAttributeMapping;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.reactive.logging.impl.Log;
+import org.hibernate.reactive.logging.impl.LoggerFactory;
+import org.hibernate.reactive.util.impl.CompletionStages;
+import org.hibernate.type.ArrayType;
+import org.hibernate.type.CollectionType;
+import org.hibernate.type.CustomCollectionType;
+import org.hibernate.type.EntityType;
+import org.hibernate.type.ForeignKeyDirection;
+import org.hibernate.type.MapType;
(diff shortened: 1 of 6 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/OneToManyArrayMergeTest.java
```diff
@@ -0,0 +1,192 @@
...
+						)
+						.chain( user -> {
+									assertThat( user.getFirstname() ).isEqualTo( UPDATED_FIRSTNAME );
+									assertThat( user.getLastname() ).isEqualTo( UPDATED_LASTNAME );
+									assertThat( user.getRoles() ).hasSize( 2 );
+									return getMutinySessionFactory()
+											.withTransaction( s -> s.merge( user ) );
+								}
+						)
+						.onFailure().invoke( throwable -> fail( throwable ) )
```
DavideD, line 76: Unless I'm missing something, there's no need to call `.onFailure` in this test

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
```diff
@@ -0,0 +1,480 @@
...
+			return replaceOriginal( type, original, target, session, owner, copyCache );
+		}
+	}
+
+	// todo: make org.hibernate.type.CollectionType#replaceNullOriginal public ?
+	private static CompletionStage<Object> replaceNullOriginal(
+			Object target,
+			SessionImplementor session) {
+		if ( target == null ) {
+			return null;
```
DavideD, line 88: Method returning `CompletionStage` should never return null in principle. You can return `CompletionStages.nullFuture()`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
```diff
@@ -0,0 +1,481 @@
...
+
+		final Type elemType = type.getElementType( session.getFactory() );
+		return loop(
+				0, length, i -> {
+					return getReplace( elemType, Array.get( original, i ), owner, session, copyCache )
+							.thenCompose( o -> {
+											  Array.set( result, i, o );
+											  return completedFuture( result );
+										  }
+							);
```
DavideD, line 329: you can rewrite this block as:
```java
.thenApply( o -> {
    Array.set( result, i, o );
    return result;
});
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
```diff
@@ -0,0 +1,481 @@
...
+			PersistentCollection<?> original,
+			PersistentCollection<?> result,
+			Type elemType,
+			Object owner,
+			Map<Object, Object> copyCache,
+			SessionImplementor session) {
+		final CollectionEntry ce = session.getPersistenceContextInternal().getCollectionEntry( result );
+		if ( ce != null ) {
+			return createSnapshot( original, result, elemType, owner, copyCache, session )
+					.thenCompose( CompletionStages::completedFuture )
```
DavideD, line 369: I think this line can be removed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/CollectionTypes.java
```diff
@@ -0,0 +1,481 @@
...
+			Map<Object, Object> copyCache,
+			SessionImplementor session) {
+		return loop(
+				0, array.length,
+				i ->
+						getReplace( elemType, array[i], owner, session, copyCache )
+								.thenCompose( o -> {
+									array[i] = o;
+									return voidFuture();
+								} ).thenApply( unused -> voidFuture() )
```
DavideD, line 412: You shouldn't return `.voidFuture()` when using `.thenApply`
In any case, you can replace this
```java
.thenCompose( o -> {
    array[i] = o;
    return voidFuture();
} ).thenApply( unused -> voidFuture() )
```
with
```java
.thenCompose( o -> {
    array[i] = o;
} )
```

DavideD (commented):
Thanks a lot,
I left some comments about a couple of small things that I notice after a quick scan.

I will have a better look on Monday

DavideD:
Thanks a lot!
Superseded by https://github.com/hibernate/hibernate-reactive/pull/2066

Issue #2060, opened 2025-01-08
ClassCastException when using embeddable ids


DavideD:
Issue found in Quarkus: https://github.com/quarkusio/quarkus/issues/43326#issuecomment-2568096080

Adapted test case draft for Hibernate Reactive: https://github.com/DavideD/hibernate-reactive/commit/b46d379a3c3f5b7084a077bbd46574ca18561476

PR #2067, opened 2025-01-20
ClassCastException when using embeddable ids

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/2060
Related Jira https://issues.redhat.com/browse/HREACT-65


DavideD:
In my original test, I've used native queries to populate the database because it was a quick draft to provide a test case.
It seems that the approach I've used doesn't work in Oracle, though: https://github.com/hibernate/hibernate-reactive/actions/runs/12869137204/job/35877269478#step:8:1496

Could you update the test case so that it works with all the databases? Ideally without native queries.

Issue #1904, opened 2024-05-02
PropertyAccessException when creating a new object with a one-to-one association


DavideD:
Original quarkus issue: https://github.com/quarkusio/quarkus/issues/40374

PR #2081, opened 2025-01-22
PropertyAccessException when creating a new object with a one-to-one association

dreab8:
Add test showing https://github.com/hibernate/hibernate-reactive/issues/1904 has been resolved on main

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OneToOneGeneratedIdWithMapsIdTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/OneToOneGeneratedIdWithMapsIdTest.java
@@ -1,0 +1,153 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
```diff
@@ -182,20 +182,21 @@ private CompletionStage<Object> generateId(
...
-								persister.getIdentifierType().toLoggableString( generatedId, source.getFactory() ),
-								generator.getClass().getName()
-						);
-					}
-					return completedFuture( generatedId );
-				} );
+				.thenApply( id -> {
+								final Object generatedId = castToIdentifierType( id, persister );
+								if ( generatedId == null ) {
+									return failedFuture( new IdentifierGenerationException( "null id generated for: " + entity.getClass() ) );
```
DavideD, line 188: Now that this is a `.thenApply`, you have to throw the exception like in the original code. Otherwise, the error will be ignored

DavideD (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
DavideD: Same here: https://github.com/hibernate/hibernate-reactive/pull/2082

DavideD (approved):

PR #2082, opened 2025-01-22
PropertyAccessException when creating a new object with a one-to-one association

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/1904

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
@@ -144,55 +144,59 @@ abstract class AbstractReactiveSaveEventListener<C> implements CallbackRegistryC
 			// the entity instance, so it will be available
 			// to the entity in the @PrePersist callback
 			if ( generator instanceof ReactiveIdentifierGenerator ) {
-				return ( (ReactiveIdentifierGenerator<?>) generator )
-						.generate( ( ReactiveConnectionSupplier ) source, entity )
-						.thenApply( id -> castToIdentifierType( id, persister ) )
-						.thenCompose( gid -> performSaveWithId(
-								entity,
-								context,
-								source,
-								persister,
-								generator,
-								gid,
-								requiresImmediateIdAccess,
-								false
-						) );
+				return generateId( entity, source, (ReactiveIdentifierGenerator<?>) generator, persister )
+						.thenCompose( gid -> {
+							if ( gid == SHORT_CIRCUIT_INDICATOR ) {
+								source.getIdentifier( entity );
+								return voidFuture();
+							}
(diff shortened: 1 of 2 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
```diff
@@ -420,7 +420,7 @@ private CompletionStage<AbstractEntityInsertAction> addInsertAction(
 			boolean useIdentityColumn,
 			EventSource source,
 			boolean shouldDelayIdentityInserts) {
-		final ReactiveActionQueue actionQueue = source.unwrap( ReactiveSession.class ).getReactiveActionQueue();
+		final ReactiveActionQueue actionQueue = source.unwrap(ReactiveSession.class).getReactiveActionQueue();
```
DavideD, line 423: Not critical, but are you using the right formatter? Doesn't the ORM formatter keep the spaces in this case?

dreab8: comparing the code with main I noticed there were no spaces so I thought it was right and removed it, going to add them back

hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/AbstractReactiveSaveEventListener.java
```diff
@@ -144,55 +144,58 @@ else if ( generator instanceof Assigned ) {
...
-				!source.isTransactionInProgress()
-						&& !requiresImmediateIdAccess
-						&& generatedOnExecution;
-		return reactivePerformSave( entity, generatedId, persister, false, context, source, delayIdentityInserts );
+			ReactiveIdentifierGenerator<?> generator,
+			EntityPersister persister) {
+		return generator
+				.generate( (ReactiveConnectionSupplier) source, entity )
+				.thenApply( id -> castToIdentifierType( id, persister ) )
+				.thenCompose( generatedId -> {
```
DavideD, line 186: The code works, but it doesn't seem like you have to use  `.thenCompose`.
You could keep everything in the `.thenApply`, this way we avoid the creation of a couple of `CompletionStage` objects:
```java
.thenApply( id -> {
	Object generatedId = castToIdentifierType( id, persister );
	if ( generatedId == null ) {
		throw new IdentifierGenerationException( "null id generated for: " + entity.getClass() );
	}
	if ( LOG.isDebugEnabled() ) {
		LOG.debugf(
				"Generated identifier: %s, using strategy: %s",
				persister.getIdentifierType().toLoggableString( generatedId, source.getFactory() ),
				generator.getClass().getName()
		);
	}
	return generatedId;
} );
```

DavideD (commented):
Thanks a lot, looks good.

I have added a small suggestion.

Issue #2097, opened 2025-02-05
Use SHA in github actions workflow


marko-bekhta:
To match what we are doing in other projects
See https://issues.redhat.com/browse/HIB-123

PR #2098, opened 2025-02-05
Use SHA instead of versions in GH actions / Add dependabot config to update GH actions

marko-bekhta:
fixes https://github.com/hibernate/hibernate-reactive/issues/2097


DavideD:
Merged, thanks

PR #2100, opened 2025-02-05
[2.4] Use SHA instead of versions in GH actions 

DavideD:
Backport #2097 for 2.4
PR:  https://github.com/hibernate/hibernate-reactive/pull/2098

@marko-bekhta, could you double check that I didn't do anything stupid, please?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,7 +81,7 @@ jobs:
           - 5432:5432
     steps:
       - name: Checkout ${{ inputs.branch }}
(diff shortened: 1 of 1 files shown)


marko-bekhta:
I think dependabot only cares about the config in the default branch, so we could skip adding it to other branches. (https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#target-branch-)

DavideD:
OK, thanks, I will remove the dependabot commit and merge the PR

Issue #1931, opened 2024-06-05
Implement InformationExtractorJdbcDatabaseMetaDataImpl#processCrossreferenceResultSet


DavideD:
This is a new method in the latest Hibernate ORM 6.5 snapshot.
This method has been added as fix for [HHH-18221](https://hibernate.atlassian.net/browse/HHH-18221)

[HHH-18221]: https://hibernate.atlassian.net/browse/HHH-18221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DavideD:
I don't think we need to fix this because none of the dialect we support require this method.
We can create a new issue if we need it in the future

PR #2101, opened 2025-02-05
[#1931] Implement InformationExtractorJdbcDatabaseMetaDataImpl#processCrossreferenceResultSet

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/1931

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
@@ -165,10 +165,16 @@ public abstract class AbstractReactiveInformationSchemaBasedExtractorImpl extend
 			String foreignCatalog,
 			String foreignSchema,
 			String foreignTable,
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
```diff
@@ -165,10 +165,16 @@ protected <T> T processCrossReferenceResultSet(
...
+			throws SQLException {
 		// This method has been added as fix for https://hibernate.atlassian.net/browse/HHH-18221
 		// The issue is only for Informix that we don't currently support.
-		throw LOG.notYetImplemented();
+		try ( ResultSet resultSet =
+					  getExtractionContext()
+							  .getJdbcDatabaseMetaData()
+							  .getCrossReference( parentCatalog, parentSchema, parentTable, foreignCatalog, foreignSchema, foreignTable) ) {
+			return processor.process( resultSet );
+		}
```
DavideD, line 177: I don't know if this solution is correct. It seems it's using something related to JDBC.
This method is only needed for Informix that we don't support at the moment.

I think we can just close this issue. and let the code as is.
Unless there's a way with the current dialects to call it.

Issue #2055, opened 2024-12-23
Setting hibernate.query.mutation_strategy.global_temporary.create_tables to false has no effect


fdittz:
Hello,

We are using Micronaut 4.7.2, Hibernate Reactive 2.4.2.Final, and Oracle as our database. Since we are working with legacy data, our model entities utilize inheritance, specifically `InheritanceType.JOINED`.

I noticed that when we start our application, it attempts to create global temporary tables regardless of the `hibernate.query.mutation_strategy.global_temporary.create_tables` setting. This issue does not occur if we disable Hibernate Reactive by setting `jpa.default.reactive` to `false`.

While debugging the startup process, I came across the `org.hibernate.reactive.query.sqm.mutation.internal.temptable.ReactiveGlobalTemporaryTableStrategy` class. In its `prepare` method, the following code caught my attention:

```java
default void prepare(MappingModelCreationProcess mappingModelCreationProcess, JdbcConnectionAccess connectionAccess, CompletableFuture<Void> tableCreatedStage) {
    if (isPrepared()) {
        return;
    }

    setPrepared(true);

    final ConfigurationService configService = mappingModelCreationProcess
            .getCreationContext().getBootstrapContext()
            .getServiceRegistry().getService(ConfigurationService.class);
    boolean createIdTables = configService
            .getSetting(GlobalTemporaryTableStrategy.CREATE_ID_TABLES, StandardConverters.BOOLEAN, true);

    if (!createIdTables) {
        tableCreatedStage.complete(null);
    }

    LOG.debugf("Creating global-temp ID table: %s", getTemporaryTable().getTableExpression());

    connectionStage()
            .thenCompose(this::createTable)
            .whenComplete((connection, throwable) -> releaseConnection(connection)
                    .thenAccept(v -> {
                        if (throwable == null) {
                            setDropIdTables(configService);
                            tableCreatedStage.complete(null);
                        } else {
                            tableCreatedStage.completeExceptionally(throwable);
                        }
                    })
            );
}
```

I believe the issue lies in lines 65-67:

```java
if (!createIdTables) {
    tableCreatedStage.complete(null);
}
```

Although the setting is read correctly, the method does not stop execution at this point. It proceeds to create the tables in the subsequent lines.

To address this, I overwrote this class in my project and added a `return` statement inside this conditional block. After making this change, the tables were no longer created. However, I am uncertain if this is the best approach and suspect this behavior might be a bug.

Thank you.

PR #2103, opened 2025-02-06
Setting hibernate.query.mutation_strategy.global_temporary.create_tables to false has no effect

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/2055

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java
@@ -27,8 +27,6 @@ import static org.hibernate.reactive.util.impl.CompletionStages.voidFuture;
 
 public interface ReactiveGlobalTemporaryTableStrategy {
 
-	Log LOG = make( Log.class, lookup() );
-
 	static String sessionIdentifier(SharedSessionContractImplementor session) {
 		return session.getSessionIdentifier().toString();
 	}
@@ -65,22 +63,23 @@ public interface ReactiveGlobalTemporaryTableStrategy {
 		if ( !createIdTables ) {
 			tableCreatedStage.complete( null );
 		}
+		else {
+			make( Log.class, lookup() ).debugf( "Creating global-temp ID table : %s", getTemporaryTable().getTableExpression() );
 
-		LOG.debugf( "Creating global-temp ID table : %s", getTemporaryTable().getTableExpression() );
-
-		connectionStage()
-				.thenCompose( this::createTable )
-				.whenComplete( (connection, throwable) -> releaseConnection( connection )
-						.thenAccept( v -> {
-							if ( throwable == null ) {
-								setDropIdTables( configService );
-								tableCreatedStage.complete( null );
-							}
-							else {
-								tableCreatedStage.completeExceptionally( throwable );
-							}
-						} )
-				);
+			connectionStage()
+					.thenCompose( this::createTable )
+					.whenComplete( (connection, throwable) -> releaseConnection( connection )
+							.thenAccept( v -> {
+								if ( throwable == null ) {
+									setDropIdTables( configService );
+									tableCreatedStage.complete( null );
+								}
+								else {
+									tableCreatedStage.completeExceptionally( throwable );
+								}
+							} )
+					);
+		}
 	}
 
 	private CompletionStage<Void> releaseConnection(ReactiveConnection connection) {
@@ -103,7 +102,7 @@ public interface ReactiveGlobalTemporaryTableStrategy {
 
 	private static void logConnectionClosedError(Throwable t) {
 		if ( t != null ) {
-			LOG.debugf( "Ignoring error closing the connection: %s", t.getMessage() );
+			make( Log.class, lookup() ).debugf( "Ignoring error closing the connection: %s", t.getMessage() );
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/GlobalTemporaryTableTest.java
```diff
@@ -0,0 +1,125 @@
...
+							bar.name = "Noble";
+							bar.name2 = "Experiment";
+
+							return session.persist( bar );
+						} )
+				.thenAccept( v -> {
+					List<String> loggedQueries = sqlTracker.getLoggedQueries();
+					assertThat( loggedQueries ).isNotEmpty();
+					loggedQueries.forEach( q -> {
+						assertThat( q ).doesNotContain( TemporaryTable.ID_TABLE_PREFIX ).as("");
```
DavideD, line 92: `.as("")` doesn't do anything in this case, I think you can remove it.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/GlobalTemporaryTableTest.java
```diff
@@ -0,0 +1,125 @@
...
+		// Skip delete step.
+		// We don't insert any value, so there's nothing to delete
+		// Avoid having extra stuff in the log that's not relevant to the test
+		return voidFuture();
+	}
+
+	@Override
+	protected void setProperties(Configuration configuration) {
+		super.setProperties( configuration );
+		configuration.setProperty( CREATE_ID_TABLES, "false" );
```
DavideD, line 56: I think you also need to test that when the property is set to `true`, it creates the temporary tables as expected (on the databases that supports it).

You can see an example of how we test different session factory configurations in [SchemaUpdateTest](https://github.com/hibernate/hibernate-reactive/blob/42d5c0479d6bab0a9a49a360f3999fe82d39328a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/SchemaUpdateTest.java#L44)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/GlobalTemporaryTableTest.java
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	private static boolean filter(String s) {
+		String[] accepted = { "create " };
+		for ( String valid : accepted ) {
+			if ( s.toLowerCase().startsWith( valid ) ) {
+				return true;
+			}
+		}
+		return false;
```
DavideD, line 107: This could just be:
```java
	private static boolean filter(String s) {
		return s.toLowerCase().startsWith( "create " );
	}

```
(I usually also include the space after in the check)

DavideD: Or:
```java
sqlTracker = new SqlStatementTracker( s -> s.toLowerCase().startsWith( "create " ), configuration.getProperties() );
```

hibernate-reactive-core/src/test/java/org/hibernate/reactive/schema/GlobalTemporaryTableTest.java
DavideD: While reviewing the code, I noticed we also have a property `hibernate.query.mutation_strategy.global_temporary.drop_tables`, could we also test this, please?
See `GlobalTemporaryTableStrategy.DROP_ID_TABLES`

DavideD (changes requested):
It seems this test fails for CockroachDB, probably because it always uses temporary tables (just a guess, I haven't actually checked).

I also noticed that we make the same mistake [when we need to drop temporary tables](https://github.com/hibernate/hibernate-reactive/blob/42d5c0479d6bab0a9a49a360f3999fe82d39328a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java#L151).  Could you fix this as well, please?
[There is also a typo](https://github.com/hibernate/hibernate-reactive/blob/42d5c0479d6bab0a9a49a360f3999fe82d39328a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java#L154).

Thanks!

DavideD:
@dreab8, could you also remove this log declaration from the interface, please? https://github.com/hibernate/hibernate-reactive/blob/42d5c0479d6bab0a9a49a360f3999fe82d39328a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java#L30

When it's used, you can replace it with:
```java
make( Log.class, lookup() ).debugf( "Creating global-temp ID table : %s", getTemporaryTable().getTableExpression() );
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sqm/mutation/internal/temptable/ReactiveGlobalTemporaryTableStrategy.java
```diff
@@ -65,22 +65,23 @@ default void prepare(MappingModelCreationProcess mappingModelCreationProcess, Jd
...
+								if ( throwable == null ) {
+									setDropIdTables( configService );
+									tableCreatedStage.complete( null );
+								}
+								else {
+									tableCreatedStage.completeExceptionally( throwable );
+								}
+							} )
+					);
+		}
```
DavideD, line 84: This solution is fine, but we can keep the changes smaller and the code more similar to ORM if we add the `return` in the if-block: 
```
if ( !createIdTables ) {
   tableCreatedStage.complete( null );
   return;
}
```

It also reduces the indentation.

DavideD: Sure... I'm fine with the original solution

DavideD:
@dreab8, I refactored the test class, could you have a look and let me know what you think? Thanks
The new PR is here: https://github.com/hibernate/hibernate-reactive/pull/2109

PR #2109, opened 2025-02-13
Fix for creation and dropping options of ids temporary tables

### LINKED ISSUE #2055: Setting hibernate.query.mutation_strategy.global_temporary.create_tables to false has no effect

DavideD:
Fix #2055 
Supersedes #2103

I refactored the test class and [removed a leftover](https://github.com/hibernate/hibernate-reactive/pull/2103/files#diff-af65b7392ff92aec713d7287985477eaf9d77f8ed2fdecc81df8f566b5aa0c09R194)


dreab8 (approved):
LGTM, thanks for the review and the improvement :+1:

DavideD:
Merged, thanks!

PR #2116, opened 2025-02-13
Remove warnings

DavideD:
Solves a miscellaneous list of warnings from IDEA.
I'm not sure it's worth it to have an issue for this


hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/AbstractReactiveInformationSchemaBasedExtractorImpl.java
```diff
@@ -192,7 +192,7 @@ protected boolean appendClauseAndParameterIfNotNullOrEmpty(
 			StringBuilder sb,
 			List<Object> parameters) {
 
-		if ( parameter != null && ( ! String.class.isInstance( parameter ) || ! ( (String) parameter ).isEmpty() ) ) {
+		if ( parameter != null && ( !( parameter instanceof String ) || !( (String) parameter ).isEmpty() ) ) {
```
gavinking, line 195:
```suggestion
		if ( parameter != null && ( !( parameter instanceof String string ) || !string.isEmpty() ) ) {
```

DavideD:
Merged

PR #2117, opened 2025-02-13
Reactive find with lock in Quarkus with reactive hibernate

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/1905

PR #2121, opened 2025-02-17
Fix exception when using find with lock

### LINKED ISSUE #1905: Reactive find with lock in Quarkus with reactive hibernate - Issue reproduced
Hi, this is related to this issue https://github.com/hibernate/hibernate-reactive/issues/1896.

> Hi, when using Quarkus with reactive hibernate and trying to run this findById(id, LockModeType.PESSIMISTIC_WRITE), I get following Error: JPA error: java.lang.UnsupportedOperationException
The error seems to be coming from here:
[hibernate-reactive/hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/DefaultReactiveLockEventListener.java](https://github.com/hibernate/hibernate-reactive/blob/5cb015d24703a493d9d745f7df0aa4d362a19b28/hibernate-reactive-core/src/main/java/org/hibernate/reactive/event/impl/DefaultReactiveLockEventListener.java#L238-L241)
Lines 238 to 241 in [5cb015d](https://github.com/hibernate/hibernate-reactive/commit/5cb015d24703a493d9d745f7df0aa4d362a19b28)
 @Override 
 public void onLock(LockEvent event) throws HibernateException { 
 	throw new UnsupportedOperationException(); 
 } 
Is this intended, or is the implementation not finished? I can see that the last time this was edited was 4 years ago.

I was able to reproduce the issue when using a `OneToMany` and `ManyToOne` relationship.
By doing this, hibernate is forced to do a `JOIN` and the `SELECT FOR UDPATE` cannot be done in the same request.
When hibernate is trying to do the followup Lock, it throws an error.

The error should already show up after creating a `Fruit` and fetching it.
No `Owner` object is needed.
It's only there to trigger the issue.

This is the repository with the reproduced issue: https://github.com/alexjaravete/quarkus-quickstarts/tree/findById/hibernate-reactive-panache-quickstart

DavideD:
Fix #1905 
Supersedes https://github.com/hibernate/hibernate-reactive/pull/2117

@dreab8, I applied the changes we discussed, could yuo double check I didn't miss anything, please?

Thanks

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/engine/internal/ReactivePersistenceContextAdapter.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/engine/internal/ReactivePersistenceContextAdapter.java
@@ -7,6 +7,7 @@ package org.hibernate.engine.internal;
 
 import java.io.Serializable;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletionStage;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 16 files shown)


dreab8 (approved):
LGTM

PR #2123, opened 2025-02-17
Hardening transaction state

### LINKED ISSUE #1514: Need to abort/rollback transactions on error
It would seem that if I introduce assertions to check that the current transaction is terminated (in any way) before we close the connection, there's at least one test failing:
"ReactiveStatelessProxyUpdateTest. testLazyInitializationException"

We need to investigate this to ensure such resources don't leak under any circumstances.

Providing a test as a PR soon

DavideD:
Fix #1514

@Sanne, In case you want to see what the error was


Sanne:
Nice! Thanks

PR #2124, opened 2025-02-17
fix up weird signature of ReactiveEntityPersister.insertReactive()

gavinking:


DavideD:
I think you can just remove the method without deprecating.
The interface is internal, it exists only for us to be able to extend the ORM implementation and add reactive methods.

gavinking:
OK

Issue #2108, opened 2025-02-12
StatelessSession insertAll in batch does not do batching


s4iko:
I tried the method `insertAll(int batchSize, Object... entities)` but it does not do batching to insert the entities, they are all inserted using a separated query.
I didn't find any documentation about insertAll.
Do I need to add some configuration to enable batching with StatelessSession ?
Do the dialect limit the use of the StatelessSession (mine is PostgreSQLDialect included in Hibernate core 6.6.6) ?

Thanks for your help :)

Quarkus: 3.18.2
Hibernate reactive: 2.4.4

gavinking:
> but it does not do batching to insert the entities, they are all inserted using a separated query.

How did you make this determination?

Impossible to help you when you have not attached any sort of reproducer.

DavideD:
I've created a quick test and it actually seems something is not right: https://github.com/DavideD/hibernate-reactive/commit/d6a35be8f7f398a500f5800b78d25703c8a448fa

DavideD:
The vert.x SQL client expect a query and a list of parameters, we log before calling the client, so checking the log should work for testing. The batch execution happens here: https://github.com/hibernate/hibernate-reactive/blob/e1aff8f360d17dc90d077422f5e2739c233a646d/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java#L258 

I can also see that it works for the regular session, and that the batch options is skipped for some reason using the debugger.

gavinking:
OK, fine.

gavinking:
OK, @DavideD I believe I found the bug and have a fix.

gavinking:
> OK, [@DavideD](https://github.com/DavideD) I believe I found the bug and have a fix.

Or maybe not. Haha.

DavideD:
> Or maybe not. Haha.

:) It happens to me all the time

PR #2132, opened 2025-02-20
StatelessSession insertAll in batch does not do batching

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/2108

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
@@ -150,6 +150,9 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl implement
 			PersistenceContext persistenceContext) {
 		super( factory, options );
 		this.persistenceContext = persistenceContext;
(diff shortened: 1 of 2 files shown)


gavinking (approved):

dreab8:
Unfortunately with this change it works but only in conjunction with `hibernate.jdbc.batch_size` > 1.
I'm working on a change that should allow `insertAll(int batchSize, Object... entities)` to respect the `batchSize` even when  `hibernate.jdbc.batch_size` is not set.

gavinking:
@dreab8 you see how I did it in ORM, right?

```java
	@Override
	public void insertMultiple(List<?> entities) {
		final Integer batchSize = getJdbcBatchSize();
		setJdbcBatchSize( entities.size() );
		try {
			for ( Object entity : entities ) {
				insert( null, entity );
			}
		}
		finally {
			setJdbcBatchSize( batchSize );
		}
	}
```

dreab8:
In reactive we need also to use a  `BatchingConnection` and at the moment if `batcSize <2 ` we use the original connection also for the `batchingHelperSession`

dreab8:
@gavinking https://github.com/hibernate/hibernate-reactive/pull/2136

dreab8:
superseeded by https://github.com/hibernate/hibernate-reactive/pull/2136

PR #2136, opened 2025-02-26
StatelessSession insertAll in batch does not do batching

dreab8:
Fix https://github.com/hibernate/hibernate-reactive/issues/2108

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
@@ -138,7 +138,7 @@ public class MutinyStatelessSessionImpl implements Mutiny.StatelessSession {
 
 	@Override
 	public Uni<Void> insertAll(Object... entities) {
-		return uni( () -> delegate.reactiveInsertAll( entities ) );
+		return uni( () -> delegate.reactiveInsertAll( entities.length, entities ) );
 	}
 
 	@Override
@@ -158,12 +158,12 @@ public class MutinyStatelessSessionImpl implements Mutiny.StatelessSession {
 
 	@Override
 	public Uni<Void> deleteAll(Object... entities) {
-		return uni( () -> delegate.reactiveDeleteAll( entities ) );
+		return uni( () -> delegate.reactiveDeleteAll( entities.length, entities ) );
 	}
 
 	@Override
 	public Uni<Void> deleteAll(int batchSize, Object... entities) {
-		return uni( () -> delegate.reactiveDeleteAll( entities ) );
+		return uni( () -> delegate.reactiveDeleteAll( batchSize, entities ) );
 	}
 
 	@Override
@@ -178,7 +178,7 @@ public class MutinyStatelessSessionImpl implements Mutiny.StatelessSession {
 
 	@Override
 	public Uni<Void> updateAll(Object... entities) {
-		return uni( () -> delegate.reactiveUpdateAll( entities ) );
+		return uni( () -> delegate.reactiveUpdateAll( entities.length, entities ) );
 	}
 
 	@Override
@@ -208,7 +208,7 @@ public class MutinyStatelessSessionImpl implements Mutiny.StatelessSession {
 
 	@Override
 	public Uni<Void> refreshAll(Object... entities) {
-		return uni( () -> delegate.reactiveRefreshAll( entities ) );
+		return uni( () -> delegate.reactiveRefreshAll( entities.length, entities ) );
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
@@ -128,7 +128,7 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl implement
 
 	private final ReactiveConnection reactiveConnection;
 
-	private final ReactiveStatelessSession batchingHelperSession;
+	private final ReactiveStatelessSessionImpl batchingHelperSession;
 
 	private final PersistenceContext persistenceContext;
 
@@ -150,10 +150,9 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl implement
 			PersistenceContext persistenceContext) {
 		super( factory, options );
 		this.persistenceContext = persistenceContext;
-		Integer batchSize = getConfiguredJdbcBatchSize();
-		reactiveConnection = batchSize == null || batchSize < 2
-				? connection
-				: new BatchingConnection( connection, batchSize );
+		// Setting batch size to 0 because `StatelessSession` does not consider
+		// the value of `hibernate.jdbc.batch_size`
+		reactiveConnection = new BatchingConnection( connection, 0 );
 		batchingHelperSession = this;
 		influencers = new LoadQueryInfluencers( factory );
 	}
@@ -551,9 +550,12 @@ public class ReactiveStatelessSessionImpl extends StatelessSessionImpl implement
 
 	@Override
 	public CompletionStage<Void> reactiveInsertAll(int batchSize, Object... entities) {
+		final Integer jdbcBatchSize = batchingHelperSession.getJdbcBatchSize();
+		batchingHelperSession.setJdbcBatchSize( batchSize );
 		final ReactiveConnection connection = batchingConnection( batchSize );
 		return loop( entities, batchingHelperSession::reactiveInsert )
-				.thenCompose( v -> connection.executeBatch() );
+				.thenCompose( v -> connection.executeBatch() )
+				.whenComplete( (v, throwable) -> batchingHelperSession.setJdbcBatchSize( jdbcBatchSize ) );
 	}
 
 	@Override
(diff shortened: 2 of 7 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/PersistAllWithNoBatchSizeConfiguredTest.java
```diff
@@ -0,0 +1,149 @@
...
+		final GuineaPig[] pigs = {
+				new GuineaPig( 11, "One" ),
+				new GuineaPig( 22, "Two" ),
+				new GuineaPig( 33, "Three" ),
+				new GuineaPig( 44, "Four" ),
+				new GuineaPig( 55, "Five" ),
+				new GuineaPig( 66, "Six" ),
+		};
+		test( context, getMutinySessionFactory()
+				.withStatelessTransaction( s -> s.insertAll( 10, pigs ) )
```
DavideD, line 81: If it doesn't take too long, I think it would be helpful to test also test the other methods that accept the batch size as parameter: `deleteAll`, `updateAll`, and `refreshAll`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -551,9 +554,16 @@ public CompletionStage<Void> reactiveInsertAll(Object... entities) {
...
-		return loop( entities, batchingHelperSession::reactiveInsert )
-				.thenCompose( v -> connection.executeBatch() );
+		final Integer jdbcBatchSize = batchingHelperSession.getJdbcBatchSize();
+		batchingHelperSession.setJdbcBatchSize( batchSize );
+		try {
+			final ReactiveConnection connection = batchingConnection( batchSize );
+			return loop( entities, batchingHelperSession::reactiveInsert )
+					.thenCompose( v -> connection.executeBatch() );
+		}
+		finally {
```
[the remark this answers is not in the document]
DavideD: the same fix is required in `reactiveUpdateAll`, `reactiveDeleteAll`, and `reactiveRefreshAll`

hibernate-reactive-core/src/test/java/org/hibernate/PersistAllWithNoBatchSizeConfiguredTest.java
DavideD: Being a public API, I think we should also test with the `Stage` API.

DavideD (changes requested):
Thanks a lot!
Unfortunately, there a a couple of changes needed.

hibernate-reactive-core/src/test/java/org/hibernate/ReactiveStatelessDefaultBatchSizeTest.java
```diff
@@ -0,0 +1,466 @@
...
+		sqlTracker.clear();
+	}
+
+	@Override
+	protected void addServices(StandardServiceRegistryBuilder builder) {
+		sqlTracker.registerService( builder );
+	}
+
+	private static boolean filter(String s) {
+		String[] accepted = { "select", "insert ", "update ", "delete " };
```
DavideD, line 83: Why are you also including the `select`?
Basically, you are not excluding any log with this filter.

dreab8: removed it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -150,10 +150,12 @@ private ReactiveStatelessSessionImpl(
...
-		Integer batchSize = getConfiguredJdbcBatchSize();
-		reactiveConnection = batchSize == null || batchSize < 2
-				? connection
-				: new BatchingConnection( connection, batchSize );
+		// Don't know why ORM set the batch size to 0 in the super constructor,
+		// but the default should be the one in the factory options
+		int defaultBatchSize = factory.getSessionFactoryOptions().getJdbcBatchSize();
+		setJdbcBatchSize( defaultBatchSize );
+		// BatchingConnection will use the original connection if defaultBatchSize <= 1
+		reactiveConnection = new BatchingConnection( connection, defaultBatchSize );
```
gavinking, line 158: Need to be careful about this. I'm not certain, but I think this might have the side-effect of making the stateless session asynchronous, which we don't want it to be by default.

gavinking: Yeah, so I just checked. Reconstructing the history behind this, it looks like historically `StatelessSession` did not allow JDBC batching, because that would change it's "immediate synchronous execution" model into something more like transactional write-behind and be confusing.

It appears that Vlad _accidentally_ added `setJdbcBatchSize()` to `StatelessSession`, breaking its historical guarantees, in https://hibernate.atlassian.net/browse/HHH-10431. I've attempted to move _back away from that_ in H7 by introducing `insertMultiple()` and friends as an alternative.

Anyway:

1. I would prefer if `StatelessSession` did _not_ have a way to set a session-level batch size, and
2. it certainly should default to a batch size of zero.

So I think this change is wrong.

gavinking (changes requested):

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
DavideD: OK, it makes sense now. Thanks.

> 1. I would prefer if StatelessSession did not have a way to set a session-level batch size, and

I don't think we need them in the API, everything should work if we change the type of `batchingHelperSession`  to `ReactiveStatelessSessionImpl` (in `ReactiveStatelessSessionImpl`)

> 2. it certainly should default to a batch size of zero.

I think we can just change the original code to
```java
```

DavideD: OK, it makes sense now. Thanks.

> 1. I would prefer if StatelessSession did not have a way to set a session-level batch size, and

I don't think we need them in the API, everything should work if we change the type of `batchingHelperSession`  to `ReactiveStatelessSessionImpl` (in `ReactiveStatelessSessionImpl`)

> 2. it certainly should default to a batch size of zero.

I think this means that we can now simplify the code to 
```java
reactiveConnection = new BatchingConnection( connection, 0 );
```
but, `statelessSession.insertAll(objects)` will never use batching, right?.

dreab8: if we cannot also set the jdbc bacth size of the batchingHelperSession then methods like `insertAll(int batchSize, ... )` are not going to work  because we are going to use `new BatchingConnection( connection, 0 )` that will not execute any batch.

gavinking: @DavideD 

`statelessSession.insertAll(objects)` will use batching because it's directly calling `setJdbcBatchSize()` on the internal SPI. And since it uses _exactly the right batch size_, there's no write-behind behavior.

DavideD: so, insertAll will always batch, right?

gavinking: That's the idea, yeah.

dreab8: working on it

dreab8: @gavinking pushed the changes.

gavinking (approved):
LGTM, thanks!

dreab8:
@gavinking @DavideD thanks a lot for the review

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyStatelessSessionImpl.java
```diff
@@ -163,7 +163,7 @@ public Uni<Void> deleteAll(Object... entities) {
 
 	@Override
 	public Uni<Void> deleteAll(int batchSize, Object... entities) {
-		return uni( () -> delegate.reactiveDeleteAll( entities ) );
+		return uni( () -> delegate.reactiveDeleteAll( batchSize, entities ) );
```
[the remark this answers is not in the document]
dreab8: they were delegating to the correct method, only `deleteAll` was wrong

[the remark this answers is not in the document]
DavideD: Sorry, I meant the insertAll withouth the bacthSize.
`updateAll` looks like this:
```
	@Override
	public Uni<Void> updateAll(Object... entities) {
		return uni( () -> delegate.reactiveUpdateAll( entities ) );
	}
```

shouldn't it be:
```
	@Override
	public Uni<Void> updateAll(Object... entities) {
		return uni( () -> delegate.reactiveUpdateAll( entities.lenght, entities ) );
	}
```
?

DavideD:
> Are we missing a StatelessSession.upsertAll()? Did someone (me, I guess) forget about it?

I feel like this issue is never going to end :D
Let's add it as new feature in separate PR/issue

DavideD:
Merged, thanks

DavideD:
Actually, what's the plan for all the `*All` methods? Are we going to remove them or are they here to stay?
I'm asking because we don't have `upsertAll` but we have `upsertMultiple`

gavinking:
Good question, I'm not sure. But I guess I'm not keen on deprecating things unless there's a good reason to do so.

Issue #2156, opened 2025-03-25
Add runWithConnection(), callWithConnection()


gavinking:
We should add the ability to do work against a Vert.x `SqlConnection` by adding operations like `EntityManager.xxxWithConnection()`.

DavideD:
We decided not to do this:
* We don't want to expose the Vert.x client in the API
* When using Quarkus, it's possible to inject the pool and get a connection: https://quarkus.io/guides/reactive-sql-clients#multiple-datasources

We can re-open this if we change mind

PR #2157, opened 2025-03-25
[#2156] add withConnection()

gavinking:


DavideD:
Closing this one because we decided that we don't need it at the moment

Issue #2138, opened 2025-03-03
Updating an entity updates the creation timestamp (and the update timestamp)


DavideD:
See https://github.com/quarkusio/quarkus/issues/39016

We need to check if this is actually happening in Hibernate Reactive and add tests for it (if we don't have some already).

> After updating an entity both createdAt and updatedAt are returned with the the updatedAt value. However, createdAt is not being persisted to the database. When I list the records from the database the results are returned correctly. I'm using the same DTO object for both operations.

```java
Column(name = "created_at", updatable = false, nullable = false)
@CreationTimestamp
public LocalDateTime createdAt;

@Column(name = "updated_at", nullable = false)
@UpdateTimestamp
public LocalDateTime updatedAt;
```

PR #2172, opened 2025-03-31
Use UPDATE event type for generator where applicable

marko-bekhta:
fixes https://github.com/hibernate/hibernate-reactive/issues/2138

I've added the test as disabled, as we'd need an ORM release that fixes https://hibernate.atlassian.net/browse/HHH-19306 first

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/mutation/ReactiveUpdateCoordinatorStandard.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/mutation/ReactiveUpdateCoordinatorStandard.java
@@ -30,7 +30,7 @@ import org.hibernate.sql.model.MutationOperationGroup;
 import org.hibernate.tuple.entity.EntityMetamodel;
 
 import static org.hibernate.engine.jdbc.mutation.internal.ModelMutationHelper.identifiedResultsCheck;
-import static org.hibernate.generator.EventType.INSERT;
+import static org.hibernate.generator.EventType.UPDATE;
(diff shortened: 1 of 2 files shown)


DavideD:
Thanks for the PR, but does it make sense to merge this without the fix for ORM?
I don't see any benefit, I think I would prefer to have the test enabled and then apply this fix after the next ORM release.

marko-bekhta:
yeah... that would be best to wait for the ORM fix and then apply this patch... I did want to run the CI build against all DBs (was only testing postgres locally) to make sure it doesn't break something else 😃

DavideD:
Make sense, but the fix looks simple enough that I don't expect problems with the other databases (finger crossed).
And even if it does, it shouldn't be too hard to fix (or we can fix it as a separate issue).

You've already done the hard part 😃 
(figuring out the issue and propose a solution)

I'm more worried about merging this and forgetting to reenable the test.

marko-bekhta:
👍🏻 rebased and changed the patch for wip/3.0. 
I'm thinking since we have that fix for 6.6 too (https://github.com/hibernate/hibernate-orm/pull/9946), maybe we want to backport to 2.4 so we could bump it in Quarkus?

DavideD:
I will backport it, thanks

DavideD:
@marko-bekhta Could you rebase this on top of main, please? We can merge it now

marko-bekhta:
changing the base branch in PR makes GH do strange things 😄, but looks like tests are running now 🤞🏻

DavideD:
Thanks @marko-bekhta, perfect!

Issue #2159, opened 2025-03-25
Fix issue with JDK 23 on GitHub


DavideD:
The configured version cannot be downloaded

PR #2183, opened 2025-04-07
[2.4] Remove JDK 23 and use JDK 24

DavideD:
Backport fix for #2159


DavideD:
I've already done this

PR #2225, opened 2025-04-25
log the HR version

gavinking:


DavideD:
I haven't tried it, but LGTM

PR #2246, opened 2025-05-05
Remove MutinyStatelessSessionDelegator#upsert(String entityName, Object entity) method

dreab8:
The code was not compiling
```
delegate().upsert(entityName, entity);
```
has been removed so I think the aim was to get rid of this method.


gavinking:
Hah! Shit, I had two parallel pull requests, and they broke each other. (Even though the CI passed for both.) Strange that this is actually the first time this has ever happened to me.

gavinking (approved):

PR #2262, opened 2025-05-21
Upgrade Vert.x to 4.5.15 (from 4.5.14)

### LINKED ISSUE #2261: Upgrade Vert.x to 4.5.15 (from 4.5.14)

DavideD:
Fix #2261

We should wait until Quarkus upgrades this too

See https://github.com/quarkusio/quarkus/blob/4abbcf233f6e75bc86ab070ad3d40ada0502e0e3/bom/application/pom.xml#L114

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,11 @@ Hibernate Reactive has been tested with:
 - MS SQL Server 2022
 - Oracle 23
 - [Hibernate ORM][] 7.0.0.Final
-- [Vert.x Reactive PostgreSQL Client](https://vertx.io/docs/vertx-pg-client/java/) 4.5.14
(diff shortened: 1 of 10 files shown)


yrodiere:
It doesn't look like you had to change any code, so I expect that even after the upgrade, Hibernate Reactive will still work in Quarkus, even if it overrides the Vert.x dependencies to 4.5.14... no?

DavideD:
It requires no changes in Hibernate Reactive, but with Quarkus I never know.

yrodiere:
Feel free to merge, we can always delay the upgrade to Reactive 3 in Quarkus until the vertx upgrade if this turns out to be a problem.

PR #2263, opened 2025-05-21
Upgrade database images for testing

DavideD:
* Upgrade PostgreSQL to 17.5 (from 17.4)
* Upgrade CockroachDB to v24.3.13 (from v24.1.15)

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/PostgreSQLDatabase.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/PostgreSQLDatabase.java
@@ -87,7 +87,7 @@ class PostgreSQLDatabase implements TestableDatabase {
 	 * TIP: To reuse the same containers across multiple runs, set `testcontainers.reuse.enable=true` in a file located
 	 * at `$HOME/.testcontainers.properties` (create the file if it does not exist).
 	 */
-	public static final PostgreSQLContainer<?> postgresql = new PostgreSQLContainer<>( imageName( "postgres", "17.4" ) )
+	public static final PostgreSQLContainer<?> postgresql = new PostgreSQLContainer<>( imageName( "postgres", "17.5" ) )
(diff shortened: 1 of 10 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/containers/PostgreSQLDatabase.java
```diff
@@ -87,7 +87,7 @@ class PostgreSQLDatabase implements TestableDatabase {
 	 * TIP: To reuse the same containers across multiple runs, set `testcontainers.reuse.enable=true` in a file located
 	 * at `$HOME/.testcontainers.properties` (create the file if it does not exist).
 	 */
-	public static final PostgreSQLContainer<?> postgresql = new PostgreSQLContainer<>( imageName( "postgres", "17.4" ) )
+	public static final PostgreSQLContainer<?> postgresql = new PostgreSQLContainer<>( imageName( "postgres", "17.5" ) )
```
marko-bekhta, line 90: IDK if you'll find it usefull .. but I noticed you are updating DB versions  😃.
In Search we moved the "version" into the docker file:
https://github.com/hibernate/hibernate-search/blob/main/build/container/database/postgresql.Dockerfile
and then the testcontainer uses that docker file to start a db: https://github.com/hibernate/hibernate-search/blob/14a8255a221bfd49b09ad205ab197fe21b572087/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java#L373

this way you can make dependabot keep your database versions updated:
https://github.com/hibernate/hibernate-search/blob/14a8255a221bfd49b09ad205ab197fe21b572087/.github/dependabot.yml#L167-L183

DavideD: Thanks, that's a good hint. I will create an issue for it

Issue #1793, opened 2023-11-21
HR000075: Detected call to non reactive method. Alternative reactive method: `reactiveAssemble`


Xset-s:
Hi, I'm using Hibernate Reactive 2.0.6.Final and getting HR000075 when I try to call a custom constructor in an HQL query

book (from an example https://github.com/hibernate/hibernate-reactive/tree/main/examples/session-example, but with a custom constructor):

```@Entity
public class Book {
	@Id @GeneratedValue
	private Integer id;

	@Size(min=13, max=13)
	private String isbn;

	...
	
	public Book(Author author) {
		this.author = author;
	}
}```



author:

```
@Entity
public class Author {
	@Id @GeneratedValue
	private Integer id;

	@NotNull @Size(max=100)
	private String name;

	@OneToMany(mappedBy = "author", cascade = PERSIST)
	private List<Book> books = new ArrayList<>();
	
	...
}```

code:

```
// define some test data
Author author1 = new Author("Iain M. Banks");
Author author2 = new Author("Neal Stephenson");
Book book1 = new Book("1-85723-235-6", "Feersum Endjinn", author1, LocalDate.of(1994, JANUARY, 1));
Book book2 = new Book("0-380-97346-4", "Cryptonomicon", author2, LocalDate.of(1999, MAY, 1));
Book book3 = new Book("0-553-08853-X", "Snow Crash", author2, LocalDate.of(1992, JUNE, 1));
author1.getBooks().add(book1);
author2.getBooks().add(book2);
author2.getBooks().add(book3);

factory.withSession(
                            // query the Book titles
                            session -> session.createQuery(
                                            "SELECT NEW Book(book.author) from Book book order by title desc",
                                            Book.class
                                    )
                                    .getResultList()
                    )
                    .await().indefinitely();
```

exception:

```
Exception in thread "main" org.hibernate.HibernateException: HR000075: Detected call to non reactive method. Alternative reactive method: `reactiveAssemble`
	at org.hibernate.reactive.sql.results.graph.ReactiveDomainResultsAssembler.assemble(ReactiveDomainResultsAssembler.java:32)
	at org.hibernate.sql.results.graph.instantiation.internal.ArgumentReader.assemble(ArgumentReader.java:35)
	at org.hibernate.sql.results.graph.instantiation.internal.DynamicInstantiationAssemblerConstructorImpl.assemble(DynamicInstantiationAssemblerConstructorImpl.java:46)
[...]

PR #2280, opened 2025-05-28
Test case showing the issue has been resolved

dreab8:
test to show #1793 has been fixed

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/HQLQueryTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/HQLQueryTest.java
@@ -5,6 +5,10 @@
  */
 package org.hibernate.reactive;
 
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/HqlSelectNewConstructorTest.java
```diff
@@ -0,0 +1,127 @@
...
+	public void testSimpleUpdateSameTransaction(VertxTestContext context) {
+		test( context, getMutinySessionFactory()
+				.withTransaction( session ->
+					session.createQuery(
+									"SELECT NEW Book(book.author) from Book book order by title desc",
+									Book.class
+							)
+							.getResultList()
+				 )
+				.invoke( books -> assertThat( books ).hasSize( 3 ) )
```
DavideD, line 54: Could you also check that the books actually contain the correct author, please?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/HqlSelectNewConstructorTest.java
```diff
@@ -0,0 +1,127 @@
...
+import jakarta.persistence.OneToMany;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import static jakarta.persistence.CascadeType.PERSIST;
+import static jakarta.persistence.FetchType.LAZY;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HqlSelectNewConstructorTest extends BaseReactiveTest {
```
DavideD, line 25: Sorry, I should have mentioned this before, but we already have [a class that tests HQL queries](https://github.com/hibernate/hibernate-reactive/blob/a98716ba3b861ee1194895566c963c216e7c7421/hibernate-reactive-core/src/test/java/org/hibernate/reactive/HQLQueryTest.java#L31).

We create a SessionFactory for each separate class, and in the long run it could slow down the build having too many of them.

DavideD (changes requested):
Thanks a lot!

It looks good overall, but the test seems simple enough to be included in one of the existing test classes. I left a comment about it.

PR #2291, opened 2025-06-05
Add AUTHORS.txt

marko-bekhta:
see https://github.com/hibernate/hibernate-orm/pull/10266


DavideD:
Is this change only for main, or should I update the other branches too?

marko-bekhta:
> Is this change only for main, or should I update the other branches too?

This one is probably for @yrodiere ^ ?

yrodiere:
It's fine to add the file to `main` only IMO. Thanks.

Issue #2154, opened 2025-03-25
Composite key with generated value not working properly


majchrzw:
Hello, I encountered a bug when implementing composite key with generated value. My code is based on this tutorial: https://vladmihalcea.com/how-to-map-a-composite-identifier-using-an-automatically-generatedvalue-with-jpa-and-hibernate/ .
I did everything like in the link but it does not work with hibernate reactive. When i try to persist my entity i receive weird error:
```
org.hibernate.HibernateException: java.util.concurrent.CompletionException: org.hibernate.exception.GenericJDBCException: Unable to acquire JDBC Connection [Not using JDBC] [n/a]
```
When I used the same code in non-reactive hibernate everything is working as intended.
I created 2 reproducers here: https://github.com/majchrzw/composite-key .
`main` branch show the issue with hibernate reactive, while `hibernate-orm` branch show the working example. 
Can this be somehow fixed?

adampoplawski:
also note that this is working with single keys (non-composite)

https://hibernate.zulipchat.com/#narrow/channel/132096-hibernate-user/topic/Weird.20error.20while.20using.20hibernate.20reactive.20and.20composite.20key/with/508017968
But according to chat it seems that this would be invoking blocking call in background

https://hibernate.org/reactive/documentation/3.0/reference/html_single/#_identifier_generation
Also maybe I am not understanding docs but this states that @GeneratedValue is supported.

yrodiere:
I would imagine the problem boils down to this:

```
	at org.hibernate.id.CompositeNestedGeneratedValueGenerator.generate(CompositeNestedGeneratedValueGenerator.java:135)
```

Hibernate Reactive just needs its own version of `CompositeNestedGeneratedValueGenerator`, most likely?

gavinking:
Can we please calm down here?

`@GeneratedValue` is supported in Hibernate Reactive for simple primary keys. Yes, nonblockingly _of course_. 

Note that the Javadoc for `@GeneratedValue` clearly states that:

> The persistence * provider is only required to support the {@code GeneratedValue} for simple primary keys.

Hibernate Reactive does not (yet) allow `@GeneratedValue` with **composite primary keys**, and, indeed, as far as I recall, this is something Hibernate ORM itself did not historically support and has only recently begun to allow.

If this is a problem, code contributions are always welcome!

yrodiere:
Gavin, the fact that we report, and track, a problem, does not mean anyone is panicking.

[...]

PR #2293, opened 2025-06-06
Improve error message when asking for a JDBC connection

### LINKED ISSUE #2186: Improve error message for obtaining a JDBC connection
When the application tries to access a JDBC connection, we throw the following error:

```
Caused by: java.sql.SQLException: Not using JDBC
```

This happens in [NoJdbcConnectionProvider#getConnection](https://github.com/hibernate/hibernate-reactive/blob/3d6cc0fb74c729788132ce35da948705ac03ab2d/hibernate-reactive-core/src/main/java/org/hibernate/reactive/provider/service/NoJdbcConnectionProvider.java#L25).

When this happens, it means that we haven't tested a particular mapping or configuration. In a perfect world, a user should never see this error.

We should think of a better error message to explain the user that their use case is currently non supported.
And it makes sense to add the error in the logger.

DavideD:
Fix #2186

@yrodiere, is this better?

@dreab8, You are workging on https://github.com/hibernate/hibernate-reactive/issues/2154, right? Do you have a test case I can borrow to see the error? Thanks

The new error message:
```
The application requested a JDBC connection, but Hibernate Reactive doesn't use JDBC.
This could be caused by a bug or the use of an unsupported feature in Hibernate Reactive
```

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
@@ -7,6 +7,7 @@ package org.hibernate.reactive.logging.impl;
 
 
 
+import java.sql.SQLException;
 import java.sql.SQLWarning;
 
 import jakarta.persistence.PersistenceException;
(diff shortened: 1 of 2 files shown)


yrodiere (approved):
LGTM, thanks!

PR #2331, opened 2025-07-04
Migrate to Gradle version catalogs

### LINKED ISSUE #1989: Consider using Gradle version catalogs
Consider using Gradle [version catalogs](https://docs.gradle.org/current/userguide/platforms.html) which have a number of benefits -

1. centralized definiton of versions
2. centralized definiton of dependency artifacts
3. DSL validation (auto completion, etc) of references to catalog versions and artifacts

exoego:
Closes #1989 

- Definition of versions is centralized in version catalogs
- Some core lib versions can still be overridden in the same manner. See below:

```bash
# default (no override)
❯ ./gradlew compileJava                                                                
Java versions for main code: [compiler:21, release:17]
Java versions for tests: [compiler:21, release:21, launcher:21]

> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
ORM version: 7.0.2.Final
ORM Gradle plugin version: 7.0.2.Final
Vert.x SQL Client version: 5.0.0
```

```bash
# core libs version override 
❯ ./gradlew compileJava -PhibernateOrmVersion=7.0.4.Final -PvertxSqlClientVersion=5.0.1
Java versions for main code: [compiler:21, release:17]
Java versions for tests: [compiler:21, release:21, launcher:21]

> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
ORM version: 7.0.4.Final    <--- ✅ override
ORM Gradle plugin version: 7.0.2.Final
Vert.x SQL Client version: 5.0.1  <--- ✅ override
```

Mostly auto-transformed by [gradle-version-catalog-cli](https://github.com/exoego/gradle-version-catalogs-cli) which I wrote.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -33,23 +33,3 @@ org.gradle.java.installations.auto-download=false
 
 # Enable the maven local repository (for local development when needed) when present (value ignored)
 #enableMavenLocalRepo = true
-
-# The default Hibernate ORM version (override using `-PhibernateOrmVersion=the.version.you.want`)
-hibernateOrmVersion = 7.0.2.Final
-
-# Override default Hibernate ORM Gradle plugin version
-# Using the stable version because I don't know how to configure the build to download the snapshot version from
-# a remote repository
-#hibernateOrmGradlePluginVersion = 7.0.2.Final
-
-# If set to true, skip Hibernate ORM version parsing (default is true, if set to null)
-# this is required when using intervals or weird versions or the build will fail
-#skipOrmVersionParsing = true
-
-# Override default Vert.x Sql client version
-#vertxSqlClientVersion = 5.0.0-SNAPSHOT
-
-# Override default Vert.x Web client and server versions. For integration tests, both default to vertxSqlClientVersion
-#vertxWebVersion = 5.0.0
-#vertxWebtClientVersion = 5.0.0
-
--- a/build.gradle
+++ b/build.gradle
@@ -3,32 +3,14 @@ plugins {
 
     id 'java-library'
     id 'maven-publish'
-    id 'com.diffplug.spotless' version '6.25.0'
-    id 'org.asciidoctor.jvm.convert' version '4.0.2' apply false
-    id 'io.github.gradle-nexus.publish-plugin' version '1.3.0'
+    alias(libs.plugins.com.diffplug.spotless)
+    alias(libs.plugins.org.asciidoctor.jvm.convert) apply false
+    alias(libs.plugins.io.github.gradle.nexus.publish.plugin)
 }
 
 group = "org.hibernate.reactive"
 // leverage the ProjectVersion which comes from the `local.versions` plugin
 version = project.projectVersion.fullName
-
-// Versions which need to be aligned across modules; this also
-// allows overriding the build using a parameter, which can be
-// useful to monitor compatibility for upcoming versions on CI:
-//
-// ./gradlew clean build -PhibernateOrmVersion=5.6.15-SNAPSHOT
-ext {
-    // Mainly, to allow CI to test the latest versions of Vert.X
-    // Example:
-    // ./gradlew build -PvertxSqlClientVersion=4.0.0-SNAPSHOT
-    if ( !project.hasProperty( 'vertxSqlClientVersion' ) ) {
-        vertxSqlClientVersion = '5.0.0'
-    }
-
-    testcontainersVersion = '1.21.0'
-
-    logger.lifecycle "Vert.x SQL Client Version: " + project.vertxSqlClientVersion
-}
 
 subprojects {
     apply plugin: 'java-library'
@@ -135,3 +117,10 @@ subprojects {
     }
 }
 
+rootProject.afterEvaluate {
+    // Workaround since "libs.versions.NAME" notation cannot be used here
+    def libs = project.extensions.getByType(VersionCatalogsExtension).named('libs')
+    logger.lifecycle "ORM version: ${libs.findVersion('hibernateOrm').get().requiredVersion}"
+    logger.lifecycle "ORM Gradle plugin version: ${libs.findVersion('hibernateOrmGradlePlugin').get().requiredVersion}"
+    logger.lifecycle "Vert.x SQL Client version: ${libs.findVersion('vertxSqlClient').get().requiredVersion}"
+}
--- a/documentation/build.gradle
+++ b/documentation/build.gradle
@@ -52,7 +52,7 @@ def aggregateJavadocsTask = tasks.register( 'aggregateJavadocs', Javadoc ) {
 		use = true
 		options.encoding = 'UTF-8'
 
-		def matcher = hibernateOrmVersion =~ /\d+\.\d+/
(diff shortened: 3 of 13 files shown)


DavideD:
Thanks, I will have a look at this soon

DavideD (changes requested):
Thanks a lot, looks great!
But I think something got lost in the changes.

When I run a test in the main branch, it prints:
```
> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
ORM version: 7.0.2.Final
ORM Gradle plugin version: 7.0.2.Final
Vert.x SQL Client Version: 5.0.0
```

But after your changes, I see:
```
> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
```

This information is quite because we tend to run Hibernate Reactive with different configurations.

Also, very nitpicky, but could you also add the issue number before the commit message? It makes my life easier when I need to track down changes.
I usually add it like this:
```
[#2331] Migrate to Gradle version catalogs
```

@sebersole Could you also have a look if this is what you had in mind when you created the issue?  The approach looks good to me overall, but I don't have a lot of experience with catalogs.

Thanks a lot [exoego](https://github.com/exoego)!

gradle.properties
```diff
@@ -33,23 +33,3 @@ org.gradle.java.installations.auto-download=false
...
-hibernateOrmVersion = 7.0.2.Final
-
-# Override default Hibernate ORM Gradle plugin version
-# Using the stable version because I don't know how to configure the build to download the snapshot version from
-# a remote repository
-#hibernateOrmGradlePluginVersion = 7.0.2.Final
-
-# If set to true, skip Hibernate ORM version parsing (default is true, if set to null)
-# this is required when using intervals or weird versions or the build will fail
-#skipOrmVersionParsing = true
```
DavideD, line 47: I don't see this property mentioned anywhere anymore

gradle.properties
```diff
@@ -33,23 +33,3 @@ org.gradle.java.installations.auto-download=false
...
-
-# Override default Vert.x Sql client version
-#vertxSqlClientVersion = 5.0.0-SNAPSHOT
-
-# Override default Vert.x Web client and server versions. For integration tests, both default to vertxSqlClientVersion
-#vertxWebVersion = 5.0.0
-#vertxWebtClientVersion = 5.0.0
```
DavideD, line 54: It's quite helpful to have all these properties in one place, will they still work if we keep them here?

gradle.properties
exoego: Is it still needed?
IIUC, if `skipOrmVersionParsing ` is false, `hibernateOrmVersion` is parsed and may raise if version has no `.` in it.
https://github.com/hibernate/hibernate-reactive/blob/c254ebb5a6cda29843b8a3c1a8d6f8713450eec5/local-build-plugins/src/main/java/org/hibernate/reactive/env/ProjectVersion.java#L17-L24

When one wants to test "weird versions", why not simply use that version ? 🤔  

`❯ ./gradlew compileJava -PhibernateOrmVersion=1234weiredversion`

gradle.properties
exoego: I see.
Brought back the same feature here:
https://github.com/hibernate/hibernate-reactive/pull/2331/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R120-R126

```
❯ ./gradlew compileJava -PhibernateOrmVersion=7.0.4.Final -PvertxSqlClientVersion=5.0.1
Java versions for main code: [compiler:21, release:17]
Java versions for tests: [compiler:21, release:21, launcher:21]

> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
ORM version: 7.0.4.Final    <--- ✅ 
ORM Gradle plugin version: 7.0.2.Final
Vert.x SQL Client version: 5.0.1  <--- ✅
```

I initially thought it should be implemented near the place where version override is implemented.
However, realized that  `settings.gradle` is evaluated before `build.gradle` hence before `VersionsPlugin`.
So I used `rootProject.afterEvaluate` so the versions are shown as same as before.

exoego:
> This information is quite because we tend to run Hibernate Reactive with different configurations.


I see.
Brought back the same feature here:
https://github.com/hibernate/hibernate-reactive/pull/2331/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R120-R126

```
❯ ./gradlew compileJava -PhibernateOrmVersion=7.0.4.Final -PvertxSqlClientVersion=5.0.1
Java versions for main code: [compiler:21, release:17]
Java versions for tests: [compiler:21, release:21, launcher:21]

> Configure project :
Project version: 4.0.0-SNAPSHOT (4.0)
Release version: n/a
Development version: n/a
ORM version: 7.0.4.Final    <--- ✅ 
ORM Gradle plugin version: 7.0.2.Final
Vert.x SQL Client version: 5.0.1  <--- ✅
```

If no override, it shows default versions defined in `libs.versions.toml` (version catalogs).


I initially thought it should be implemented near the place where version override is implemented.
However, realized that  `settings.gradle` is evaluated before `build.gradle` hence before `VersionsPlugin`.
So I used `rootProject.afterEvaluate` in `build.gradle`, so the versions are shown as same as before.

gradle.properties
exoego: > will they still work if we keep them here?

No, keeping versions in `gradle.properties` does not make sense, since there is no way to refer these gradle properties within `libs.versions.toml`.
All versions are now managed in one place, version catalog `gradle/libs.versions.toml`.

Please refer `[versions]` section in the catalog.
The section defines version variables like before in `gradle.properties`.

DavideD: Having the properties here (even if commented) highlights which versions you can override from the command line with a parameter. 
But I guess it's not a big deal, I will need to update the README

gradle.properties
DavideD: I thought I answered this, but it seems I dreamt it.

The reason the property and these checks exist is to prevent releasing a library with the wrong version or pushing the wrong version upstream.

DavideD: > hibernateOrmVersion is parsed and may raise if version has no . in it.

Perfect, because the official Hibernate ORM versions  are always in the format `major.minor.micro` at least. Anything else should fail.

DavideD:
Superseded by https://github.com/hibernate/hibernate-reactive/pull/2337

@exoego, Could you have a look and let me know if it makes sense? 
Thanks a lot!

sebersole:
> @sebersole Could you also have a look if this is what you had in mind when you created the issue? The approach looks good to me overall, but I don't have a lot of experience with catalogs.

Roughly, yes.  This does a bit more than I had anticipated doing, not sure yet whether they are good additions or not.  I can take a deeper look if you want.  

I see you created an alternative PR... I'll look through that one.

PR #2337, opened 2025-07-08
Migrate to Gradle version catalogs

### LINKED ISSUE #1989: Consider using Gradle version catalogs

DavideD:
Follows https://github.com/hibernate/hibernate-reactive/pull/2331
Fix #1989  

@exoego, I reverted the checks you have removed and made a few changes. Now we have the versions in the catalog but we also have the same behaviour as before.

I've also fixed a few mistakes I found along the way. I will apply these changes to other branches and make sure that everything still works.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,32 +3,14 @@ plugins {
 
     id 'java-library'
     id 'maven-publish'
-    id 'com.diffplug.spotless' version '6.25.0'
-    id 'org.asciidoctor.jvm.convert' version '4.0.2' apply false
-    id 'io.github.gradle-nexus.publish-plugin' version '1.3.0'
+    alias(libs.plugins.com.diffplug.spotless)
+    alias(libs.plugins.org.asciidoctor.jvm.convert) apply false
+    alias(libs.plugins.io.github.gradle.nexus.publish.plugin)
 }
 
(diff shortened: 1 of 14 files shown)


DavideD:
I think you have introduced the VersionsPlugin class to check that there are no errors in the versions name. The logic was initially in the gradle build and you moved it in a plugin (this is on top of my head).

But the VersionsPlugin expects to have a property with the Hibernate ORM version somewhere.
Before we had it in the gradle.properties or command line, but now the version is declared in the catalog and there's no property in the default build. This causes the plugin to fail.

I've created the VersionsTomlParser to read the versions from the [catalog](https://github.com/hibernate/hibernate-reactive/pull/2337/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df) as a fallback. 

I have the feeling that we can simplify a bit some part of this process, but I guess we can do it as a separate issue.

PR #2344, opened 2025-07-11
Update dependabot configuration

### LINKED ISSUE #1136: Check if we can integrate with Dependabot
See https://dev.to/cricketsamya/gradle-with-dependabot-1o47

~~See also this comment from @marko-bekhta: https://github.com/hibernate/hibernate-reactive/pull/2263#discussion_r2100247299~~
I've created a separate issue: https://github.com/hibernate/hibernate-reactive/issues/2269

DavideD:
Fix #1136 

Built on top of  https://github.com/hibernate/hibernate-reactive/pull/2341
Includes:

* Update project dependencies
* Update images in Dockerfile files under tooling/docker/
* Update images we use as services in the GitHub workflow (mysql and
  postgres to test the examples)

@marko-bekhta, could you have a look at the dependabot configuration in this commit and let me know if it makes sense?


.github/dependabot.yml
```diff
@@ -14,5 +14,49 @@ updates:
         patterns:
           - "*"
     allow:
-      - dependency-name: "actions/*"
-      - dependency-name: "redhat-actions/*"
+      - dependency-type: all
```
marko-bekhta: it's probably fine to include all actions to check for the updates since we currently only have the official `actions/` and the `codeql` actions used in the workflows...
For the background, why it was originally limited: the idea was to only allow updates for the explicitly defined trusted sources of actions. So I'll let you decide how you'd like to define that 😃, maybe just


```suggestion
    - dependency-name: "actions/*"
    - dependency-name: "github/codeql-action/*"
```
?

DavideD: I don't think it's that risky to enable all for now, but I will try to only add the configuration to check for the databases we are using for services.

DavideD: Actually, I just realized that there's no need to change those permissions... I guess I just needed a relaxing week-end

PR #2371, opened 2025-07-16
"Not using JDBC" when locking previously loaded entities in Hibernate Reactive 3

### LINKED ISSUE #2332: "Not using JDBC" when locking previously loaded entities in Hibernate Reactive 3
This seems to be a regression of #1258 in Hibernate Reactive 3. When trying to create a lock on an entity that already exists within the current session, this fails with the following error:

```
Caused by: java.sql.SQLException: HR000084: The application requested a JDBC connection, but Hibernate Reactive doesn't use JDBC. This could be caused by a bug or the use of an unsupported feature in Hibernate Reactive
	at org.hibernate.reactive.provider.service.NoJdbcConnectionProvider.getConnection(NoJdbcConnectionProvider.java:30)
	at org.hibernate.internal.NonContextualJdbcConnectionAccess.obtainConnection(NonContextualJdbcConnectionAccess.java:44)
	at org.hibernate.resource.jdbc.internal.LogicalConnectionManagedImpl.acquireConnectionIfNeeded(LogicalConnectionManagedImpl.java:118)
	... 104 more
```

This was still working fine with Hibernate 6 / HR 2.

Reproducer: https://github.com/markusdlugi/hr-lock-reproducer

Steps to reproduce:

1. Run PersonResourceTest
2. Observe that the test fails with an HTTP 500 and the aforementioned error message
3. Remove the lock and observe that the test now passes
4. Alternatively, change the first `findById()` call to fetch the entity with a lock from the start and observe that the test also passes this way

dreab8:
Fix #2332 

When loading an entity already present in the PersistenceContext but with a greater lock mode `DefaultReactiveLoadEventListener` ends up calling `EntityPersister#lock()` method instead of `ReactiveEntityPersister#reactiveLock` 

To be merged the PR needs the following change in Hibernate Core https://github.com/hibernate/hibernate-orm/pull/10550

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/LockOnLoadTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/LockOnLoadTest.java
@@ -1,0 +1,67 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import org.hibernate.LockMode;
+
+import org.junit.jupiter.api.Test;
+
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import java.util.Collection;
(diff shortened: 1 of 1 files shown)


DavideD:
Superseded by https://github.com/hibernate/hibernate-reactive/pull/2436

PR #2380, opened 2025-07-24
Support @IdGeneratorType

### LINKED ISSUE #1906: Support `@IdGeneratorType`
`@GenericGenerator` has been deprecated in 6.5. See: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/annotations/GenericGenerator.java#L64

dreab8:
Fix #1906

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BeforeExecutionIdGeneratorTypeTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BeforeExecutionIdGeneratorTypeTest.java
@@ -1,0 +1,107 @@
+/* Hibernate, Relational Persistence for Idiomatic Java
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright: Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import org.hibernate.annotations.IdGeneratorType;
+import org.hibernate.generator.EventType;
+import org.hibernate.reactive.id.ReactiveIdentifierGenerator;
+import org.hibernate.reactive.session.ReactiveConnectionSupplier;
+import org.hibernate.reactive.util.impl.CompletionStages;
+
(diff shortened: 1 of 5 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/BeforeExecutionIdGeneratorTypeTest.java
```diff
@@ -0,0 +1,107 @@
...
+	@Override
+	protected Collection<Class<?>> annotatedEntities() {
+		return List.of( Person.class );
+	}
+
+	@Test
+	public void testPersist(VertxTestContext context) {
+		Person person = new Person();
+		test(
+				context, openSession()
```
DavideD, line 45: Could you use `getMutinySessionFactory.withTransaction`?
I think it's more similar to a typical use case.

Maybe we could also test without transactions, but we should at least add a comment to explain what's going on

DavideD: Would it make sense to also read the value from the database to make sure  that they match? Maybe it's overkill

DavideD (changes requested):
Thanks!
I gave it a quick look, but I will check better tomorrow

hibernate-reactive-core/src/test/java/org/hibernate/reactive/BeforeExecutionIdGeneratorTypeTest.java
```diff
@@ -0,0 +1,107 @@
...
+						.thenCompose( session -> session.persist( person ) )
+						.thenAccept( v -> {
+							assertThat( person.getId() ).isNotNull();
```
DavideD, line 48: I just realized that this test is wrong because the id is a primitive type so it's never going to be null

PR #2476, opened 2025-08-28
Support @IdGeneratorType

dreab8:
and fix few issues with `OnExecutionGenerator`s


hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveAbstractReturningDelegate.java
```diff
@@ -6,133 +6,132 @@
...
 					}
 					valueBindings.afterStatement( statementDetails.getMutatingTableDetails() );
 				} );
 	}
 
-	default GeneratedValues validateGeneratedIdentityId(GeneratedValues generatedId) {
-		if ( generatedId == null ) {
+	public GeneratedValues validateGeneratedIdentityId(GeneratedValues generatedValues) {
+		if ( generatedValues == null ) {
 			throw make( Log.class, MethodHandles.lookup() ).noNativelyGeneratedValueReturned();
```
DavideD: Now that this is not an interface, you can create a LOG field for this.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveAbstractReturningDelegate.java
```diff
@@ -6,133 +6,132 @@
...
 
 		// CockroachDB might generate an identifier that fits an integer (and maybe a short) from time to time.
 		// Users should not rely on it, or they might have random, hard to debug failures.
-		Type identifierType = getPersister().getIdentifierType();
+		Type identifierType = persister.getIdentifierType();
 		if ( ( identifierType.getReturnedClass().equals( Short.class ) || identifierType.getReturnedClass().equals( Integer.class ) )
-				&& getPersister().getFactory().getJdbcServices().getDialect() instanceof CockroachDialect ) {
-			throw make( Log.class, MethodHandles.lookup() ).invalidIdentifierTypeForCockroachDB( identifierType.getReturnedClass(), getPersister().getEntityName() );
+				&& persister.getFactory().getJdbcServices().getDialect() instanceof CockroachDialect ) {
+			throw make( Log.class, MethodHandles.lookup() ).invalidIdentifierTypeForCockroachDB( identifierType.getReturnedClass(), persister.getEntityName() );
```
DavideD: Now that this is not an interface, you can create a LOG field for this.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/generator/values/internal/ReactiveGeneratedValuesHelper.java
```diff
@@ -81,26 +93,32 @@ && noCustomSql( persister, timing ) ) {
...
-		return timing == EventType.INSERT ? dialect.supportsInsertReturning() : dialect.supportsUpdateReturning();
+	public static boolean supportReactiveGetGeneratedKey(Dialect dialect, List<? extends ModelPart> generatedProperties) {
+		return dialect instanceof OracleDialect
+				|| (dialect instanceof MySQLDialect && generatedProperties.size() == 1 && !(dialect instanceof MariaDBDialect));
+	}
+
+	public static boolean supportsReturning(Dialect dialect, EventType timing) {
+		return timing == EventType.INSERT
+				? supportsInsertReturning(dialect)
+				: dialect.supportsUpdateReturning() || dialect instanceof CockroachDialect;
```
DavideD: Please, keep track of the various `instance of CockroachDialect`  and the fact that we can remove them once we upgrade ORM (if the the relative issue has been solved).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveAbstractReturningDelegate.java
DavideD: Double check if there are other places we do this, please

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutationDelegateIdentityTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.reactive.containers.DatabaseConfiguration.DBType.ORACLE;
+
+/**
+ * Inspired by the test
+ * {@code org.hibernate.orm.test.mapping.generated.delegate.MutationDelegateIdentityTest}
+ * in Hibernate ORM.
+ */
+public class MutationDelegateIdentityTest extends BaseReactiveTest {
```
DavideD, line 52: I noticed this test doesn't include a couple of things from the test in ORM (the same applies for `MutationDelegateTest`):
1. It doesn't enable the batch size (the test in ORM says that it should be ignored)
2. It doesn't test native queries: in reactive we can test this via the [SqlStatementTracker](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/testing/SqlStatementTracker.java)


By the way, I noticed in ORM there's also a `MutationDelegateJoinedInheritanceTest`, shouldn't we port that too?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -138,6 +139,10 @@ public ResultSetAdaptor(RowSet<Row> rows, PropertyKind<Row> propertyKind, String
 		this( rows, rows.property( propertyKind ), idColumnName, idClass );
```
DavideD: It seems that this constructor is not used anymore and can be removed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/adaptor/impl/ResultSetAdaptor.java
```diff
@@ -150,6 +155,17 @@ private ResultSetAdaptor(RowSet<Row> rows, Row row, String idColumnName, Class<?
 		this.columnDescriptors = List.of( toColumnDescriptor( idClass, idColumnName ) );
```
DavideD: This constructor can become:

```java
	private ResultSetAdaptor(RowSet<Row> rows, Row row, String idColumnName, Class<?> idClass) {
		this( rows, row, List.of( idColumnName ), List.of( idClass ) );
	}
```

This way we avoid some duplication

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveAbstractReturningDelegate.java
```diff
@@ -6,133 +6,134 @@
...
 	@Override
-	default CompletionStage<GeneratedValues> reactivePerformInsertReturning(String sql, SharedSessionContractImplementor session, Binder binder) {
-		final String identifierColumnName = getPersister().getIdentifierColumnNames()[0];
-		final JdbcServices jdbcServices = session.getJdbcServices();
-		final String insertSql = createInsert( sql, identifierColumnName, jdbcServices.getDialect() );
+	public CompletionStage<GeneratedValues> reactivePerformInsertReturning(String sql, SharedSessionContractImplementor session, Binder binder) {
+		final String insertSql = createSqlString( sql, getGeneratedColumnNames(), session.getJdbcServices().getDialect() );
 		final Object[] params = PreparedStatementAdaptor.bind( binder::bindValues );
 		return reactiveExecuteAndExtractReturning( insertSql, params, session )
 				.thenApply( this::validateGeneratedIdentityId );
```
DavideD: I think we can skip this validation. 
`reactiveExecuteAndExtractReturning` already does it.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java
```diff
@@ -123,18 +101,40 @@ public GeneratedValues performMutation(
...
+		String sql = "";
+		for(int i = 0; i < generatedColumNames.size(); i++) {
+			if(i == generatedColumNames.size()-1 ){
+				sql += "inserted." + generatedColumNames.get( i );
+			}
+			else {
+				sql += "inserted." + generatedColumNames.get( i ) + ", ";
+			}
+		}
+		return sql;
```
DavideD: We can refactor this method:
```java
	private static String getSQLServerDialectInserted(List<String> generatedColumNames) {
        // Do we need this? Or do we already know that there's at least one generated column?
		if ( generatedColumNames.isEmpty() ) {
			return "";
		}
		String sql = "";
		for ( String generatedColumName : generatedColumNames ) {
			sql += ", inserted." + generatedColumName;
		}
		// Remove the initial comma
		return sql.substring( 2 );
	}
```

This way we don't need the `if-then-else` block in the loop.
Using a `StringBuilder` would remove a warning from IDEA, but I don't think it's necessary in this scenario.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
```diff
@@ -21,11 +31,39 @@ public class ReactiveIdentityGenerator extends IdentityGenerator {
 	/**
 	 * @see CockroachDBIdentityColumnSupport#supportsIdentityColumns() for some limitations related to CockroachDB
 	 */
+//	@Override
+//	public InsertGeneratedIdentifierDelegate getGeneratedIdentifierDelegate(EntityPersister persister) {
+//		Dialect dialect = persister.getFactory().getJdbcServices().getDialect();
+//		// Hibernate ORM allows the selection of different strategies based on the property `hibernate.jdbc.use_get_generated_keys`.
+//		// But that's a specific JDBC property and with Vert.x we only have one viable option for each supported database.
+//		return new ReactiveInsertReturningDelegate( persister, dialect );
+//	}
```
DavideD: You can remove this code if we don't need it anymore

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
```diff
@@ -21,11 +31,39 @@ public class ReactiveIdentityGenerator extends IdentityGenerator {
...
-		// But that's a specific JDBC property and with Vert.x we only have one viable option for each supported database.
-		return new ReactiveInsertReturningDelegate( persister, dialect );
+		final Dialect dialect = persister.getFactory().getJdbcServices().getDialect();
+		/*
+			Hibernate ORM allows the selection of different strategies based on the property `hibernate.jdbc.use_get_generated_keys`
+			but Vertex driver does not support get generated keys.
+			For db like Oracle and MySQL that don't support insert returning
+			it's still possible to have access to these values but for MySQL this is true only for one generated property.
+			SQLServerDialect does not fully support insert returning but in Hibernate Reactive
+			we are using `output` clause ( it does not work for columns using formulas or having a trigger).
```
DavideD: I don't think this is the right place to comment about the differences in MySQL and Oracle.

And I guess the comment about ignoring the value of `hibernate.jdbc.use_get_generated_keys` is not true anymore

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
```diff
@@ -21,11 +31,39 @@ public class ReactiveIdentityGenerator extends IdentityGenerator {
...
+		 */
+		final boolean supportsInsertReturning = ReactiveGeneratedValuesHelper.supportsInsertReturning( dialect );
+		if ( supportsInsertReturning && noCustomSql( persister, INSERT ) ) {
+			return new ReactiveInsertReturningDelegate( persister, dialect );
+		}
+		else if ( supportReactiveGetGeneratedKey( dialect, persister.getGeneratedProperties( INSERT ) ) ) {
+			return new ReactiveGetGeneratedKeysDelegate( persister, false, INSERT );
+		}
+		// Try to use generic delegates if the dialects supports them
```
DavideD: What do you mean? What's a `generic delegate`? I think you are trying to use the natural id if it makes sense, or you will run an extra query, right?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -241,6 +248,71 @@ public <T> CompletionStage<T> insertAndSelectIdentifier(String sql, Tuple parame
...
+				} );
+	}
+
+	private ResultSet getLastInsertedGeneratedValuesAsResultSet(
+			RowSet<Row> rows,
+			List<String> generatedColumnNames,
+			List<Class<?>> generatedValueClasses) {
+		final Long mySqlId = rows.property( MYSQL_LAST_INSERTED_ID );
+		if ( mySqlId != null ) {
+			// mysql vertex driver seems not supporting return of multiple generated values
```
DavideD: I would replace it with:
`The MySQL Vert.x driver does not seem to support returning multiple generated values.`

In particular, it's **Vert.x** (not vertex)

DavideD: Do we throw a meaningful exception if a user needs multiple generated values with MySQL?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
DavideD: Probably, you should move this comment into `ReactiveGeneratedValuesHelper#supportReactiveGetGeneratedKey`

DavideD: I wonder if we should log something if the user set the property and we are ignoring it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -223,7 +230,7 @@ public CompletionStage<ResultSet> insertAndSelectIdentifierAsResultSet(
 					RowIterator<Row> iterator = rows.iterator();
 					return iterator.hasNext()
 							? new ResultSetAdaptor( rows )
-							: getLastInsertedIdAsResultSet( rows, idClass, idColumnName );
+							: getLastInsertedGeneratedValuesAsResultSet( rows, List.of(idColumnName), List.of(idClass) );
```
DavideD: If I'm not mistaken, you can rewrite the whole method as:
```java

	@Override
	public CompletionStage<ResultSet> insertAndSelectIdentifierAsResultSet(
			String sql,
			Object[] parameters,
			Class<?> idClass,
			String idColumnName) {
		return executeAndSelectGeneratedValues( sql, parameters, List.of( idClass ), List.of( idColumnName ) );
	}
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java
```diff
@@ -7,51 +7,42 @@
...
 
 	private static final Log LOG = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
-	private final EntityPersister persister;
 	private final MutatingTableReference tableReference;
-	private final List<ColumnReference> generatedColumns;
 
 	public ReactiveInsertReturningDelegate(EntityPersister persister, EventType timing) {
-		this( persister, timing, false );
+		this( persister, timing, true );
```
DavideD, line 44: In this class there's an extra constructor that set the value to false for MySQL (see method [supportsArbitraryValues](https://github.com/hibernate/hibernate-reactive/blob/23b63a692e802d67ac177737d941e150d3f44364/hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java#L61)).

I think we can remove it, MySQL works fine now (probably because of the changes in this PR).

DavideD (commented):
Thanks a lot! Looks great.

I proposed some changes, but I think we could move it from draft to `Ready for review`.
Unless you think there's something missing that needs to be done before we can merge it.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutationDelegateIdentityTest.java
dreab8: I'll port `MutationDelegateJoinedInheritanceTest`, add the batch size to the tests and the sql tracker

DavideD: If you need an example, `BatchingConnectionTest` shows how to applies both

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java
dreab8: I think we don't need 
```
if ( generatedColumNames.isEmpty() ) {
	return "";
}
```

DavideD: OK, maybe let's add an assert then

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: > Do we throw a meaningful exception if a user needs multiple generated values with MySQL?

Maybe we don't need to... I'm wondering what happens in that scenario.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
dreab8: it comes from ORM, but I slightly changed the code and just forgot to remove it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
dreab8: in that case we are trying to use a delegate that uses `ReactiveUniqueKeySelectingDelegate` in case the id is generated otherwise no delegate that means executing a second query

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java
dreab8: probably better to add the assert [here](https://github.com/hibernate/hibernate-reactive/pull/2476/files#diff-6052b25d94a96d22d3a23645e4311db0c9f2799b02c189be42feda89bb2af3eaR111)

DavideD: OK

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Ok, no need to throw anything then

hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveIdentityGenerator.java
dreab8: I think this property in Reactive is just meaningless

DavideD: OK, let's ignore it silently for now

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveEntityIdentityInsertAction.java
```diff
@@ -108,6 +109,12 @@ private CompletionStage<Void> processInsertGeneratedProperties(
 			Object instance,
 			GeneratedValues generatedValues,
 			SharedSessionContractImplementor session) {
+		if ( persister.getRowIdMapping() != null ) {
+			rowId = generatedValues.getGeneratedValue( persister.getRowIdMapping() );
```
gavinking:
```suggestion
		final EntityRowIdMapping rowIdMapping = persister.getRowIdMapping();
		if ( rowIdMapping != null ) {
			rowId = generatedValues.getGeneratedValue( rowIdMapping );
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveEntityRegularInsertAction.java
```diff
@@ -111,8 +111,16 @@ private CompletionStage<Void> processInsertGeneratedProperties(
...
-					.thenAccept( v -> entry.postUpdate( instance, getState(), getVersion() ) );
-
+					.thenAccept( v -> {
+						// Process row-id values when available early by replacing the entity entry
+						if ( generatedValues != null && persister.getRowIdMapping() != null ) {
+							final Object rowId = generatedValues.getGeneratedValue( persister.getRowIdMapping() );
+							if ( rowId != null ) {
+								session.getPersistenceContext().replaceEntityEntryRowId( getInstance(), rowId );
+							}
+						}
```
gavinking:
```suggestion
						if ( generatedValues != null ) {
							final EntityRowIdMapping rowIdMapping = persister.getRowIdMapping();
							if (rowIdMapping != null ) {
								final Object rowId = generatedValues.getGeneratedValue( rowIdMapping );
								if ( rowId != null ) {
									session.getPersistenceContext().replaceEntityEntryRowId( getInstance(), rowId );
								}
							}
						}
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/engine/impl/ReactiveEntityUpdateAction.java
```diff
@@ -143,7 +143,16 @@ private CompletionStage<Void> processGeneratedProperties(
...
-			return persister.reactiveProcessUpdateGenerated( id, instance, getState(), generatedValues, session );
+			return persister.reactiveProcessUpdateGenerated( id, instance, getState(), generatedValues, session )
+					.thenAccept( v -> {
+						// Process row-id values when available early by replacing the entity entry
+						if ( generatedValues != null && persister.getRowIdMapping() != null ) {
+							final Object rowId = generatedValues.getGeneratedValue( persister.getRowIdMapping() );
+							if ( rowId != null ) {
+								session.getPersistenceContext().replaceEntityEntryRowId( getInstance(), rowId );
+							}
+						}
```
gavinking:
```suggestion
						if ( generatedValues != null ) {
							final EntityRowIdMapping rowIdMapping = persister.getRowIdMapping();
							if ( rowIdMapping != null ) {
								final Object rowId = generatedValues.getGeneratedValue( rowIdMapping );
								if ( rowId != null ) {
									session.getPersistenceContext().replaceEntityEntryRowId( getInstance(), rowId );
								}
							}
						}
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/insert/ReactiveInsertReturningDelegate.java
```diff
@@ -123,18 +89,41 @@ public GeneratedValues performMutation(
...
+			}
+			if ( sql.endsWith( "default values" ) ) {
+				sql = sql.substring( 0, sql.indexOf( "default values" ) );
+				sql = sql + "output " + getSQLServerDialectInserted( columnNames ) + " default values";
+			}
+			else {
+				sql = sql.replace( ") values (", ") output " + getSQLServerDialectInserted( columnNames ) + " values (" );
+			}
+		}
+		return sql;
```
gavinking: Would it be better to use a `StringBuilder` here, like we usually do?

dreab8: Hi @gavinking ,

 do you mean something like that?
```
              if ( dialect instanceof SQLServerDialect ) {
			assert !columnNames.isEmpty();
			StringBuilder builder = new StringBuilder( sql );
			int index = builder.lastIndexOf( " returning " + columnNames.get( 0 ) );
			// FIXME: this is a hack for HHH-16365
			if ( index > -1 ) {
				builder.delete( index, builder.length() );
			}
			int defaultValues = sql.indexOf( "default values" );
			if ( defaultValues > -1 ) {
				builder.delete( defaultValues, builder.length() );
				builder.append( "output " )
						.append( getSQLServerDialectInserted( columnNames ) )
						.append( " default values" );
			}
			else {
				int start = builder.lastIndexOf( ") values (" );
				builder.replace(
						start,
						start + 10,
						") output " + getSQLServerDialectInserted( columnNames ) + " values ("
				);
			}
			return builder.toString();
		}
		return sql;
```

gavinking: yeah ... wdyt?

dreab8: it works for me. Thanks

gavinking: OK, great, cheers!

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutationDelegateIdentityTest.java
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	private static void assertExecutedQueriesCount(int expected) {
+		assertThat( sqlTracker.getLoggedQueries().size() ).isEqualTo( expected );
+	}
+
+	private static void assertNumberOfOccurrenceInQueryNoSpace(int queryNumber, String toCheck, int expectedNumberOfOccurrences) {
+		String query = sqlTracker.getLoggedQueries().get( queryNumber );
+		int actual = query.split( toCheck, -1 ).length - 1;
+		Assert.assertThat( "number of " + toCheck, actual, is( expectedNumberOfOccurrences ) );
```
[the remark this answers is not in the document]
DavideD: Please, use AssertJ (`org.assertj.core.api.Assertions.assertThat`) for the assertions

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MutationDelegateTest.java
```diff
@@ -0,0 +1,309 @@
...
+					}
+				} )
+			  )
+		);
+	}
+
+	private static void assertNumberOfOccurrenceInQueryNoSpace(int queryNumber, String toCheck, int expectedNumberOfOccurrences) {
+		String query = sqlTracker.getLoggedQueries().get( queryNumber );
+		int actual = query.split( toCheck, -1 ).length - 1;
+		Assert.assertThat( "number of " + toCheck, actual, is( expectedNumberOfOccurrences ) );
```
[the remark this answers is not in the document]
DavideD: Please, use AssertJ (org.assertj.core.api.Assertions.assertThat) for the assertions

PR #2479, opened 2025-08-30
build: improve perf parse of versions plugin

kingg22:
Hi, based on [configuration cache valid inputs](https://docs.gradle.org/current/userguide/configuration_cache.html#build_configuration_inputs) I do a refactor on versions plugins in your local plugins.

Other suggestions, but out of scope of this PR:

In root `build.gradle` can use libs, explicit () and "":
```groovy
plugins {
    id( "local.versions" )

    id( "java-library" )
    id( "maven-publish" )
    alias( libs.plugins.com.diffplug.spotless )
    alias( libs.plugins.org.asciidoctor.jvm.convert ) apply false
}

group = "org.hibernate.reactive"
// leverage the ProjectVersion which comes from the `local.versions` plugin
version = project.projectVersion.fullName

rootProject.afterEvaluate {
    logger.lifecycle( "ORM version: ${libs.versions.hibernateOrmVersion.get()}" )
    logger.lifecycle( "ORM Gradle plugin version: ${libs.versions.hibernateOrmGradlePluginVersion.get()}" )
    logger.lifecycle( "Vert.x SQL Client version: ${libs.versions.vertxSqlClientVersion.get()}" )
}
```

In root `gradle.properties` adding those reduce to 30s build time, similar to [hibernate orm](https://github.com/hibernate/hibernate-orm/blob/main/gradle.properties#L11) (properties, not build time xD)
```toml
# asciidoctor gradle plugin and local plugins don't support configuration cache
org.gradle.configuration-cache=false
# run test in parallel, caution with it **
org.gradle.parallel=true
org.gradle.caching=true
org.gradle.daemon=true
```

I recommend use version catalog for jdks version, like [hibernate orm](https://github.com/hibernate/hibernate-orm/blob/main/settings.gradle#L59) do.

Without further ado, if you can try it and give me feedback :)


local-build-plugins/src/main/java/org/hibernate/reactive/env/VersionsPlugin.java
```diff
@@ -28,137 +25,144 @@ public class VersionsPlugin implements Plugin<Project> {
...
+
+    public static ProjectVersion determineProjectVersion(
+            Project project,
+            ProjectVersion releaseVersion,
+            Provider<String> projectVersionString
+    ) {
+        if ( releaseVersion != null ) {
+            return releaseVersion;
+        }
+        // Si no hay release explícito, usamos el valor del archivo (vía ValueSource)
```
DavideD: Comments should be in english

DavideD:
Thanks a lot, much better. I decided to apply a few changes, can you have a look at them and let me know what you think?
They are in this commit: https://github.com/DavideD/hibernate-reactive/commit/5414a2b3b1bccac95916cfa1771b2da05c3fc200

Nothing major, just some stylistic changes and some refactoring to remove duplicated code.

PR #2504, opened 2025-09-11
[4.2] Add automatic module name to the manifest 

marko-bekhta:
I also moved the `ReactivePersistenceContextAdapter` out of `org.hibernate.engine.internal` so that there won't be a "split package thing" and it seems all worked ok (at least locally except some links in the javadocs of that class, which is ... ok)

- Fixes https://github.com/hibernate/hibernate-reactive/issues/2503


yrodiere (commented):
LGTM, let's merge and backport the second commit (about wrong package being used)?

yrodiere (approved):

PR #2513, opened 2025-09-16
[4.2] Better error message when transparently loading lazy fields

### LINKED ISSUE #1770: Check error message when lazy loading fields
It's related to this issue: https://github.com/quarkusio/quarkus/issues/36211
The error message is quite cryptic and I'm not sure if it's an issue in Quarkus or Hibernate Reactive.

We should check if this is an error in Hibernate Reactive and, if that's the case,  provide more useful feedback to the user.

dreab8:
Fix #1770 

This PR aims to intercept direct access to lazy fields in order to throw a better error message, see [ReactiveEnhancementAsProxyLazinessInterceptor](https://github.com/hibernate/hibernate-reactive/compare/main...dreab8:hibernate-reactive:issue_1770_last?expand=1#diff-57132104a550f7514bb9c4ee1a97c5fa92e863d47e30581b944754b60dcbf825R34) and [ReactiveLazyAttributeLoadingInterceptor](https://github.com/hibernate/hibernate-reactive/compare/main...dreab8:hibernate-reactive:issue_1770_last?expand=1#diff-11857c516ed1f79e0c08f43fef4e4525a2cce4d644d0300b4313273347e9351eR34)

This PR needs https://github.com/hibernate/hibernate-orm/pull/10950

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
@@ -18,6 +18,7 @@ import org.hibernate.TypeMismatchException;
 import org.hibernate.UnknownEntityTypeException;
 import org.hibernate.UnresolvableObjectException;
 import org.hibernate.bytecode.enhance.spi.interceptor.EnhancementAsProxyLazinessInterceptor;
+import org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor;
 import org.hibernate.collection.spi.PersistentCollection;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.reactive.engine.impl.ReactivePersistenceContextAdapter;
@@ -359,8 +360,20 @@ public class ReactiveSessionImpl extends SessionImpl implements ReactiveSession,
 
 	@Override
 	public <E, T> CompletionStage<T> reactiveFetch(E entity, Attribute<E, T> field) {
-		return ( (ReactiveEntityPersister) getEntityPersister( null, entity ) )
-				.reactiveInitializeLazyProperty( field, entity, this );
+		final ReactiveEntityPersister entityPersister = (ReactiveEntityPersister) getEntityPersister( null, entity );
+		LazyAttributeLoadingInterceptor lazyAttributeLoadingInterceptor = entityPersister.getBytecodeEnhancementMetadata()
+				.extractInterceptor( entity );
+		final String attributeName = field.getName();
+		if ( !lazyAttributeLoadingInterceptor.isAttributeLoaded( attributeName ) ) {
+			return ( (CompletionStage<T>) lazyAttributeLoadingInterceptor.fetchAttribute( entity, field.getName() ) )
+					.thenApply( value -> {
+						lazyAttributeLoadingInterceptor.attributeInitialized( attributeName );
+						return value;
+					} );
+		}
+		else {
+			return completedFuture( (T) entityPersister.getPropertyValue( entity, attributeName ) );
+		}
 	}
 
 	@Override


integration-tests/bytecode-enhancements-it/src/test/java/org/hibernate/reactive/it/LazyBasicFieldTest.java
```diff
@@ -47,4 +50,46 @@ public void testFetchBasicField(VertxTestContext context) {
...
+					  .chain( () -> getMutinySessionFactory().withSession( session -> {
+						  Crew crew = session.getReference( Crew.class, emily.getId() );
+						  String role = crew.getRole();
+						  return session.flush();
+					  } ) )
+			  ).invoke( exception -> assertThat( exception.getMessage() ).contains( "Reactive sessions do not support transparent lazy fetching" ) )
+		);
+	}
+
+	public static <U extends Throwable> Uni<U> assertThrown(Class<U> expectedException, Uni<?> uni) {
```
DavideD, line 146: Don't we already have this in [ReactiveAssertions#assertThrown](https://github.com/hibernate/hibernate-reactive/blob/82340d1dfb1ae6c074cae226c2ff04c17b25b40b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/testing/ReactiveAssertions.java#L65)?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
```diff
@@ -270,6 +270,9 @@ public interface Log extends BasicLogger {
 	@Message(id = 84, value = "The application requested a JDBC connection, but Hibernate Reactive doesn't use JDBC. This could be caused by a bug or the use of an unsupported feature in Hibernate Reactive")
 	SQLException notUsingJdbc();
 
+	@Message(id = 85, value = "Reactive sessions do not support transparent lazy fetching - use Session.fetch() (property '%1$S' of entity '%2$s' was not loaded)")
```
DavideD: Could you also mention `Mutiny.fetch` in the error message?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -359,8 +360,20 @@ else if ( isPersistentAttributeInterceptable( association ) ) {
...
+		final String attributeName = field.getName();
+		if ( !lazyAttributeLoadingInterceptor.isAttributeLoaded( attributeName ) ) {
+			return ( (CompletionStage<T>) lazyAttributeLoadingInterceptor.fetchAttribute( entity, field.getName() ) )
+					.thenApply( value -> {
+						lazyAttributeLoadingInterceptor.attributeInitialized( attributeName );
+						return value;
+					} );
+		}
+		else {
+			return completedFuture( (T) entityPersister.getPropertyValue( entity, attributeName ) );
```
DavideD, line 375: If I understand correctly, this change will prevent an extra query. Could you make sure we test for it?
Otherwise, future changes might remove the improvement.

DavideD (commented):
Thanks a lot,
a couple of things caught my eyes while giving it a very quick pass.

I will have a better look as soon as I finished with all the other non-draft PRs

integration-tests/bytecode-enhancements-it/src/test/java/org/hibernate/reactive/it/LazyBasicFieldTest.java
dreab8: not in this module

dreab8: maybe we can copy it from `hibernate-reactive-core`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/Log.java
dreab8: do you mean something like 
```
 "Reactive sessions do not support transparent lazy fetching - use Stage.fetch() or Mutiny.fetch() (property '%1$S' of entity '%2$s' was not loaded)"
`
?

DavideD: Yes, that works. Thanks

integration-tests/bytecode-enhancements-it/src/test/java/org/hibernate/reactive/it/LazyBasicFieldTest.java
DavideD: Never mind, this is fine then. I didn't notice it was in a separate module. Sorry.
One day we will move the testing utilities in a separate module.

PR #2514, opened 2025-09-18
ReactivePersistenceContextAdapter does not need to be in an ORM package

gavinking:


gavinking:
> #2504 🤷🏻 🙂

OK, great, perfect. But why not just split that PR into two bits and get the bit which isn't to do with module descriptors integrated right away?

DavideD:
Thanks.
Closing this one because I've already merged @marko-bekhta  PR.

PR #2531, opened 2025-09-24
[4.2] Migrate to release scripts for documentation publishing

### LINKED ISSUE #2519: Migrate to release scripts for documentation publishing
Currently it’s done through custom Gradle tasks, but that gets in the way of changing where we publish our docs for all projects at the same time.

See https://github.com/hibernate/hibernate-release-scripts/blob/main/upload-documentation.sh
See use in Hibernate Search for example.

See https://hibernate.atlassian.net/browse/HHH-19800 for a similar task in Hibernate ORM.

dreab8:
Fix #2519

### DIFF
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -211,7 +211,7 @@ pipeline {
 								string(credentialsId: 'release.gpg.passphrase', variable: 'JRELEASER_GPG_PASSPHRASE'),
 								string(credentialsId: 'Hibernate-CI.github.com', variable: 'JRELEASER_GITHUB_TOKEN')
 						]) {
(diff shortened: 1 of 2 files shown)


DavideD (commented):
I don't know if this is the right approach.
We already have a task that copies the documentation from one place to another.
It's here: https://github.com/hibernate/hibernate-reactive/blob/21a019cac705f1d314145d130a951c38740229bb/release/build.gradle#L140

I'm not sure exactly what we need to do here, but it seems we could just copy the documentation in the right place during the release. RIght?

marko-bekhta:
right ... you've got a point here 🙂, utill we update the release script, we'd just copy a few extra files somewhere for no reason 🫣 😃
we just wanted to prepare the changes for all projects before merging

DavideD:
OK, I'm going to merge these. This way we don't have to deal with future rebase and conflicts. Feel free to create a new issue to keep track of what's left to do when the other part of the release scripts is done

PR #2561, opened 2025-10-01
[#2560] Apply the unified Hibernate Documentation theme

marko-bekhta:
- Fixes https://github.com/hibernate/hibernate-reactive/issues/2560

This PR applies both Javadoc and Asciidoc themes 😃, to not repeat everything, here are a few additional notes:
- https://github.com/hibernate/hibernate-orm/pull/11049#issue-3473798017

also maybe it's worth to restructure the doc a bit so that there is not just one top level section:

image
image

Issue #1990, opened 2024-09-12
Consider moving away from subprojects block for subproject configuration


sebersole:
Using `subprojects {}` from build is considered an anti-pattern.  Actually, referencing any project from another is consider an anit-pattern.  It forces all of the projects to be evaluated and configured even if running tasks in just one.

In ORM, e.g., we instead use common script fragments applied to subprojects as appropriate.

An even better option is local plugins to accomplish this - I have just not done this yet with ORM which is more complex.

PR #2565, opened 2025-10-02
[3.1] Apply the unified Hibernate Documentation theme

DavideD:
Backport #2560 (PR #2561) to 3..1


DavideD:
I haven't backported #1990, so I cannot apply the commit as is.

PR #2690, opened 2025-10-28
[4.2] Open connections lazily

### LINKED ISSUE #2518: Lazy connection opening / transaction starting
I would like a way for Hibernate Reactive to open its connections lazily, instead of when the session gets created.

## Motivation

First, opening the connection eagerly makes some patterns problematic: if the user opens a session, then starts some long-running external process, and only uses the session when that process finishes, they'll be holding a connection for a long time. Same goes for transactions.
That (among others, I'm sure) is the reason Hibernate ORM opens connections only when they are first needed, and I believe this holds for Hibernate Reactive.

Second, opening the connection on session creation means `openSession` must return a `Uni<Mutiny.Session>` (or `CompletionStage`). This makes it hard (impossible?) to implement request-scoped or transaction-scoped sessions as described [here](https://github.com/quarkusio/quarkus/issues/47462#issuecomment-2820982896).
To be precise, we _could_ implement it, but users would be forced to first call a reactive method on the `Session` before they can do anything else. Or to inject the `Session` as a `Uni<Session>`.

## Implementation

I would suggest we expose a `openSessionWithLazyConnection()` method (bikeshedding needed) on `SessionFactory`, which would return a `Session` instead of `Uni<Session>`.

Internally, that session would only create a new connection when it first needs to access the database.

I believe the implementation would share a lot of ideas with #2494, in particular the idea of having a field in the session that represents the "last operation" -- in our case it would rather be the connection opening / transaction starting.

DavideD:
Fix #2518 

This PR add the following methods to the Mutiny and Stage API:
```java
Session openSessionWithLazyConnection();
Session openSessionWithLazyConnection(String tenantId);
StatelessSession openStatelessSessionWithLazyConnection();
StatelessSession openStatelessSessionWithLazyConnection(String tenantId);
```
I've annotated them with `@Incubating`.

Overall, it seems to work fine. But I decided to add a test that mimic the id generation and, if I don't use transactions, it seems to get stuck (even with a low number of ids). I don't know why. The same test seems to work fine when the connection is not opened lazily. It's possible that there's something wrong with the test.
I've applied the changes that make the test fail on a different branch: https://github.com/hibernate/hibernate-reactive/commit/b511efaf634779538b5204978d8b43c70fbfade6

Basically, if I call `persist` and then `flush`, instead of `withTransaction`, the test get stuck (the test is `MultithreadedInsertionWithLazyConnectionTest`)

`openSessionWithLazyConnection` seems a bit of a mouthful but I don't have strong opinions about it.
Originally, it was called only `createSession`.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultithreadedInsertionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultithreadedInsertionTest.java
@@ -162,10 +162,12 @@ public class MultithreadedInsertionTest {
 					.whenComplete( (o, throwable) -> {
 						endLatch.reached();
 						if ( throwable != null ) {
+							prettyOut( throwable.getMessage() );
 							startPromise.fail( throwable );
 						}
 						else {
 							if ( !initialThreadName.equals( Thread.currentThread().getName() ) ) {
+								prettyOut( "Thread switch detected. Expecting " + initialThreadName + ", actual " + Thread.currentThread().getName() );
 								startPromise.fail( "Thread switch detected!" );
 							}
 							else {


hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -1971,6 +1972,59 @@ interface Transaction {
...
+		 * Reactive.
+		 * <p>
+		 * The underlying database connection is obtained lazily
+		 * when the returned {@link Session} needs to access the
+		 * database.
+		 * <p>
+		 * The client must close the session using {@link Session#close()}.
+		 */
+		@Incubating
+		Session openSessionWithLazyConnection();
```
gavinking: We need to come up with better names for these operations.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -186,4 +212,170 @@ private SqlClientConnection newConnection(SqlConnection connection) {
...
+		}
+
+		@Override
+		public boolean isTransactionInProgress() {
+			return connection != null && connection.isTransactionInProgress();
+		}
+
+		@Override
+		public DatabaseMetadata getDatabaseMetadata() {
+			Objects.requireNonNull( connection, "Database metadata not available until the connection is opened" );
```
DavideD, line 249: Note that I'm throwing a `NullPointerException` here for now if they try to `getDatabaseMetadata` without a connection

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
DavideD: I can rename them `createSession` (the same names we used in the past). It should work

gavinking: That's fine by me; at least, I don't have a better idea.

DavideD:
After looking into it, I figured out that the test [I added for the id generation](https://github.com/hibernate/hibernate-reactive/commit/b511efaf634779538b5204978d8b43c70fbfade6) fails because the event loop thread changes during the execution when using the lazy connection approach. The test was getting stuck because of an error in the [session.close method](https://github.com/hibernate/hibernate-reactive/pull/2700). 

@yrodiere what should we do? Merge these changes and fixed the issues later? Or wait until we figure out how to solve this?
To summarize:
* [The original id generation test](https://github.com/hibernate/hibernate-reactive/blob/25256a135d40f2b5c55999b3b691e332d1efaf1f/hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultithreadedInsertionTest.java#L184): the connection is open as soon  as the session is created; we insert entities using persist and flush operations (without opening a transaction);
* [The new test for this issue that fails](https://github.com/DavideD/hibernate-reactive/commit/1c0d1d6d108410f368ee82eafaab8bbd56c778c3): the connection is open lazily; it fails with `HR000069: Detected use of the reactive Session from a different Thread than the one which was used to open the reactive Session`; it works fine if we use transactions

@tsegismont, maybe you can help with this?

gavinking:
I doubt we can remove this sort of check.

We need to be sure that we're always being called with the same Vert.x duplicated context, otherwise we open ourselves up to very subtle bugs.

DavideD:
Thanks a lot @tsegismont,
another thing that we were wondering is if this is happening because we are using `CompletionStage`. I've recreated the same test with Mutiny and it seemed to work fine (but I had to reduce the number of thread because we don't have the same infrastracture to prevent StackOverflow errors at the moment).

DavideD:
> I think I just found the issue (in `org.hibernate.reactive.id.impl.BlockingIdentifierGenerator`). I'll double-check after lunch and will share my findings (hopefully with a fix)

Thanks a lot!

PR #2732, opened 2025-11-04
[4.2] Implemented Reactive versions of PostAction, PreAction and JdbcSelectWithActions

### LINKED ISSUE #2651: Re-enable FindByIdWithLockTest.testFindUpgradeNoWait
We disabled it during [the Hibernate ORM 7.2.0.CR1 upgrade](#2640 ).

https://github.com/hibernate/hibernate-reactive/blob/1adce185e648c07f90432a86dd7d7841056841f2/hibernate-reactive-core/src/test/java/org/hibernate/reactive/FindByIdWithLockTest.java#L89

For branches: `4.2` and `3.2`

dreab8:
Fix #2651 

The PR implements the Reactive version of  `JdbcSelectWithActions` and of the new `PostAction` and `PreAction` (`ReactiveLockTimeoutHandler` and `ReactiveLockTimeoutHandler`).

The PR does not implement the Reactive version of `CollectionLockingAction` but in order to use it a Reactive version of   `org.hibernate.query.SelectionQuery#setLockScope(Locking.Scope lockScope);` method need to be added.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
@@ -22,6 +22,7 @@ import org.hibernate.reactive.sql.results.spi.ReactiveListResultsConsumer;
 import org.hibernate.reactive.util.impl.CompletionStages;
 import org.hibernate.sql.ast.tree.select.SelectStatement;
 import org.hibernate.sql.exec.internal.BaseExecutionContext;
+import org.hibernate.sql.exec.internal.JdbcOperationQuerySelect;
 import org.hibernate.sql.exec.internal.lock.EntityDetails;
 import org.hibernate.sql.exec.internal.lock.TableLock;
 import org.hibernate.sql.results.graph.DomainResult;
@@ -94,11 +95,12 @@ public class ReactiveTableLock extends TableLock {
 		final var jdbcServices = sessionFactory.getJdbcServices();
 		final var selectStatement = new SelectStatement( querySpec, domainResults );
 
+		final JdbcOperationQuerySelect jdbcSelect = jdbcServices.getDialect().getSqlAstTranslatorFactory()
+				.buildSelectTranslator( sessionFactory, selectStatement )
+				.translate( jdbcParameterBindings, lockingQueryOptions );
 		return StandardReactiveSelectExecutor.INSTANCE
 				.list(
-						jdbcServices.getDialect().getSqlAstTranslatorFactory()
-								.buildSelectTranslator( sessionFactory, selectStatement )
-								.translate( jdbcParameterBindings, lockingQueryOptions ),
+						jdbcSelect,
 						jdbcParameterBindings,
 						// IMPORTANT: we need a "clean" ExecutionContext to not further apply locking
 						new BaseExecutionContext( session ),
@@ -107,7 +109,7 @@ public class ReactiveTableLock extends TableLock {
 						ReactiveListResultsConsumer.UniqueSemantic.ALLOW
 				).thenCompose( results -> {
 					if ( isEmpty( results ) ) {
-						throw new AssertionFailure( "Expecting results" );
+						throw new AssertionFailure( "Expecting results from table locking query : '" + jdbcSelect.getSqlString() + "', but none were returned" );
 					}
 					return CompletionStages.loop( results, row -> {
 						final var entityDetails = entityDetailsMap.get( row[0] );


hibernate-reactive-core/src/test/java/org/hibernate/reactive/FindByIdWithLockTest.java
```diff
@@ -167,7 +165,7 @@ public static class Child {
 
 		public String name;
 
-		@ManyToOne
+		@ManyToOne(fetch = FetchType.LAZY)
```
DavideD: I don't think you should change the mapping of an existing test. ALso, does it matter if you use lazy here?

dreab8: With the new Lock implementation in ORM, it affects the number of queries executed.
I'm going to revert this change and modify the `testFindUpgradeNoWait` assertions accrdingly

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveConnectionLockTimeoutStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+import jakarta.persistence.Timeout;
+import java.lang.invoke.MethodHandles;
+import java.sql.Connection;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Reactive version of {@link ConnectionLockTimeoutStrategy}
+ */
+public interface ReactiveConnectionLockTimeoutStrategy extends ConnectionLockTimeoutStrategy {
+	Log LOG = LoggerFactory.make( Log.class, MethodHandles.lookup() );
```
DavideD: I know that we use this pattern in some places, but 
I think it's better if we don't put the LOG in the interface and override the method in the implemetations instead, even if it means we have some duplication.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/LockHelper.java
```diff
@@ -0,0 +1,80 @@
...
+			Function<Timeout, Integer> valueStrategy,
+			String sqlFormat,
+			ReactiveConnection connection) {
+		final int milliseconds = valueStrategy.apply( timeout );
+		return setLockTimeout( milliseconds, sqlFormat, connection );
+	}
+
+	@FunctionalInterface
+	public interface TimeoutExtractor {
+		CompletionStage<Timeout> extractFrom(ReactiveConnection.Result resultSet);
```
DavideD: Are you sure this needs to return a `CompletionStage`? All the implementations I see return a constant.

DavideD: I did a little refactoring and it seems you can just return a `Timeout` object: https://github.com/DavideD/hibernate-reactive/commit/f76290d7a075a865b0d5b9933d29af6fa49e8555

Feel free to squash the commit in your PR, if you agree with the changes

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveCockroachConnectionLockTimeoutStrategyImpl.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	@Override
+	public CompletionStage<Timeout> getReactiveLockTimeout(
+			ReactiveConnection connection,
+			SessionFactoryImplementor factory) {
+		return LockHelper.getLockTimeout(
+				"show lock_timeout",
+				(resultSet) -> {
+					// see https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_lock_wait_timeout
```
DavideD: I think you copied and paste the link: it links to the MySQL docs but it's in the code for the Cockroach strategy


The same error happens in all the other implementations

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveCockroachConnectionLockTimeoutStrategyImpl.java
```diff
@@ -0,0 +1,74 @@
...
+				},
+				connection
+		);
+	}
+
+	public CompletionStage<Void> setReactiveLockTimeout(
+			Timeout timeout,
+			ReactiveConnection connection,
+			SessionFactoryImplementor factory) {
+		return LockHelper.setLockTimeout(
```
DavideD: This code is the same as the one in Postgres. If Cockroach calls the one in Postgres, we remove some duplicated code.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.concurrent.CompletionStage;
+
+import static org.hibernate.Hibernate.isEmpty;
+import static org.hibernate.reactive.util.impl.CompletionStages.nullFuture;
+
+/**
+ * Reactive version of {@link TableLock}
+ */
+public class ReactiveTableLock extends TableLock {
+	Log LOG = LoggerFactory.make( Log.class, MethodHandles.lookup() );
```
DavideD: this `LOG` should be `private static final`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
```diff
@@ -0,0 +1,194 @@
...
+								.translate( jdbcParameterBindings, lockingQueryOptions ),
+						jdbcParameterBindings,
+						// IMPORTANT: we need a "clean" ExecutionContext to not further apply locking
+						new BaseExecutionContext( session ),
+						row -> row,
+						Object[].class,
+						ReactiveListResultsConsumer.UniqueSemantic.ALLOW
+				).thenCompose( results -> {
+					if ( isEmpty( results ) ) {
+						throw new AssertionFailure( "Expecting results" );
```
DavideD: Can't we add anything else to the error message? What should have the query returned? Something related to a lock I imagine.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
```diff
@@ -0,0 +1,194 @@
...
+			if ( stateValue == null ) {
+				if ( !toOne.isNullable() ) {
+					throw new IllegalStateException( "Retrieved key was null, but to-one is not nullable : " + toOne.getNavigableRole()
+							.getFullPath() );
+				}
+				else {
+					reference = null;
+				}
+				applyLoadedState( entityDetails, statePosition, reference );
+				applyModelState( entityDetails, statePosition, reference );
```
DavideD: `reference` is always `null`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	protected static class ReactiveToOneResultHandler extends ToOneResultHandler implements ReactiveResulHandler {
+
+		public ReactiveToOneResultHandler(Integer statePosition, ToOneAttributeMapping toOne) {
+			super( statePosition, toOne );
+		}
+
+		@Override
+		public CompletionStage<Void> reactiveApplyResult(
```
DavideD, line 166: Unless I'm missing something, nothing in this method seems reactive. Do you need to return a `CompletionStage`?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
```diff
@@ -0,0 +1,194 @@
...
+				Object stateValue,
+				EntityDetails entityDetails,
+				ReactiveSessionImpl session) {
+			final Object reference;
+			if ( stateValue == null ) {
+				if ( !toOne.isNullable() ) {
+					throw new IllegalStateException( "Retrieved key was null, but to-one is not nullable : " + toOne.getNavigableRole()
+							.getFullPath() );
```
DavideD: Shouldn't this have a error code in the Log?

DavideD (changes requested):
Thanks @dreab8, 
it looks good overall, but there some small things to change.

DavideD:
I upgraded main to ORM 7.2.0.CR2, could you rebase it there, please?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/sql/exec/internal/lock/ReactiveTableLock.java
dreab8: `session.reactiveInternalLoad()` is a reactive method

DavideD: ok

PR #2745, opened 2025-11-07
[4.2] Simplify the release process

marko-bekhta:
Hey Davide 🙂 👋🏻 
I'm opening this as a draft for now. We'll need to merge the scripts update first. 

the idea behind this iteration of release updates is .... project builds should only build the artifacts (e.g. jars/javadocs/html docs etc) and put them in the specific location. Then the release scripts would do the rest. 
This means that the project build does not need to do any git operations or deployments or ... 😃

### DIFF
--- a/github_release_notes.md
+++ b/github_release_notes.md
@@ -1,0 +1,3 @@
+
+* See the [website](https://hibernate.org/reactive/releases/{{releaseVersionFamily}}) for requirements and compatibilities.
+* See the [What's New](https://hibernate.org/reactive/releases/{{releaseVersionFamily}}/#whats-new) guide for details about new features and capabilities.
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -217,7 +217,8 @@ pipeline {
 								withEnv([
 										"DISABLE_REMOTE_GRADLE_CACHE=true"
 								]) {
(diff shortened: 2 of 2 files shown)


documentation/build.gradle
```diff
@@ -182,6 +182,9 @@ def renderReferenceDocumentationTask = tasks.register( "renderReferenceDocumenta
 			"html-meta-version-family": versionFamily.get(),
 	)
 
+	// See https://docs.asciidoctor.org/gradle-plugin/latest/common-task-configuration/#choosing-an-execution-mode-for-asciidoctorj
+	executionMode = org.ysb33r.grolifant.api.core.jvm.ExecutionMode.JAVA_EXEC
```
marko-bekhta, line 186: this one is a bonus 😄 
we don't really need it for releases or anything ... but it helps with asciidoc building 🙂

DavideD:
Nice, thanks.

github_release_notes.md
```diff
@@ -0,0 +1,3 @@
+
+* See the [website](https://hibernate.org/reactive/releases/{{releaseVersionFamily}}) for requirements and compatibilities.
+* See the [What's New](https://hibernate.org/reactive/releases/{{releaseVersionFamily}}/#whats-new) guide for details about new features and capabilities.
```
marko-bekhta, line 3: The idea is that each project/branch can add something extra about a particular release if it wants to.
This template is then included in the reactive template https://github.com/hibernate/hibernate-release-scripts/blob/b5852b88067e9b71fd746a69917a2b92576ce7f1/jreleaser/github-release-note-template/reactive.tpl 

but if you'd want something different for your github releases we can discuss that 🙂

Issue #2736, opened 2025-11-04
Handle/prevent event loop switching in Hibernate Reactive


yrodiere:
After #2690, it appears that under some conditions, execution might suddenly move to a different thread / event loop. This has only been observed when opening connections lazily.

This breaks a core assumption that we verify here:

https://github.com/hibernate/hibernate-reactive/blob/2fffcce1f8bcde80e73d13de8f999832ec83f6cb/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java#L201-L203

... and results in failures as described there:

https://github.com/hibernate/hibernate-reactive/pull/2690#issuecomment-3479643373

We have two options.

## Find out why we're switching event loops

If we consider that event loop switching is unexpected / unacceptable, then should work on understanding why we're switching event loops exactly.

@tsegismont may be able to help here but we will need to provide more information.

I personally suspect that connection opening (sometimes?) leads to event loop switching, but I cannot back that claim at the moment, and @DavideD mentioned that the problem happens on session closing, which would prove me wrong.

## Revise our checks

It's been mentioned in the past that our checks [are too strict](https://github.com/quarkusio/quarkus/issues/32533#issuecomment-1505639669). 

I believe what we actually want to check is that:

1. We're not using a session with a different context active than the one used to create the session.
2. We're never using the same session from multiple event loops _at the same time_. (Perhaps not the initial intent, but we're effectively checking it and it's a nice bonus).
 
Both should generally be the case, but then users could mistakenly end up using a `Session` instance on some worker thread, or passing it around to different reactive chains in exotic ways. And it would result in hard-to-debug issues.

#2494 will make the second check pretty much irrelevant by making sure we never execute two session operations at the same time.

For the context, we could perhaps replace the current check based on the thread, with one based on the context? For instance:

1. Store the current context on session creation, and check before every operation that it's active (`ContextInternal#isRunningOnContext()`).
2. Store the session token (`session.getSessionToken()`) in the context on session creation, and check before every operation that the token is present in the context . Note we'd have to use a `Set`, because multiple sessions (e.g. from multiple PUs) can be active at the same time.

The second solution seems more flexible because it would allow duplicating the context during the session's lifetime, which is something we're planning to do to implement https://github.com/quarkusio/quarkus/issues/47698.

gavinking:
Oh I get it I had forgotten that worker threads are only intended for blocking operations. Carry on. :)

[...]

PR #2793, opened 2025-11-18
[4.2] Prevent context switching during id generation

### LINKED ISSUE #2768: A connection can be opened in the wrong Context
Follows https://github.com/hibernate/hibernate-reactive/pull/2690

@tsegismont  [already provided solution](https://github.com/hibernate/hibernate-reactive/pull/2690#issuecomment-3513163421) but it needs to be converted into a proper fix for Hibernate Reactive.

DavideD:
Fix #2768 

Follows https://github.com/hibernate/hibernate-reactive/pull/2690 and it's a refactoring of @tsegismont [suggestion](https://github.com/hibernate/hibernate-reactive/pull/2690#issuecomment-3513163421).

* I changed the warning about the swithching context into [assertions](https://github.com/hibernate/hibernate-reactive/commit/9dce31a9a503b8a3f3ac5189540677da7c41b799#diff-a3210bb1373db55b4feb0d8c4031314d1117c6747351f5cd267f8cc65ab8ab63R350:~:text=InternalStateAssertions%2EassertCurrentContextMatches%28%20this%2C%20connectionContext%20%29%3B)
* I had to change where we read the context when a new connection is created. The problem is that the session factory can be used in a worker thread at start up, and therefore the context could be null. [This is the line that saves the context](https://github.com/hibernate/hibernate-reactive/compare/main...DavideD:hibernate-reactive:2768-Connection-context#diff-9aa5cbca30d703f6ba7e72d257b7e5a9acf0ef42035f1787ab3743aba45137deR226:~:text=newConnection)
* [ProxyConnection](https://github.com/DavideD/hibernate-reactive/blob/e4fdfd4e99f0158f365f67d15816566f47c58d21/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java#L233) synchronize the creation of a connection using reflection. I don't know if this can be a problem with Quarkus.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BatchingConnectionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BatchingConnectionTest.java
@@ -8,11 +8,13 @@ package org.hibernate.reactive;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.cfg.AvailableSettings;
 import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.mutiny.Mutiny;
 import org.hibernate.reactive.mutiny.impl.MutinySessionImpl;
 import org.hibernate.reactive.mutiny.impl.MutinyStatelessSessionImpl;
 import org.hibernate.reactive.pool.BatchingConnection;
 import org.hibernate.reactive.pool.ReactiveConnection;
 import org.hibernate.reactive.pool.impl.SqlClientConnection;
+import org.hibernate.reactive.stage.Stage;
 import org.hibernate.reactive.stage.impl.StageSessionImpl;
 import org.hibernate.reactive.stage.impl.StageStatelessSessionImpl;
 import org.hibernate.reactive.testing.SqlStatementTracker;
@@ -83,6 +85,56 @@ public class BatchingConnectionTest extends ReactiveSessionTest {
 		}
 		assertThat( actualConnection.getClass().getName() )
 				.isEqualTo( org.hibernate.reactive.pool.impl.SqlClientPool.class.getName() + "$ProxyConnection" );
+	}
+
+	@Test
+	public void testBatchingWithPersistAllAndProxyConnection(VertxTestContext context) {
+		Stage.Session session = getSessionFactory().createSession();
+		test( context, session
+				.persist(
+						new GuineaPig( 11, "One" ),
+						new GuineaPig( 22, "Two" ),
+						new GuineaPig( 33, "Three" )
+				)
+				// Auto-flush
+				.thenCompose( v -> session
+						.createSelectionQuery( "select name from GuineaPig", String.class )
+						.getResultList()
+						.thenAccept( names -> {
+							assertThat( names ).containsExactlyInAnyOrder( "One", "Two", "Three" );
+							assertThat( sqlTracker.getLoggedQueries() ).hasSize( 1 );
+							// Parameters are different for different dbs, so we cannot do an exact match
+							assertThat( sqlTracker.getLoggedQueries().get( 0 ) )
+									.startsWith( "insert into pig (name,version,id) values " );
+							sqlTracker.clear();
+						} )
+				)
+		);
+	}
+
+	@Test
+	public void testBatchingWithPersistAllAndProxyConnectionAndMutiny(VertxTestContext context) {
+		Mutiny.Session session = getMutinySessionFactory().createSession();
+		test( context, session
+				.persistAll(
+						new GuineaPig( 11, "One" ),
+						new GuineaPig( 22, "Two" ),
+						new GuineaPig( 33, "Three" )
+				)
+				// Auto-flush
+				.chain( v -> session
+						.createSelectionQuery( "select name from GuineaPig", String.class )
+						.getResultList()
+						.invoke( names -> {
+							assertThat( names ).containsExactlyInAnyOrder( "One", "Two", "Three" );
+							assertThat( sqlTracker.getLoggedQueries() ).hasSize( 1 );
+							// Parameters are different for different dbs, so we cannot do an exact match
+							assertThat( sqlTracker.getLoggedQueries().get( 0 ) )
+									.startsWith( "insert into pig (name,version,id) values " );
+							sqlTracker.clear();
+						} )
+				)
+		);
 	}
 
 	@Test


hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/InternalStateAssertions.java
```diff
@@ -52,4 +54,13 @@ public static void assertCurrentThreadMatches(Thread expectedThread) {
 		}
 	}
 
+	public static void assertCurrentContextMatches(Object object, ContextInternal expectedContext, Exception creationTrace) {
```
yrodiere: With this new assertion, do we still need `assertCurrentThreadMatches`? If not, fixing #2736 would be trivial...

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -123,12 +125,18 @@ public CompletionStage<ReactiveConnection> getConnection(String tenantId, SqlExc
 	}
 
 	private CompletionStage<ReactiveConnection> getConnectionFromPool(Pool pool) {
-		return completionStage( pool.getConnection().map( this::newConnection ), ReactiveConnection::close );
+		final Exception creationTrace = new Exception();
```
yrodiere: This is something I wouldn't have done out of concern performance would be bad.

Is this wise?
Is it already done elsewhere?

At the very least I'd avoid creating this exception if `InternalStateAssertions.ENFORCE` is `false`. Though that won't help Quarkus.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -205,18 +213,38 @@ private <T> CompletionStage<T> completionStage(Future<T> future, Consumer<T> onC
...
 	}
 
 	private static class ProxyConnection implements ReactiveConnection {
+
+		private static final VarHandle OPENED_HANDLE;
+
+		static {
+			try {
+				MethodHandles.Lookup lookup = MethodHandles.lookup();
+				OPENED_HANDLE = lookup.findVarHandle( ProxyConnection.class, "opened", boolean.class );
```
yrodiere, line 241: This is new to me, but it looks like a common pattern in `CompletableFuture`. I guess the idea is to get access to `compareAndSet` without the overhead of `AtomicBoolean`. Whether this overhead is significant in this specific case is quite questionable if you ask me, but I assume Thomas knows better than I do. 

In any case, back to your question @DavideD: this kind of "reflection" in a `static` block should generally be fine in Quarkus even with native compilation, because it's executed during static init, before native compilation. If I turn out to be wrong, we'll improvise :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -356,13 +390,8 @@ public CompletionStage<Void> rollbackTransaction() {
...
-			if ( connection == null ) {
-				this.batchSize = batchSize;
-			}
-			else {
-				connection = connection.withBatchSize( batchSize );
-			}
+		public ProxyConnection withBatchSize(int batchSize) {
+			connectionFuture.thenApply( reactiveConnection -> reactiveConnection.withBatchSize( batchSize ) );
 			return this;
 		}
```
yrodiere: This seems wrong?

The method is supposed to return a batched connection, and from what I can see we rely on the return value:

https://github.com/hibernate/hibernate-reactive/blob/066917f1b35c39b81ddfd3793ea75336cd418220/hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java#L1753-L1757

Returning `this` will not work because we don't update the delegate, so if the caller uses `this`, they'll still be using a non-batching connection.

I'll assume the contract is that `this` can be ignored after this method is called, and only the return value will be used by the caller. Which seems consistent with what I'm seeing in `BatchedConnection#withBatchSize()`, where we brutally change the batch size of an existing instance.

If that assumption is correct, I'd suggest replacing this code with something like:

```suggestion
		public ProxyConnection withBatchSize(int batchSize) {
		  ReactiveConnection reactiveConnection = connectionFuture.getNow( null );
		  if (reactiveConnection != null) {
			// Common case (hopefully): connection was already opened,
			// we can let callers use the delegate going forward and forget about the proxy.
		    return reactiveConnection.withBatchSize(batchSize);
		  }
		  else {
			// Uncommon case (hopefully): connection wasn't opened yet,
			// we will ask callers to use a new proxy that wraps the connection.
			return new ProxyConnection( () -> connectionSupplier.get().withBatchSize( batchSize ) );
		  }
		}
```

NOTE: If you add something to mark the proxy as closed (see my comment on `close`), it would be a good idea mark the current proxy as closed in the `else` block here, just in case someone (incorrectly) goes against the assumption that I stated above.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -373,8 +402,8 @@ public CompletionStage<Void> executeBatch() {
 
 		@Override
 		public CompletionStage<Void> close() {
-			return connection != null
-					? connection.close().thenAccept( v -> connection = null )
+			return opened
+					? connectionFuture.getNow( null ).close()
 					: voidFuture();
```
yrodiere, line 430: Shouldn't we remember that `close` was called in the proxy, so that if another method is called on the proxy, we don't mistakenly create the delegate?

I'm talking about the second case in particular:

1. Create connection proxy
2. Call close(): nothing happens
3. Call`execute(...)`: delegate gets created! Instead of erroring out.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultithreadedInsertionWithLazyConnectionTest.java
```diff
@@ -197,8 +197,13 @@ private CompletionStage<Void> storeEntity(Stage.Session s) {
 			final EntityWithGeneratedId entity = new EntityWithGeneratedId();
 			entity.name = beforeOperationThread + "__" + localVerticleOperationSequence;
 
+			// We are not using transactions on purpose here, because this approach will cause a context switch
+			// and an assertion error if things aren't handled correctly. See Hibernate Reactive issue #2768:
+			// https://github.com/hibernate/hibernate-reactive/issues/2768
```
yrodiere: Would it make sense to duplicate the test, to _also_ test the "transaction" case?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/BlockingIdentifierGenerator.java
```diff
@@ -137,35 +134,38 @@ public Task execute(GeneratorState state) {
...
+						}
+						else {
+							//We ignore the state argument as we actually use the field directly
+							//for convenience, but they are the same object.
+							executor.submit( stateIgnored -> {
+								result.complete( next( newlyGeneratedHi ) );
+								return null;
+							} );
+						}
+					} );
```
yrodiere: Are we sure `nextHiValue` will never throw an exception? I mean the method, not the future it returns.

Because if it does throw an exception, the `whenComplete` call here will be skipped and the operation will hang forever from the user's point of view.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
tsegismont: The idea was that, since the application will create a lot of these objects during its lifetime, it would be better to store a simple `boolean` field in `ProxyConnection` instead of a complete object (own headers + field).

@franz1981 would you recommend against it?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: I wasn't 100% sure about it either. 
Let's remove it for now, I left it in a separate commit for this reason.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: > In any case, back to your question @DavideD: this kind of "reflection" in a `static` block should generally be fine in Quarkus even with native compilation, because it's executed during static init, before native compilation. If I turn out to be wrong, we'll improvise :)

Thanks, that's all I needed in writing :D

hibernate-reactive-core/src/test/java/org/hibernate/reactive/MultithreadedInsertionWithLazyConnectionTest.java
DavideD: Yeah, it makes sense.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/common/InternalStateAssertions.java
DavideD: I don't know :)

Seems like it, but I feel that this PR is complicated enough. I will look into it as soon as we merge this.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: Could you review [this commit](https://github.com/hibernate/hibernate-reactive/pull/2793/commits/5c245f90b1d3e2a270b0dcc22f09380e124c0275), please? Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -355,15 +399,22 @@ public CompletionStage<Void> rollbackTransaction() {
...
 		public ReactiveConnection withBatchSize(int batchSize) {
-			if ( connection == null ) {
-				this.batchSize = batchSize;
+			if ( closed ) {
+				throw LOG.connectionIsClosed();
+			}
+
+			if ( connectionFuture.isDone() ) {
+				// connection exists, we can let callers use the delegate and forget about the proxy.
+				return connectionFuture.getNow( null ).withBatchSize( batchSize );
```
yrodiere, line 409: Note the pattern `if ( ... .isDone() ) { get() }` should be avoided in general, because it's sensitive to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use). That's why I suggested `var value = getNow(null); if ( value != null)` .

Though in this case... I think the worst that could happen is a missed opportunity, so it's probably fine.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -355,15 +399,22 @@ public CompletionStage<Void> rollbackTransaction() {
...
 			}
 			else {
-				connection = connection.withBatchSize( batchSize );
+				return new ProxyConnection( () -> opened
```
yrodiere, line 412: So the `opened` check here is supposed to handle a case where someone does something like this?

```
<trigger connection request>
connection = connection.withBatchSize(...); // Connection not yet initialized somehow?
connection.doSomething();
```

That seems odd... did you do this because you noticed it actually happening?

In any case, +1 to be on the safe side.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
yrodiere: I assume you meant https://github.com/hibernate/hibernate-reactive/pull/2793/commits/6b8a2f19d0704fe41643b9538e0efcffe1c1b0bd ?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: It's to be on the safe side. 

Looking at the code, it seems to me there's a small window between setting `opened` to `true` and completing `completableFuture` with the connection. So, if `opened` it's true and `completableFuture` is not done, `.getNow(null)` will return `null` (and cause a `NullPointerException`), even if the connection is getting opened.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: What I'm afraid of using `getNow(null)` in this case, it's that we will return null a bit too soon.

In this specific case, when `.isDone()` returns true, we should know for a fact that there's a result that we can return.

When it returns `false`, it's possible that by the time we call `.getNow(null)`, the result is not null and we have created a `ProxyConnection` even if we didn't need to, but my guess is that this use case won't happen often.

DavideD: > Though in this case... I think the worst that could happen is a missed opportunity, so it's probably fine.

I didn't notice this line. Yes, that's what I was thinking.
The initial solution could have potentially caused a `NullPointerException` instead.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: It's a bit late, I think I will sleep over it and merge it tomorrow morning if there are no other comments.
After that I will start the release.

Thanks

PR #2872, opened 2025-12-02
[4.2] Replace deprecated ::set-output in the workflows

marko-bekhta:
Hey 🙂 👋🏻 

I was looking at the way you've setup the scheduler for the workflows and noticed the warnings: https://github.com/hibernate/hibernate-reactive/actions/runs/19874264989, so with that ... I thiiiiiiiiiiiiink this should address that 🙂


DavideD:
Rebased and merged locally

PR #2895, opened 2025-12-08
Include a branch name (from the input) when generating artifact names

marko-bekhta:
Hi Davide 🙂 👋🏻 

I was looking at the https://github.com/hibernate/hibernate-reactive/actions/runs/20035758488 and noticed that sometimes we could fail to upload the test results on failure (if the same group failed for a different branch already and we've uploaded the report already: https://github.com/hibernate/hibernate-reactive/actions/runs/20035758488/job/57456696049)
so maybe we can jsut append the branch name there and make them more "unique"

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,6 +91,14 @@ jobs:
         run: |
           echo "yearmonth=$(/bin/date -u '+%Y-%m')" >> "$GITHUB_OUTPUT"
         shell: bash
+      - name: Sanitize the branch name
+        # We want to include the branch in artifact names to keep them unique.
+        #  Branch names may have some symbols that are not "friendly" for file names e.g. we may have a branch `wip/4.0`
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -91,6 +91,12 @@ jobs:
         run: |
           echo "yearmonth=$(/bin/date -u '+%Y-%m')" >> "$GITHUB_OUTPUT"
         shell: bash
+      - name: Sanitize the branch name
+        # So that if we end up using the branch name in the artifact name we won't have any unexpected chars in the name
```
DavideD: Thanks for the change, it makes sense. But I don't think this comment it's that helpful.
Could you mention instead that we use it for distinguishing reports for different branches, and maybe add an example of the conversion.  For example, `wip/4.2` becomes `wip-4.2`.

PR #2928, opened 2025-12-18
[4.2] Validate transaction and roll back them in case of error 

### LINKED ISSUE #2926: Roll back live transaction when closing the connection
Follows this quarkus issue: https://github.com/quarkusio/quarkus/issues/51576

DavideD:
Fix #2926

There are two places where we check if a transaction is already open: `SqlConnection#beginTransaction()` and `SqlConnection#close()`

The open transaction will be roll backed in both cases.

I wasn't sure how to recreate the [Quarkus use case](https://github.com/quarkusio/quarkus/issues/51576) using only Hibernate Reactive, but I tested it locally and this solution fixes the problem.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/NoLiveTransactionValidationErrorTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/NoLiveTransactionValidationErrorTest.java
@@ -1,0 +1,189 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.reactive.pool.ReactiveConnection;
+import org.hibernate.reactive.stage.Stage;
+import org.hibernate.reactive.stage.impl.StageSessionImpl;
+import org.hibernate.reactive.util.impl.CompletionStages;
+
+import org.junit.jupiter.api.Test;
+
+import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.reactive.testing.ReactiveAssertions.assertThrown;
+
+public class NoLiveTransactionValidationErrorTest extends BaseReactiveTest {
+
+	@Override
+	protected Collection<Class<?>> annotatedEntities() {
+		return List.of( Comic.class );
+	}
+
+	@Test
+	public void beginTransactionError(VertxTestContext context) {
+		final Stage.Session[] session = { null };
+		test( context, assertThrown( IllegalStateException.class, getSessionFactory()
+				.openSession()
+				.thenCompose( s -> {
+					session[0] = s;
+					ReactiveConnection connection = reactiveConnection( s );
+					return connection.beginTransaction()
+							.thenCompose( t1 -> connection.beginTransaction() );
+				} )
+			  )
+				.thenAccept( e -> {
+					assertThat( e )
+							.hasMessageContaining( "HR000091" )
+							.hasMessageContaining( "Can't begin a new transaction" );
+				} )
+				.handle( CompletionStages::handle )
+				// We try to close the session we opened
+				.thenCompose( assertionHandler -> close( session[0] )
+						.thenCompose( assertionHandler::getResultAsCompletionStage ) )
+		);
+	}
+
+	// Try to close the session ignoring any error
+	private static CompletionStage<Void> close( Stage.Session session) {
+		return closeSession( session ).handle( (unused, throwable) -> null );
+	}
+
+	private static ReactiveConnection reactiveConnection(Stage.Session s) {
+		return ( (StageSessionImpl) s ).getReactiveConnection();
+	}
+
+	@Test
+	public void rollbackOnCloseWithStage(VertxTestContext context) {
+		Comic beneath = new Comic( "979-8887241081", "Beneath The Trees Where Nobody Sees" );
+
+		test(
+				context,
+				assertThrown( IllegalStateException.class, getSessionFactory()
+						.withTransaction( s -> s
+								.persist( beneath )
+								.thenCompose( v -> s.flush() )
+								// Close the connection before committing
+								.thenCompose( v -> s.close() )
+						)
+				)
+						.thenAccept( e -> assertThat( e )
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/test/java/org/hibernate/reactive/NoLiveTransactionValidationErrorTest.java
```diff
@@ -0,0 +1,175 @@
...
+								.persist( beneath )
+								.call( s::flush )
+								// Close the connection before committing
+								.call( s::close )
+						)
+				)
+						.invoke( e -> assertThat( e )
+								.hasMessageContaining( "HR000090" )
+								.hasMessageContaining( "closing the connection" )
+						)
```
yrodiere, line 135: Could you assert that there is no active connection after this? Since that was the main problem :)

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NoLiveTransactionValidationErrorTest.java
yrodiere, line 1: I'm surprised you couldn't also test the Quarkus scenario -- AFAIK you'd just have to do something similar to `CancelSignalTest` and assert that the number of active connections is 0 after execution?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -362,52 +367,86 @@ private SqlConnection client() {
...
+								closed = true;
+								LOG.tracef( "Connection closed: %s", connection );
+							}
+							else {
+								LOG.tracef( "Connection was already closed: %s", connection );
+							}
+							// Connection closed, return the result of the validation
+							return validationHandler.getResultAsCompletionStage();
+						} )
+				);
```
yrodiere: There's... a lot to unpack here, and I wish we had an easier way to express all this. IIUC this is mostly to suppress exceptions?

Anyway, as I can see, this does what we want.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -362,52 +367,86 @@ private SqlConnection client() {
...
 	}
 
-	private void clearTransaction(Void v, Throwable x) {
-		LOG.tracef( "Clearing current transaction instance from connection: %s", transaction );
-		transaction = null;
+	/**
+	 * If there's a transaction open, roll back it and return a failed CompletionStage.
+	 * @param operation It will be included in the error message
+	 */
+	private CompletionStage<Void> validateLiveTransaction(String operation) {
```
yrodiere: Looks to me like this should be named `validateNoLiveTransaction`, given the "no live transaction" state is what we want?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
```diff
@@ -362,52 +367,86 @@ private SqlConnection client() {
...
 	public CompletionStage<Void> beginTransaction() {
-		if ( transaction != null ) {
-			throw new IllegalStateException( "Can't begin a new transaction as an active transaction is already associated to this connection" );
-		}
-		return connection.begin()
-				.onSuccess( tx -> LOG.tracef( "Transaction started: %s", tx ) )
-				.onFailure( v -> LOG.errorf( "Failed to start a transaction: %s", transaction ) )
-				.toCompletionStage()
-				.thenAccept( this::setTransaction );
+		return validateLiveTransaction( "starting a new transaction" )
```
yrodiere: I don't understand why a call to `beginTransaction` on a connection where a transaction already exists should trigger a rollback... IMO it should just fail, and that failure will inevitably trigger closing the connection, which itself will roll back the transaction.

I agree the difference is minimal, and I expect any use case taking advantage of it (e.g. to "start a transaction if there isn't one already") would be exotic. Still, that seems more sensible to me?

yrodiere (approved):
Thanks. Feel free to merge, but I have a few comments below -- especially about the behavior of `beginTransaction`...

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Yes, it's basically a try-catch-finally block. We need to make sure that we are still closing the connection when there's a validation error.  If closing the connection fails, I add the validation error as a suppressed exception. This way we don't hide errors. 

Maybe if there's a validation error, but I can still close the connection, I should mention in the error message that the connection was closed

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Sure, make sense :)

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientConnection.java
DavideD: Yes, OK. make sense.
I will change this part. Because, it was the same type of error, I thought we could do the same.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/NoLiveTransactionValidationErrorTest.java
DavideD: I don't have the same metrics when I use vanilla Hibernate Reactive. I think Quarkus has an extra dependency that keep track of it. I will have another quick look to see if I missed something yesterday

DavideD: I mean,  the test in Quarkus is checking `sql_pool_active`, but I can only check `vertx.sql.processing.pending`. 
I think I didn't add `vertx.sql.processing.pending`, it because I could never see it go to `1`. But I will have another quick look today and see if I can think of something.

One Idea I had was to add a quick quarkus-it module to test these type of things. But I didn't have time to follow through.

yrodiere: I see -- I didn't know this metric was specific to Quarkus. Thanks.

In that case +1 to merge without this extra test -- yours should be enough -- so that _maybe_ we can release something today...

PR #2943, opened 2025-12-22
[4.3] add xxxxMultiple() operations to Mutiny.Session

gavinking:


DavideD:
> Need to add `@since` tags to all that, but I dunno what the next version is @DavideD ?

Next version family will be `4.3`

DavideD:
Thanks

Issue #1890, opened 2024-04-19
Missing tests for `@SoftDelete`


DavideD:
We have created a test for [the support of `@SoftDelete`](https://docs.jboss.org/hibernate/orm/6.5/userguide/html_single/Hibernate_User_Guide.html#soft-delete): [SoftDeleteTest](https://github.com/hibernate/hibernate-reactive/blob/4650538d32fb9e8095a24ae373c8b5397f8edaac/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SoftDeleteTest.java#L44)

The problem is that we only test entities that are mapped on a single table in the database. And, we don't test the use of `@SoftDelete` [on associations](https://docs.jboss.org/hibernate/orm/6.5/userguide/html_single/Hibernate_User_Guide.html#soft-delete-entity).

We need additional tests with entities mapped using the different [inheritance types](https://docs.jboss.org/hibernate/orm/6.5/userguide/html_single/Hibernate_User_Guide.html#entity-inheritance).

As a guideline for first timers, to solve this issue one has to:
- [ ] Rename SoftDeleteTest to SoftDeleteSingleTableTest
- [ ] Mimic the existing test, but change the entities so that they use `@Inheritance(strategy = InheritanceType.JOINED)`
- [ ] Mimic the existing test, but change the entities so that they use `@Inheritance(strategy = InheritanceType.TABLE_PER_CLASS)`
- [ ] Test `@SoftDelete` applied to a collection (might be a separate test class or a test method added to the previous one)
- [ ] Maybe it would helpful to also add a test for the other converters (now we only test the `YesNoConverter`)

I've collected everything in a single issue, but we could split this task in several issues when required.

AmbarMishra973:
i am intrested in this issue can you assign it to me

DavideD:
Thanks, I've assigned it to you @AmbarMishra973

hoyeonyy:
Hi, I'm interested in working on this issue. 
is this issue still open? If so, I'd like to work on it.

PR #3013, opened 2026-01-23
Add comprehensive tests for @SoftDelete annotation (#1890)

hoyeonyy:
Add comprehensive tests for @SoftDelete annotation #1890

### Change
1. Renamed existing test
- **Renamed** `SoftDeleteTest` → `SoftDeleteSingleTableTest` to clarify that it tests
  single-table entity mappings (entities without inheritance)

2. Added JOINED inheritance strategy test
- Tests `@SoftDelete` behavior with `@Inheritance(strategy = InheritanceType.JOINED)`

3. Added collection relationship test
- Tests that soft-deleted entities are properly filtered from `@OneToMany` collections
- Uses `join fetch` to load collections with soft delete filter applied

4. Added converter types test
- Tests different converter implementations: `YesNoConverter`, `TrueFalseConverter`, and
  default boolean converter
- Tests both `ACTIVE` and `DELETED` strategies

5. Added TABLE_PER_CLASS inheritance strategy test (disabled)
- Tests are written but currently disabled due to potential issues with `@SoftDelete`
  support for TABLE_PER_CLASS inheritance
- When soft delete is performed via HQL, entities are not properly marked as deleted in the
  database
- Needs investigation to determine if this is a test implementation issue or an unsupported
  feature in Hibernate ORM/Reactive

---

  ### TABLE_PER_CLASS Issue
  The `SoftDeleteTablePerClassTest` is included but disabled because I'm uncertain whether the
   test failures are due to:
  1. Incorrect test implementation on my part
  2. An actual limitation/bug in Hibernate ORM's `@SoftDelete` support for TABLE_PER_CLASS
  inheritance

  The test is well-documented with comments explaining the issue. If this is a known
  limitation, the test can be removed. If it's a bug, the test can serve as a reproduction
  case once the underlying issue is fixed.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SoftDeleteCollectionTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SoftDeleteCollectionTest.java
@@ -16,6 +16,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import io.vertx.junit5.VertxTestContext;
+import jakarta.persistence.CascadeType;
 import jakarta.persistence.Entity;
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.Id;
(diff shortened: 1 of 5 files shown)


DavideD:
Thanks a lot, I decided to apply some changes and sent a new PR with some extra commits: https://github.com/hibernate/hibernate-reactive/pull/3183

DavideD:
I will create a new issue for the table per class case

PR #3014, opened 2026-01-23
[4.3] Add transaction coordinator

DavideD:
Allow transactions to be handled outside of Hibernate Reactive.

The only breaking change is the addition of `ReactiveConnection#getTransactionCoordinator`. The problem is that
`ReactiveConnection` is a public interface. but:
* it's annotated with `@Incubating`
* the method is handy but not required (I think), we could remove it.
I kept those changes in a separate commit: 5a57b5d250a898e1e5258fffd4053d19f0e5e3ac

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
@@ -300,6 +300,13 @@ public abstract class SqlClientPool implements ReactiveConnectionPool {
 		}
 
 		@Override
+		public ReactiveTransactionCoordinator getTransactionCoordinator() {
+			return Objects
+					.requireNonNull( connectionFuture.getNow( null ), "Transaction coordinator not available until a connection has been created" )
+					.getTransactionCoordinator();
+		}
+
+		@Override
 		public DatabaseMetadata getDatabaseMetadata() {
 			if ( closed ) {
 				throw LOG.connectionIsClosed();
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ExternalTransactionCoordinatorTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/ExternalTransactionCoordinatorTest.java
@@ -48,24 +48,28 @@ public class ExternalTransactionCoordinatorTest extends BaseReactiveTest {
 				// Get connection directly from the pool (simulating external transaction manager)
 				factoryManager.getReactiveConnectionPool()
 						.getConnection()
-						.thenCompose( connection -> connection
-								// External manager begins transaction
-								.beginTransaction()
-								// Open session using this externally-managed connection
-								.thenCompose( v -> getMutinySessionFactory()
-										.openSession()
-										.chain( session -> session
-												.persist( beneath )
-												.call( session::flush )
-												// Close session (should NOT throw exception, should NOT rollback)
-												.call( session::close )
-										).subscribeAsCompletionStage()
-								)
-								// External manager commits after session is closed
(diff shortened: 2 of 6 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
```diff
@@ -299,6 +299,13 @@ public boolean isTransactionInProgress() {
 			return reactiveConnection != null && reactiveConnection.isTransactionInProgress();
 		}
 
+		@Override
+		public ReactiveTransactionCoordinator getTransactionCoordinator() {
+			return Objects
+					.requireNonNull( connectionFuture.getNow( null ), "Transaction coordinator not available until a connection has been created" )
+					.getTransactionCoordinator();
```
yrodiere, line 306: This, more than the breaking change to the public interface, makes me think tying the transaction coordinator to the connection could be a problem eventually... I can't provide an objective reason though, it's just a hunch.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/ReactiveTransactionCoordinator.java
```diff
@@ -0,0 +1,38 @@
...
+public interface ReactiveTransactionCoordinator {
+
+	/**
+	 * Indicates whether transactions are being handled externally to Hibernate Reactive.
+	 * <p>
+	 * The external coordinator is responsible for:
+	 * <ul>
+	 *     <li>Beginning transactions before providing connections</li>
+	 *     <li>Committing or rolling back transactions on session close</li>
+	 *     <li>Properly closing physical connections</li>
```
yrodiere, line 32:
```suggestion
	 *     <li>Intercepting Hibernate Reactive's call to {@code connection.close()} to avoid actually closing the physical connection</li>
	 *     <li>Properly closing physical connections after commit or roll-back</li>
```

hibernate-reactive-core/src/test/java/org/hibernate/reactive/ExternalTransactionCoordinatorTest.java
```diff
@@ -0,0 +1,148 @@
...
+		public boolean isExternallyManaged() {
+			return true;
+		}
+	}
+
+	/**
+	 * Test connection pool that uses the external transaction coordinator.
+	 * Extends DefaultSqlClientPool and only overrides the transaction coordinator.
+	 */
+	static class TestExternalConnectionPool extends DefaultSqlClientPool {
```
yrodiere, line 122: I'm surprised this test even passes...

I would expect this pool to require overrides so that:

1. It doesn't actually close connections when Hibernate Reactive calls `connection.close()` (otherwise how can the external "transaction manager" commit after closing the session?)
2. It somehow reuses the connection you created manually in the test method (otherwise it seems to me Hibernate Reactive would just use a second connection?)

But maybe I missed some important hack...

DavideD: I think you are right, I will check this

DavideD: Yep, I focused on skipping the validation on close and missed that problem. I will work on it

hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/SqlClientPool.java
DavideD: Yeah, I'm not very happy about this either. I will have another look at it as soon as I fix the general issue.

PR #3072, opened 2026-02-01
[4.3] Replace the deprecated Logger.getMessageLogger overload

### LINKED ISSUE #3134: Fix: use non-deprecated method of JBoss Logger and pass Locale.ROOT
Usage of JBoss's `Logger.getMessageLogger` to use the non-deprecated and recommended method based on `MethodHandles.Lookup`.

The change is minimal: Hibernate Reactive already used lookup; this simply adjusts the final usage with JBoss to avoid warnings.

See https://hibernate.atlassian.net/browse/HHH-20115 performance with `Locale.ROOT` and skip find the correct Locale to use.
If localized log messages are introduced in the future, this must be revisited and the explicit ROOT locale removed.

kingg22:
Hii HR team, this PR corrects the use of JBoss's `Logger.getMessageLogger` to use the non-deprecated and recommended method based on `MethodHandles.Lookup`.

The change is minimal: Hibernate Reactive already used lookup; this simply adjusts the final usage with JBoss to avoid warnings.

Please let me know if you hadn't changed this before for any reason. If you need to create a JIRA ticket, that's fine; and if you prefer to cherry-pick this specific change, that's also fine.

The other warnings come from Hibernate ORM (internal to Hibernate, not external), so you probably know better how to handle them; I'm just clarifying this for context :D

Closes #3134

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/LoggerFactory.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/logging/impl/LoggerFactory.java
@@ -5,6 +5,7 @@
 package org.hibernate.reactive.logging.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Locale;
 
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/util/impl/CompletionStages.java
```diff
@@ -21,13 +21,14 @@
...
 
+import static java.lang.invoke.MethodHandles.lookup;
 import static org.hibernate.reactive.util.async.impl.AsyncIterator.fromIterator;
 import static org.hibernate.reactive.util.async.impl.AsyncIterator.range;
 import static org.hibernate.reactive.util.async.impl.AsyncTrampoline.asyncWhile;
 
 public class  CompletionStages {
 
-	private static final Log LOG = LoggerFactory.make( Log.class, new LogCategory( "org.hibernate.reactive.errors" ) );
+	private static final Log LOG = LoggerFactory.make( Log.class, new LogCategory( "org.hibernate.reactive.errors" ), lookup() );
```
marko-bekhta, line 31: hey @DavideD  👋🏻 🙂 
we probably would want something like this: https://hibernate.atlassian.net/browse/HHH-20115 for Hibernate Reactive as well, right ?

DavideD: Yes, sounds good. Thanks

kingg22: Hey, I've added the suggestion with comments in case it needs to be changed in the future. 🙂

PR #3196, opened 2026-02-20
Add Jacoco coverage reports and Sonar analysis

marko-bekhta:
Hey Davide 🙂 👋🏻 

I'll open it as a draft for now, let's get the one for Hibernate Models in first 🙂 (in case there would be any changes ....):
* https://github.com/hibernate/hibernate-models/pull/202


DavideD:
I will start the release anyway, we can fix it later

Issue #3107, opened 2026-02-06
Test CurrentTimestamp annotation


DavideD:
We don't have any tests for [@CurrentTimestamp](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/annotations/CurrentTimestamp.java)

roguebee-droid:
Would this test better reside in hibernate-orm rather than hibernate-reactive?

DavideD:
No, this issue is Hibernate Reactive specific. ORM should already have it's own tests

PR #3223, opened 2026-02-24
Added tests for CurrentTimestamp

roguebee-droid:
@DavideD Give it a go, tell me what you think. I'm unsure whether it really needs the embedded test as well.

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
@@ -11,6 +11,7 @@ import java.util.Collection;
 import java.util.List;
 
 import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.CurrentTimestamp;
 import org.hibernate.annotations.UpdateTimestamp;
 
 import org.junit.jupiter.api.Test;
@@ -18,7 +19,6 @@ import org.junit.jupiter.api.Test;
 import io.vertx.junit5.Timeout;
 import io.vertx.junit5.VertxTestContext;
 import jakarta.persistence.Basic;
-import jakarta.persistence.Column;
 import jakarta.persistence.Embeddable;
 import jakarta.persistence.Embedded;
 import jakarta.persistence.Entity;
@@ -27,6 +27,7 @@ import jakarta.persistence.Id;
 
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.generator.EventType.INSERT;
 
 @Timeout(value = 10, timeUnit = MINUTES)
 public class TimestampTest extends BaseReactiveTest {
@@ -45,6 +46,9 @@ public class TimestampTest extends BaseReactiveTest {
 						.chain( session::flush )
 						.invoke( () -> assertThat(
 								record.updated.truncatedTo( ChronoUnit.HOURS )
+						).isEqualTo( record.created.truncatedTo( ChronoUnit.HOURS ) ) )
+						.invoke( () -> assertThat(
+								record.currentTime.truncatedTo( ChronoUnit.HOURS )
 						).isEqualTo( record.created.truncatedTo( ChronoUnit.HOURS ) ) )
 						.invoke( () -> record.text = "edited text" )
 						.chain( session::flush )
@@ -81,6 +85,7 @@ public class TimestampTest extends BaseReactiveTest {
 	private static void assertInstants(Record r) {
 		assertThat( r.created ).isNotNull();
 		assertThat( r.updated ).isNotNull();
+		assertThat( r.currentTime).isNotNull();
 		// Sometimes, when the test suite is fast enough, they might be the same
 		assertThat( r.updated )
 				.as( "Updated instant is before created. Updated[" + r.updated + "], Created[" + r.created + "]" )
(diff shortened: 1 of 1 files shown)


DavideD (changes requested):
Thanks a lot!

I need you to apply a few changes:

* The prefix for the commit message should be `[#3107]` (and not `[3107]`). The `#` is important because it's how GitHub recognizes that it's an issue (and it will render it accordingly). 

* I think for `@CurrentTimestamp` it makes sense to test the default case and the insert, so that you can compare the two value  and check that after an update, the default case is bigger or equal than the created case
 
*  Could you also add a test for `@CurrentTimestamp` with `@Version`, please? https://docs.hibernate.org/orm/7.3/userguide/html_single/#locking-optimistic-timestamp

* I know that they were already there, but we could remove all the `@Column` annotations

hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
```diff
@@ -90,11 +95,13 @@ private static void assertInstants(Record r) {
 	private static void assertInstants(Event e, History h) {
 		assertThat( e.history.created ).isNotNull();
 		assertThat( e.history.updated ).isNotNull();
+		assertThat( e.currentTimestampEmbedded.current ).isNotNull();
 		// Sometimes, when the test suite is fast enough, they might be the same:
 		assertThat( e.history.updated )
 				.as( "Updated instant is before created. Updated[" + e.history.updated + "], Created[" + e.history.created + "]" )
 				.isAfterOrEqualTo( e.history.created );
 		assertThat( e.history.created ).isEqualTo( h.created );
+		assertThat( e.currentTimestampEmbedded.current ).isEqualTo( h.created );
```
DavideD: It's a good idea, but I don't think this is going to work unless you truncate the values. The two fields aren't really updated with the same value. 

By the way, if you add two fields with `@CurrentTimeStamp` (one for inserts and one for every other change),  you can compare the two and check that the insert one comes  before the other.

roguebee-droid:
Right, think I've done everything now. I'll go make that bug report for you on the @Version annotation.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
```diff
@@ -90,12 +95,18 @@ private static void assertInstants(Record r) {
 	private static void assertInstants(Event e, History h) {
 		assertThat( e.history.created ).isNotNull();
 		assertThat( e.history.updated ).isNotNull();
+		assertThat( e.currentTimestampEmbedded.currentCreatedAt ).isNotNull();
+		assertThat( e.currentTimestampEmbedded.currentLastUpdatedAt ).isNotNull();
 		// Sometimes, when the test suite is fast enough, they might be the same:
 		assertThat( e.history.updated )
-				.as( "Updated instant is before created. Updated[" + e.history.updated + "], Created[" + e.history.created + "]" )
+				.as( "Updated instant is after created. Updated[" + e.history.updated + "], Created[" + e.history.created + "]" )
```
DavideD: The `.as("...")` message will appear when the assertion fails. It will fails when `.updated` is before `.created`. So I think It was correct before

hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
```diff
@@ -90,12 +95,18 @@ private static void assertInstants(Record r) {
...
 				.isAfterOrEqualTo( e.history.created );
 		assertThat( e.history.created ).isEqualTo( h.created );
 
+		assertThat( e.currentTimestampEmbedded.currentCreatedAt.truncatedTo( ChronoUnit.HOURS ) ).isEqualTo( h.created.truncatedTo( ChronoUnit.HOURS ) );
+		assertThat( e.currentTimestampEmbedded.currentLastUpdatedAt )
+				.as( "Updated instant is after created. Updated[" + e.currentTimestampEmbedded.currentLastUpdatedAt + "], Created[" + e.currentTimestampEmbedded.currentCreatedAt + "]" )
```
DavideD: See my other comment

hibernate-reactive-core/src/test/java/org/hibernate/reactive/TimestampTest.java
```diff
@@ -109,6 +120,8 @@ static class Record {
 		Instant created;
 		@UpdateTimestamp
 		Instant updated;
+		@CurrentTimestamp
+		Instant current;
```
DavideD: MS SQL Server is not happy with a property named `current` (it's a special keyword in it).
You can solve it escaping the name with `@Column`:
```java
@CurrentTimestamp
@Column(name="`current`")
Instant current
```
(I think "\`" is one of the escape chars, but I didn't check. This should also work `name ="[current]"`)

Or, you can just renamed the field

DavideD:
Could you squash everything in one commit? And the issue number is #3107

roguebee-droid:
Hrm unsure why it closed like that. I thought I was working in a branch but obviously wasn't so I've rebased main so it's a single commit. Sorry about the mess, the main branch on mine should now be ok I hope.

DavideD:
THanks for rebasing.
I was ready to merge this, but after a rebase on main, the test with the embedded fails on my machine with the following error:
```
rg.opentest4j.AssertionFailedError: 
expected: 2026-03-02T18:22:19 (java.time.LocalDateTime)
 but was: 2026-03-02T17:22:19 (java.time.LocalDateTime)
when comparing values using 'ChronoLocalDateTime.timeLineOrder()'
java.lang.AssertionError: org.opentest4j.AssertionFailedError: 
expected: 2026-03-02T18:22:19 (java.time.LocalDateTime)
 but was: 2026-03-02T17:22:19 (java.time.LocalDateTime)
when comparing values using 'ChronoLocalDateTime.timeLineOrder()'
	at io.vertx.junit5.VertxExtension.joinActiveTestContexts(VertxExtension.java:205)
	at io.vertx.junit5.VertxExtension.interceptTestMethod(VertxExtension.java:143)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
Caused by: org.opentest4j.AssertionFailedError: 
expected: 2026-03-02T18:22:19 (java.time.LocalDateTime)
 but was: 2026-03-02T17:22:19 (java.time.LocalDateTime)
when comparing values using 'ChronoLocalDateTime.timeLineOrder()'
	at org.hibernate.reactive.TimestampTest.assertInstants(TimestampTest.java:106)
	at org.hibernate.reactive.TimestampTest.lambda$testEmbedded$10(TimestampTest.java:78)


Expected :2026-03-02T18:22:19 (java.time.LocalDateTime)
Actual   :2026-03-02T17:22:19 (java.time.LocalDateTime)
<Click to see difference>

org.opentest4j.AssertionFailedError: 
expected: 2026-03-02T18:22:19 (java.time.LocalDateTime)
 but was: 2026-03-02T17:22:19 (java.time.LocalDateTime)
when comparing values using 'ChronoLocalDateTime.timeLineOrder()'
	at org.hibernate.reactive.TimestampTest.assertInstants(TimestampTest.java:106)
	at org.hibernate.reactive.TimestampTest.lambda$testEmbedded$10(TimestampTest.java:78)
```

Does it work for you?

Issue #3108, opened 2026-02-06
Test support for '@PartionKey


DavideD:
See https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/annotations/PartitionKey.java

DavideD:
Thanks

PR #3262, opened 2026-02-27
[4.3] test support for partition key annotation

roguebee-droid:
I've added a fairly straightforward test for partition key #3108  based on what is in [hibernate-orm](https://github.com/hibernate/hibernate-orm/blob/007a34920d10c6787823b7f6d33086b46461c57b/hibernate-core/src/test/java/org/hibernate/orm/test/mapping/identifier/SimplePartitionKeyTest.java)

Hibernate-orm does have some more tests for this annotation for specific databases as well as [association](https://github.com/hibernate/hibernate-orm/blob/018b8eeda3627e114ec25bd48407ccb9c47564ce/hibernate-core/src/test/java/org/hibernate/orm/test/sql/partition/PartitionKeyAndAssociationTest.java#L59) which I haven't done here but could add if required.


DavideD (changes requested):
See my previous comment: https://github.com/hibernate/hibernate-reactive/pull/3262#issuecomment-4030617025

DavideD:
The code is out of date and the test is not really asserting the working of the `@PartitionKey`. I could remove the annotation and it would still pass.

PR #3274, opened 2026-03-02
[3.2] In the docs, inlcude the SCRAM GAV as variable

DavideD:
The value was hardcoded in the documentation. Now it's obtaioned by the actual dependency in the catalogue


marko-bekhta (approved):
🎉 🙂

PR #3279, opened 2026-03-02
[4.3] Fix Table of Contents (TOC) rendering

DavideD:
Because we only have one top section, the Table of Contents (TOC) doesn't look good. These changes expand the subsections showing more entries in the TOC.

It should be enough to fix the issue for the time being.

Original version:
Screenshot From 2026-03-02 11-34-38

New one:
Screenshot From 2026-03-02 11-28-42


documentation/src/main/asciidoc/reference/index.adoc
```diff
@@ -1,18 +1,34 @@
...
+:toclevels: 4
 :sectanchors:
 :sourcedir: src/main/asciidoc/reference
-:leveloffset: +1
 :version-selector-enabled: true
 :html-meta-canonical-link: https://docs.hibernate.org/stable/reactive/reference/html_single
 
+:leveloffset: +1
+
 include::{sourcedir}/preface.adoc[]
```
marko-bekhta:
```suggestion
include::{sourcedir}/_preface.adoc[]
```

maybe let's update this one too ? 
(I think it's adoc "convention" to use the `_` for non-root files)

Issue #3367, opened 2026-03-05
Refactor identifier generator classes


DavideD:
We can remove or simplify a lot of code related to the sequence or table based generator.
They accumulated a lot of debt because they used to work fine after every ORM update

PR #3398, opened 2026-03-10
[4.3] Remove ReactiveSequenceIdentifierGenerator.configure

dreab8:
For #3367


hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/ReactiveSequenceIdentifierGenerator.java
```diff
@@ -76,82 +59,10 @@ protected CompletionStage<Long> nextHiValue(ReactiveConnectionSupplier session)
...
-				StandardNamingStrategy.class::getName
-		);
-
-		return strategySelector
-				.resolveStrategy( ImplicitDatabaseObjectNamingStrategy.class, namingStrategySetting )
-				.determineSequenceName( catalog, schema, params, serviceRegistry );
+		// No explicit call is required here. The identifier generator is created
+		// based on the configuration provided by SequenceStyleGenerator,
+		// the DatabaseStructure passed in the constructor, which
+		// already defines the appropriate sequence name and increment size.
```
DavideD: It's better to remove the method if we don't need it. This way we avoid future conflicts or error if the signature changes.

I think looking at the class it's pretty clear where we get the configuration. Maybe add this as a comment for the commit.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/id/impl/ReactiveSequenceIdentifierGenerator.java
```diff
@@ -52,6 +33,7 @@ public class ReactiveSequenceIdentifierGenerator extends BlockingIdentifierGener
 
 	private String sql;
 	private int increment;
+	private boolean suppoertBulkInsertion;
```
DavideD: Typo, it should be `supportBulkInsertion`

PR #3527, opened 2026-04-02
Test new Hibernate ORM 6.3 session-level batch/subselect fetching API

### LINKED ISSUE #1759: Test new Hibernate ORM 6.3 API for batching
We already test batching in some places: 
* https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/BatchFetchTest.java
*  https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SubselectFetchTest.java

This is the PR where we add the new methods: https://github.com/hibernate/hibernate-reactive/pull/1747

yunhwane:
## Summary

Add `SessionFetchBatchSizeTest` to cover the session-level batch and subselect fetching API introduced in #1747.

**Tests added:**
- `testGetAndSetFetchBatchSizeWithStageSession` — verify get/set for `Stage.Session`
- `testGetAndSetSubselectFetchingWithStageSession` — verify toggle for `Stage.Session`
- `testGetAndSetFetchBatchSizeWithMutinySession` — verify get/set for `Mutiny.Session`
- `testGetAndSetSubselectFetchingWithMutinySession` — verify toggle for `Mutiny.Session`
- `testBatchFetchingWithSessionLevelBatchSize` — verify batch fetching works with session-level batch size
- `testSubselectFetchingWithSessionLevelSetting` — verify subselect fetching initializes all collections

Closes #1759

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SessionFetchBatchSizeTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/SessionFetchBatchSizeTest.java
@@ -1,0 +1,272 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+import org.hibernate.Hibernate;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.reactive.testing.SqlStatementTracker;
+import org.hibernate.reactive.util.impl.CompletionStages;
+
+import org.junit.jupiter.api.Test;
+
+import io.smallrye.mutiny.Uni;
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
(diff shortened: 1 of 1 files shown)


DavideD (changes requested):
Thanks, but I don't think the test are quite right.

They check that the value for the properties is set, but they don't test if the actual executed SQL queries use batching. I'm pretty sure that if I remove `s.setFetchBatchSize` from testBatchFetchingWithSessionLevelBatchSize, the test won't fail.

yunhwane:
Thanks for the review @DavideD! You were right — the previous tests only
verified that the property was set, not that the SQL actually changed.

I've reworked them to use SqlStatementTracker:

- testBatchFetchingWithSessionLevelBatchSize and
  testSubselectFetchingWithSessionLevelSetting now assert exactly one
  SELECT against SFBChild after the fetch, plus that both p1.children
  and p2.children are initialized — proving the second collection was
  loaded in the same query.
- Removing s.setFetchBatchSize(10) now causes the test to fail (two
  separate SELECTs are issued and p2.children stays uninitialized).
- The SQL pattern accepts both the generic IN (?, ?, ...) form and
  PostgreSQL's = ANY ($1) array binding.
- Verified locally against PostgreSQL via Testcontainers — all 6 tests pass.

Also addressed the CodeQL note about the unused `defaultSize` variable.

PR #3602, opened 2026-04-18
[#1783] Fix NullPointerException when overriding logger

### LINKED ISSUE #1783: Overriding logger casuses NullPonterException
I'm creating an issue so that I don't forget to review it.
See https://discourse.hibernate.org/t/custom-sqlstatementlogger-setup/7680/6

utafrali:
Fixes #1783

Custom logger setups that don't provide SqlStatementLogger were causing NPEs when the pool tried to use a null reference. Added a null check to fall back to a default logger instance if the service isn't available. Includes a test that reproduces the scenario.

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPool.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/pool/impl/DefaultSqlClientPool.java
@@ -123,7 +123,8 @@ public class DefaultSqlClientPool extends SqlClientPool
 	@Override
 	public void injectServices(ServiceRegistryImplementor serviceRegistry) {
 		this.serviceRegistry = serviceRegistry;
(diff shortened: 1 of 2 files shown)


DavideD (changes requested):
Thasnk, but I don't if this is the right approach:
1. You should create the test in a new class that extends `BaseReactiveTest`
    You can then override `BaseReactiveTest#addServices` and add your custom logger class there. 
2. You need a custom logger class that extends `SqlStatementLogger` and capture log calls so that you can then assert that the logger has been called
3. I don't think you have to change `DefaultSqlClientPool#injectServices`. I don't expect `serviceRegistry.getService( SqlSatementLogger.class )` to ever return null unless there's a bug. It will use the default service or the custom one, why would you expect that to be `null`?
4. The test should assert that the custom log has been called and not the default one

I'm going to close this PR, feel free to send a new one or reopen when you made the changes.

Cheers

Issue #1766, opened 2023-09-15
Test upsert with inheritance


DavideD:
We only test the upsert for [ReactiveSingleTableEntityPersister](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveSingleTableEntityPersister.java), but it's also implemented for [ReactiveUnionSubclassEntityPersister](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveUnionSubclassEntityPersister.java) and [ReactiveJoinedSubclassEntityPersister](https://github.com/hibernate/hibernate-reactive/blob/main/hibernate-reactive-core/src/main/java/org/hibernate/reactive/persister/entity/impl/ReactiveJoinedSubclassEntityPersister.java)

We should also test the deletion of rows with upsert works (currently we only test the update)
Related to issue #1702

firatulgay:
Hello,

I want to work on this issue. I found the UpsertTest.java class which contains the upsert tests already. I will check this class for a reference. But I didnt get exactly what do you mean  here "deletion of rows with upsert works". Can you help me on this please.

DavideD:
Hello,
forget about the delete. When I created the issue I thought there was a way to also do that.

Thanks a lot for the offer.

If you create the test and  it doesn't work, make sure to sync with @blafond  before applying changes to the main code.
He's working on something related to this.

As far as contribution guides, we use the same as the ORM one: https://github.com/hibernate/hibernate-orm/blob/main/CONTRIBUTING.md

Thanks a lot

coder-github578:
Hey David,

Is there any sample example for upsert

DavideD:
You can check this test: https://github.com/hibernate/hibernate-reactive/blob/bd2566000144391b73388c1a4adf1aeceb62b019/hibernate-reactive-core/src/test/java/org/hibernate/reactive/UpsertTest.java#L76

PR #3603, opened 2026-04-18
[#1766] Add tests for upsert with inheritance

utafrali:
Added tests for upsert with inheritance strategies. Tests both JOINED and UNION inheritance, and both Mutiny and Stage APIs. Each test does an initial upsert, verifies the data, then upserts again with the same ID to check updates work.


DavideD:
Looks good thanks, I would probably add a test that changes breed but not name, and check that both fields have the expected value.

In any case, it seems that this feature doesn't work at the moment.
Do you want to try to fix it?

Issue #3417, opened 2026-03-16
Upgrade Hibernate ORM to 7.4


DavideD:

DavideD:
Merged the changes in `main` and created the branch `3.4` (for Vert.x 4): https://github.com/hibernate/hibernate-reactive/tree/3.4

PR #3678, opened 2026-05-11
[4.4] Upgrade Hibernate ORM to 7.4

dreab8:
For #3417


DavideD:
Thanks @dreab8, this works, but I need to solve some conflicts and it's easier if I send a new PR.

PR #3679, opened 2026-05-11
[4.4] Use an ssh git repository URL for the "website release" step

### LINKED ISSUE #3710: Use an ssh git repository URL for the "website release" step

marko-bekhta:
see:

- https://github.com/hibernate/hibernate-search/pull/5117#discussion_r3186742781

As we are getting these commits as Hibernate-ci user: https://github.com/hibernate/hibernate.org/commit/2c945d993907c7c922c25f8959fd3953a437d2da, I suspect that the config for a Jenkins job works and we don't need the extra config to set the user here 🤞🏻 

Let's give it a try on the next reactive release, and if it works, adjust other projects?

### DIFF
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -235,20 +235,16 @@ pipeline {
 							configFile(fileId: 'release.config.ssh', targetLocation: "${env.HOME}/.ssh/config"),
 							configFile(fileId: 'release.config.ssh.knownhosts', targetLocation: "${env.HOME}/.ssh/known_hosts")
 					]) {
-						withCredentials([
-								gitUsernamePassword(credentialsId: 'username-and-token.Hibernate-CI.github.com', gitToolName: 'Default')
-						]) {
-							sshagent( ['ed25519.Hibernate-CI.github.com'] ) {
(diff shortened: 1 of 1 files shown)


yrodiere:
> As we are getting these commits as Hibernate-ci user: [hibernate/hibernate.org@2c945d9](https://github.com/hibernate/hibernate.org/commit/2c945d993907c7c922c25f8959fd3953a437d2da), I suspect that the config for a Jenkins job works and we don't need the extra config to set the user here 🤞🏻

No idea how that works... But yeah +1 to try; worst case there's just one commit on hibernate.org with the wrong user/email.

yrodiere (approved):

DavideD:
THanks @marko-bekhta, could you add this to the commit description (or something similar):

> As we are getting these commits as Hibernate-ci user: https://github.com/hibernate/hibernate.org/commit/2c945d993907c7c922c25f8959fd3953a437d2da, I suspect that the config for a Jenkins job works and we don't need the extra config to set the user here 🤞🏻

Thanks

marko-bekhta:
Added more info to the commit message 🙈 🙂 let me know if that works 🙇🏻 🙂

PR #3813, opened 2026-05-28
[4.3] Use Vert.x 5 Context local storage API

### LINKED ISSUE #3793: Switch to Vert.x 5 API for ContextLocal storage
See https://vertx.io/docs/guides/vertx-5-migration-guide/#_context_local_storage_changes

Currently, we are still using the Vert.x 4 API:  https://github.com/hibernate/hibernate-reactive/blob/00c81250538d14ba3c7e2aedce0605fb54148d13/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/VertxContext.java#L38

But we should move from using `ContextInternal` to use `ContextLocal`.

For reference, we did it in the past, but there were some issues: https://github.com/hibernate/hibernate-reactive/issues/2234
We can probably reapply those changes and see if it works now (it might need some updates).

This issue is a follow up from https://github.com/quarkusio/quarkus/pull/54285

DavideD:
When we upgraded to Vert.x 5 we couldn't use the API because Quarkus wasn't ready.

For more details, check the description on issue #3793

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
@@ -1,0 +1,47 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.reactive.context.impl;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.hibernate.reactive.context.Context;
+
+import io.vertx.core.internal.VertxBootstrap;
+import io.vertx.core.spi.VertxServiceProvider;
+import io.vertx.core.spi.context.storage.ContextLocal;
+
+import static io.vertx.core.spi.context.storage.ContextLocal.registerLocal;
+
+/**
+ * SPI Implementation for {@link ContextLocal} storage.
+ */
+public class ContextualDataStorage implements VertxServiceProvider {
+
+	private static final ContextLocal<ConcurrentHashMap> CONTEXTUAL_DATA_KEY =
+			registerLocal( ConcurrentHashMap.class, ConcurrentHashMap::new );
+
+	@Override
+	public void init(VertxBootstrap builder) {
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> void put(io.vertx.core.Context vertxContext, Context.Key<T> key, T value) {
+		( (Map<Context.Key<T>, T>) vertxContext
+				.getLocal( CONTEXTUAL_DATA_KEY, ConcurrentHashMap::new ) )
+				.put( key, value );
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> T get(io.vertx.core.Context vertxContext, Context.Key<T> key) {
+		Map<Context.Key<T>, T> map = vertxContext.getLocal( CONTEXTUAL_DATA_KEY );
+		return map == null ? null : map.get( key );
+	}
+
+	public static boolean remove(io.vertx.core.Context vertxContext, Context.Key<?> key) {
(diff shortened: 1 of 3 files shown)


DavideD:
@jponge, @lucamolteni, could you have a look at this PR and let me know if it makes sense? Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
```diff
@@ -0,0 +1,49 @@
...
+import static io.vertx.core.spi.context.storage.AccessMode.CONCURRENT;
+import static io.vertx.core.spi.context.storage.ContextLocal.registerLocal;
+
+/**
+ * SPI Implementation for {@link ContextLocal} storage.
+ */
+public class ContextualDataStorage implements VertxServiceProvider {
+
+	@SuppressWarnings("rawtypes")
+	private static final ContextLocal<ConcurrentHashMap> CONTEXTUAL_DATA_KEY =
```
jponge, line 23: You might still want to set parameter types.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
```diff
@@ -0,0 +1,49 @@
...
+	public static <T> void put(io.vertx.core.Context vertxContext, Context.Key<T> key, T value) {
+		( (Map<Context.Key<T>, T>) vertxContext
+				.getLocal( CONTEXTUAL_DATA_KEY, CONCURRENT, ConcurrentHashMap::new ) )
+				.put( key, value );
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> T get(io.vertx.core.Context vertxContext, Context.Key<T> key) {
+		Map<Context.Key<T>, T> map = vertxContext.getLocal( CONTEXTUAL_DATA_KEY, CONCURRENT );
+		return map == null ? null : map.get( key );
```
jponge, line 40: It should never be `null`, but it doesn't hurt to have a guard 😉

jponge (commented):
It looks good to me.

Perhaps we should have another check from @tsegismont ?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
DavideD: I tried different combination of parameters but this is the only one that seems to work.
I'm open to suggestions, though

hibernate-reactive-core/src/main/java/org/hibernate/reactive/context/impl/ContextualDataStorage.java
DavideD: I'm going to add a `requireNonNull`, just to highlight that we don't expect the value to be null

DavideD: Actually, it can be null. In HIbernate REactive we check if a new session already exists before creating a new one. ANd if it's the first time, there won't be any map yet

jponge (approved):

DavideD:
I removed the `CONCURRENCY` parameter when using getLocal because it's the default already

tsegismont (commented):
It looks good to me.

If I remember correctly, we had to rollback the previous attempt because Quarkus expect Hibernate Reactive to use the shared CHM for storage.

If that's correct, I guess the issue has been addressed in Quarkus. In this case, can you please share the link to the corresponding PR?

DavideD:
Thanks,

> If that's correct, I guess the issue has been addressed in Quarkus. In this case, can you please share the link to the corresponding PR?

I don't know anything about this. The old discussion about Quarkus not supporting it is here: https://github.com/quarkusio/quarkus/pull/47441#discussion_r2065582053

PR #3847, opened 2026-06-05
Add GHA workflow to run a linter on future GHA workflow changes

yrodiere:
Same as https://github.com/hibernate/hibernate-orm/pull/12653


yrodiere:
https://github.com/hibernate/hibernate-orm/pull/12653 was approved and was identical -- merging.

PR #3848, opened 2026-06-05
Cleanup linter issues in the workflow files

marko-bekhta:


DavideD:
THanks! Do you know if I have to apply these changes to the other branches too? Or is it enough on main?

marko-bekhta:
hey 👋🏻 🙂  I think I broke something in the JDK tests 🙈 
let me fix that first 🙂 

> Or is it enough on main?

I think the linter would run on any PR that changes the workflow, so probably not super urgent, but might be good to backport it if you think we'd  need to update those workflows at some point (for other than linting reasons 🙂 )

marko-bekhta:
all ✅ , should we merge? 🙈 🙂 🤞🏻

yrodiere:
Merging because other urgent things depend on this.

PR #3925, opened 2026-06-24
Switch to JDK 25 for the release / Use the same theme for javadoc jars as the one for docs published to the site

### LINKED ISSUE #3923: Use the same theme for javadoc jars as the one for docs published to the site
And while at it -- skip copying the fonts

### LINKED ISSUE #3924: Switch to JDK 25 for the release
We can switch to the new JDK for releases (at least on new versions) and also use the new javadoc theme that other projects already applied

marko-bekhta:
* Fixes https://github.com/hibernate/hibernate-reactive/issues/3923
* Fixes https://github.com/hibernate/hibernate-reactive/issues/3924


marko-bekhta:
I think ... these `--no-fonts` and `--add-script` would only work ok on JDK 25+, so we'd only want to apply this where we are ok with building on JDK 25 (for the release) 🙂

DavideD:
Sorry, @marko-bekhta, could you rebase this and fix the conflicts so that I can merge it?

marko-bekhta:
hey no worries 🙂 btw  25 theme comes with a dark mode switch:
image

DavideD:
@marko-bekhta It seems something is not quite right with this PR

marko-bekhta:
this should be cleaner, some options like `-no-fonts` (JDK 23 ?) and `-add-script` (21 ?) are only in more recent javadoc tools, so when we build with something lower like 17, things would fail 🙈 
so we check if it's jdk 25 and only add the properties there. Should be ok since the theme that needs the script is a JDK25-based one anyway, but let's see if CI agrees 🤞🏻

PR #4009, opened 2026-07-20
Assert custom @SQL* statements in CustomStoredProcedureSqlTest

### LINKED ISSUE #1673: Check that we actually run the custom queries in CustomStoredProcedureSqlTest

arimu1:
## Summary

Fixes https://github.com/hibernate/hibernate-reactive/issues/1673

`CustomStoredProcedureSqlTest` already exercises `@SQLInsert` / `@SQLUpdate` / `@SQLDelete` via stored procedures, but never asserted that Hibernate actually emitted those custom `SELECT PROC_*` statements.

This wires in `SqlStatementTracker` the same way neighboring tests do (`OrderQueriesTestBase`, etc.) and asserts each CRUD path logged the expected `PROC_INSERT_RECORD` / `PROC_UPDATE_RECORD` / `PROC_DELETE_RECORD` call.

Prior attempts: #1709 / #1710 (closed).

## Test plan

- [x] `./gradlew :hibernate-reactive-core:test --tests org.hibernate.reactive.CustomStoredProcedureSqlTest` — 5/5 pass (PostgreSQL / Testcontainers)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CustomStoredProcedureSqlTest.java
+++ b/hibernate-reactive-core/src/test/java/org/hibernate/reactive/CustomStoredProcedureSqlTest.java
@@ -12,7 +12,10 @@ import org.hibernate.HibernateException;
 import org.hibernate.annotations.SQLDelete;
 import org.hibernate.annotations.SQLInsert;
 import org.hibernate.annotations.SQLUpdate;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.cfg.Configuration;
 import org.hibernate.reactive.testing.ReactiveAssertions;
+import org.hibernate.reactive.testing.SqlStatementTracker;
 import org.hibernate.reactive.annotations.EnabledFor;
 
 import org.junit.jupiter.api.Assertions;
@@ -40,6 +43,12 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 public class CustomStoredProcedureSqlTest extends BaseReactiveTest {
 
 	private SimpleRecord theRecord;
+
+	// Tracks the SQL statements emitted by Hibernate so that the tests can assert
+	// the custom @SQLInsert / @SQLUpdate / @SQLDelete queries are actually used.
+	// Static because BaseReactiveTest reuses one SessionFactory across test methods,
+	// so constructConfiguration() (which creates the tracker) runs only once per class.
+	private static SqlStatementTracker sqlTracker;
 
 	private static final String INITIAL_TEXT = "blue suede shoes";
 	private static final String UPDATED_TEXT = "ruby slippers";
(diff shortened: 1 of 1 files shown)


DavideD:
I will review it, but in the meanwhile, could you update the commit title with the issue number as prefix  (some thing like `[#1673] Something something`). And you should also add `@Co-authored-by Cursor` (There's an example here with Claude: https://github.com/hibernate/hibernate-reactive/commit/55f657f43b74bf7e1ae65e58ec7c3a572882893e). Make sure to add the version or the id of the AI agent you are using

hibernate-reactive-core/src/test/java/org/hibernate/reactive/CustomStoredProcedureSqlTest.java
```diff
@@ -130,6 +164,9 @@ public void testUpdateStoredProcedure(VertxTestContext context) {
 						.find( SimpleRecord.class, theRecord.id )
 						.thenAccept( foundRecord -> foundRecord.text = UPDATED_TEXT )
 						.thenCompose( v -> session.flush() ) )
+				.thenAccept( v -> assertThat( sqlTracker.getLoggedQueries() )
+						.as( "Custom @SQLUpdate statement was not emitted" )
+						.anyMatch( s -> s.contains( "SELECT PROC_UPDATE_RECORD" ) ) )
```
DavideD: Instead of using `.anyMatch` could you check that the exact number of queries we expect is executed? This is true for the other tests as well.

If you use anyMatch, there's no way to notice if we are updating the same record twice (for example).
I just usually check that the log contains the expected queries in the list using `.containsExactly`.

DavideD (changes requested):
Thanks, looks good overall.

I just prefer when the assertions are more precise because they are better at recognizing changes in behaviours or errors.

arimu1:
Thanks @DavideD — updated:

1. Assertions now use `containsExactly` with the exact `@SQL*` SQL strings (and `sqlTracker.clear()` around the update/delete actions so the count is exact).
2. Commit title prefixed with `[#1673]`.
3. Added `Co-authored-by: Cursor <cursoragent@cursor.com>` (Composer agent).

DavideD (approved):

DavideD:
Thanks a lot, this will be merged automatically if the build passes successfully.

arimu1:
@DavideD Thanks — the insert/update/delete tests already use `containsExactly` for exact query lists (no `.anyMatch`). Commit title is `[#1673] …` with `Co-authored-by: Cursor`. Re-triggered CI; the earlier red was a Db2 job timeout (~6h), not the PostgreSQL-only stored-procedure assertions.

DavideD (approved):

PR #4073, opened 2026-08-12
[4.5] Add ReactiveSelectionQuery.getKeyedResultList

### LINKED ISSUE #3811: Add ReactiveSelectionQuery.getKeyedResultList
See https://github.com/hibernate/hibernate-orm/blob/5fd28791abce49100091934abc31d3c8d03f876c/hibernate-core/src/main/java/org/hibernate/query/SelectionQuery.java#L345

We should probably check if there are other methods that are missing

dreab8:
Fix for #3811 

It needs the following changes in ORM https://github.com/hibernate/hibernate-orm/pull/13180

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
@@ -6,6 +6,7 @@ package org.hibernate.reactive.mutiny;
 
 import java.lang.invoke.MethodHandles;
 import java.util.List;
+import java.util.concurrent.CompletionStage;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
@@ -22,6 +23,8 @@ import org.hibernate.engine.spi.PersistentAttributeInterceptor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.jpa.internal.util.FlushModeTypeHelper;
 import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.query.KeyedPage;
+import org.hibernate.query.KeyedResultList;
 import org.hibernate.query.Page;
 import org.hibernate.query.criteria.HibernateCriteriaBuilder;
 import org.hibernate.query.criteria.JpaCriteriaInsert;
@@ -401,6 +404,9 @@ public interface Mutiny {
 
 		@Override
 		SelectionQuery<R> setComment(String comment);
+
+		@Incubating
+		CompletionStage<KeyedResultList<R>> getReactiveKeyedResultList(KeyedPage<R> page);
 	}
 
 	interface MutationQuery extends AbstractQuery {
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyQueryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyQueryImpl.java
@@ -9,6 +9,8 @@ import org.hibernate.FlushMode;
 import org.hibernate.LockMode;
 import org.hibernate.graph.GraphSemantic;
 import org.hibernate.graph.spi.RootGraphImplementor;
+import org.hibernate.query.KeyedPage;
+import org.hibernate.query.KeyedResultList;
 import org.hibernate.query.Page;
 import org.hibernate.reactive.mutiny.Mutiny;
 import org.hibernate.reactive.mutiny.Mutiny.Query;
@@ -245,6 +247,11 @@ public class MutinyQueryImpl<R> implements Query<R> {
 	public Query<R> setComment(String comment) {
 		delegate.setComment( comment );
 		return this;
+	}
+
+	@Override
+	public CompletionStage<KeyedResultList<R>> getReactiveKeyedResultList(KeyedPage<R> page) {
+		return delegate.getReactiveKeyedResultList( page );
 	}
 
 	@Override
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sql/internal/ReactiveNativeQueryImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sql/internal/ReactiveNativeQueryImpl.java
@@ -26,6 +26,8 @@ import org.hibernate.graph.GraphSemantic;
 import org.hibernate.graph.RootGraph;
 import org.hibernate.graph.spi.RootGraphImplementor;
 import org.hibernate.metamodel.model.domain.BasicDomainType;
+import org.hibernate.query.KeyedPage;
+import org.hibernate.query.KeyedResultList;
 import org.hibernate.query.QueryParameter;
 import org.hibernate.query.ResultListTransformer;
 import org.hibernate.query.TupleTransformer;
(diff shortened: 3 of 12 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sql/internal/ReactiveNativeQueryImpl.java
```diff
@@ -117,6 +119,11 @@ public CompletionStage<Long> getReactiveResultCount() {
 		throw LOG.notYetImplemented();
 	}
 
+	@Override
+	public CompletionStage<KeyedResultList<R>> getReactiveKeyedResultList(KeyedPage<R> page) {
+		throw new UnsupportedOperationException("native queries do not support key-based pagination");
```
DavideD, line 125: Could you add a comment to mention that this this is the same approach in ORM?
Out of curiosity, is there any specific reason why this si not supported for native queries?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
```diff
@@ -163,6 +166,40 @@ public void testMaxResultsWithStage(VertxTestContext context) {
 		);
 	}
 
+	@Test
+	public void testReactiveKeyedPagination(VertxTestContext context) {
+		test( context, openSession()
+				.thenCompose( session -> {
+					final ReactiveSession reactiveSession = ( (StageSessionImpl) session ).unwrap( ReactiveSession.class );
```
DavideD: This doesn't seem right. Shouldn't the user be able to access this from the public API?
See https://github.com/hibernate/hibernate-reactive/blob/0aefeb7c8082bf1dcb639866682cc2f51de9b51f/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java#L133

Same for `Stage.SelectionQuery`

DavideD (changes requested):
Looks good overall, but I think we need to have the new methods in the `Mutiny.SelectionQuery` and `Stage.SelectionQuery` interfaces.

Not critical, but I would also add a test to see what happens when there are no results from the query (just in case). And that we throw the exception for Native queries.

Thanks

hibernate-reactive-core/src/main/java/org/hibernate/reactive/query/sql/internal/ReactiveNativeQueryImpl.java
dreab8: I[ll add the comment but I don't know why it is not supported for native queries

hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
```diff
@@ -163,6 +164,39 @@ public void testMaxResultsWithStage(VertxTestContext context) {
 		);
 	}
 
+	@Test
+	public void testReactiveKeyedPagination(VertxTestContext context) {
+		test( context, openSession()
+				.thenCompose( session -> {
+					var query = session.createSelectionQuery(
```
DavideD, line 172: That's better. But you also need to write a test for Mutiny because this is part of the public API.
Could test what happens when the query doesn't return any result?

DavideD (changes requested):
LGTM, but you need add some tests

DavideD:
In the meanwhile, I merged these changes in `wip/main`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/Mutiny.java
```diff
@@ -401,6 +404,9 @@ default SelectionQuery<R> setLockMode(String alias, LockModeType lockModeType) {
 
 		@Override
 		SelectionQuery<R> setComment(String comment);
+
+		@Incubating
+		CompletionStage<KeyedResultList<R>> getReactiveKeyedResultList(KeyedPage<R> page);
```
DavideD: This should be `Uni<KeyedResultList<R>>`

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinyQueryImpl.java
```diff
@@ -247,6 +249,11 @@ public Query<R> setComment(String comment) {
 		return this;
 	}
 
+	@Override
+	public CompletionStage<KeyedResultList<R>> getReactiveKeyedResultList(KeyedPage<R> page) {
```
DavideD: it Should be `Uni`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
```diff
@@ -183,6 +254,77 @@ public void testMaxResultsWithMutiny(VertxTestContext context) {
...
+				.chain( session -> {
+					var query = session.createSelectionQuery(
+							"from FamousPerson p where p.status = :status order by p.name, p.id",
+							FamousPerson.class
+					);
+					final var firstPage = first( 2 ).keyedBy( List.of(
+							Order.asc( FamousPerson.class, "name" ),
+							Order.asc( FamousPerson.class, "id" )
+					) );
+					return Uni.createFrom().completionStage( query.setParameter( "status", Status.LIVING )
```
DavideD: YOu have to do this transformation because the method return a `CompletionStage` instead of a `Uni`

hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
```diff
@@ -183,6 +254,77 @@ public void testMaxResultsWithMutiny(VertxTestContext context) {
...
+	@Test
+	public void testReactiveKeyedPaginationWithNativeQueryThrowsMutiny(VertxTestContext context) {
+		test( context, assertThrown( UnsupportedOperationException.class, openMutinySession()
+				.chain( session -> Uni.createFrom().completionStage( session.createNativeQuery(
+						"select * from FamousPerson order by id",
+						FamousPerson.class
+				)
+						.getReactiveKeyedResultList( first( 2 ).keyedBy( Order.asc( FamousPerson.class, "id" ) ) ) ) ) )
+				.invoke( error -> assertThat( error )
+						.hasMessage( "native queries do not support key-based pagination" ) )
```
DavideD: Maybe it's better to have an error code here? so that you can check that instead of the error message

hibernate-reactive-core/src/test/java/org/hibernate/reactive/FilterWithPaginationTest.java
```diff
@@ -183,6 +254,77 @@ public void testMaxResultsWithMutiny(VertxTestContext context) {
...
+					);
+					final var firstPage = first( 2 ).keyedBy( List.of(
+							Order.asc( FamousPerson.class, "name" ),
+							Order.asc( FamousPerson.class, "id" )
+					) );
+					return Uni.createFrom().completionStage( query.setParameter( "status", Status.LIVING )
+							.getReactiveKeyedResultList( firstPage ) )
+							.chain( firstResults -> {
+								assertThat( firstResults.getResultList() ).containsExactly( margaret, rebeccaActress );
+								assertThat( firstResults.getKeyList() ).hasSize( 2 );
```
DavideD, line 280: Shouldn't you check that `KeyList` matches the content of the ResultList? Shoudln't it look something like: 
```java
assertThat( firstResults.getKeyList() ).containsExactly(
										List.of( margaret.name, margaret.id ),
										List.of( rebeccaActress.name, rebeccaActress.id )
								);
```

This is true for the other tests and assertions that do something similar

DavideD (changes requested):

DavideD (approved):

DavideD:
Merged, thanks!

Issue #2494, opened 2025-09-04
Guarantee serialization of calls to the session


yrodiere:
A user can write this:

```
return sessionFactory.withSession(s -> {
   return Uni.combine().combine().all().unis(s.doSomething(), s.doSomethingElse()).asTuple();
});
```

It is incorrect, because it could result in interleaved uses of the session, messing up its state: `doSomething()` could set up some temporary state, do some IO so release control of the event loop, then `doSomethingElse()` would start executing, mess up the temporary state, and we're in a bad place.

[This section of the docs](https://hibernate.org/reactive/documentation/4.1/reference/html_single/#_sessions_and_vert_x_contexts) explains a related problem -- not the same one, but the explanations are relevant.

Unfortunately -- on contrary to the problem mentioned in the docs -- this problem is not detected and will be very hard to debug.

More importantly... it's probably just as easy to fix as it is to detect. We could simply make sure that all operations in the session start running *after the previous one finished*.

E.g. this code in `ReactiveSessionImpl`:

```java
	@Override
	public <T> CompletionStage<T> reactiveFind(
			Class<T> entityClass,
			Object id,
			LockOptions lockOptions,
			EntityGraph<T> fetchGraph) {
		checkOpen();
		return supplyStage( () -> {
			...
		} );
```

Could become:

```java
	@Override
	public <T> CompletionStage<T> reactiveFind(
			Class<T> entityClass,
			Object id,
			LockOptions lockOptions,
			EntityGraph<T> fetchGraph) {
		checkOpen();
		this.previousOperation = this.previousOperation.thenCompose(ignored -> supplyStage( () -> {
			...
		} ) );
		return previousOperation;
```

(Or better yet, we'd integrate this with `supplyStage`, which seems to be doing something similar with a `voidFuture`)

`previousOperation` would just be a field in the session, initialized to a completed future.

We could do this for `StatelessSession` as well, though it will be harder to explain why state can be a problem with a "stateless" session...

yrodiere:
Note the suggested solution could... perhaps... possibly... cause big trouble if some session operations call other session operations. I'm not sure, but it's something to be careful about.

Sanne:
N.B. if you check the sources of `ReactiveSessionImpl` you'll also notice assertions for:
 - InternalStateAssertions.assertUseOnEventLoop()
 - never having switched the actual thread

[...]

PR #4080, opened 2026-08-20
Serialization of the session operations

DavideD:
A first draft for #2494 

This solution is built on top of `main` but I'm wondering if we shouldn't include after or with the upgrade to ORM 8.0

The approach is the one suggested in the issue: operations get concatenated when they get called on the session.

I might need to clean it up a little bit, but it should be ready for a first review.

Thanks

### DIFF
--- a/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
+++ b/hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
@@ -66,6 +66,10 @@ public class MutinySessionImpl implements Mutiny.Session {
 		return factory.uni( stageSupplier );
 	}
 
+	<T> Uni<T> uni(CompletionStage<T> stage) {
+		return Uni.createFrom().completionStage( stage );
+	}
+
 	@Override
 	public Uni<Void> flush() {
 //		checkOpen();
@@ -258,62 +262,62 @@ public class MutinySessionImpl implements Mutiny.Session {
 
 	@Override
 	public Uni<Void> persist(Object entity) {
-		return uni( () -> delegate.reactivePersist( entity ) );
+		return uni( delegate.reactivePersist( entity ) );
 	}
 
 	@Override
 	public Uni<Void> persist(String entityName, Object entity) {
-		return uni( () -> delegate.reactivePersist( entityName, entity ) );
+		return uni( delegate.reactivePersist( entityName, entity ) );
 	}
 
 	@Override
 	public Uni<Void> persistAll(Object... entities) {
-		return uni( () -> applyToAll( delegate::reactivePersist, entities ) );
+		return uni( applyToAll( delegate::reactivePersist, entities ) );
 	}
 
 	@Override
 	public Uni<Void> persistMultiple(List<?> entities) {
-		return uni( () -> applyToAll( delegate::reactivePersist, entities.toArray() ) );
+		return uni( applyToAll( delegate::reactivePersist, entities.toArray() ) );
 	}
 
 	@Override
 	public Uni<Void> remove(Object entity) {
-		return uni( () -> delegate.reactiveRemove( entity ) );
+		return uni( delegate.reactiveRemove( entity ) );
 	}
 
 	@Override
 	public Uni<Void> removeAll(Object... entities) {
-		return uni( () -> applyToAll( delegate::reactiveRemove, entities ) );
+		return uni( applyToAll( delegate::reactiveRemove, entities ) );
 	}
 
 	@Override
 	public Uni<Void> removeMultiple(List<?> entities) {
-		return uni( () -> applyToAll( delegate::reactiveRemove, entities.toArray() ) );
+		return uni( applyToAll( delegate::reactiveRemove, entities.toArray() ) );
 	}
 
 	@Override
 	public <T> Uni<T> merge(T entity) {
-		return uni( () -> delegate.reactiveMerge( entity ) );
+		return uni( delegate.reactiveMerge( entity ) );
 	}
 
 	@Override
 	public final Uni<Void> mergeAll(Object... entities) {
-		return uni( () -> applyToAll( delegate::reactiveMerge, entities ) );
+		return uni( applyToAll( delegate::reactiveMerge, entities ) );
 	}
 
 	@Override
 	public Uni<Void> mergeMultiple(List<?> entities) {
-		return uni( () -> applyToAll( delegate::reactiveMerge, entities.toArray() ) );
+		return uni( applyToAll( delegate::reactiveMerge, entities.toArray() ) );
 	}
 
 	@Override
 	public Uni<Void> refresh(Object entity) {
-		return uni( () -> delegate.reactiveRefresh( entity, LockOptions.NONE ) );
+		return uni( delegate.reactiveRefresh( entity, LockOptions.NONE ) );
 	}
 
 	@Override
 	public Uni<Void> refresh(Object entity, LockMode lockMode) {
-		return uni( () -> delegate.reactiveRefresh( entity, lockMode ) );
+		return uni( delegate.reactiveRefresh( entity, lockMode ) );
 	}
 
 	@Override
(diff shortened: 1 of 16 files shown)


hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveStatelessSessionImpl.java
```diff
@@ -213,39 +221,41 @@ public void checkTransactionNeededForUpdateOperation(String exceptionMessage) {
...
 	}
 
 	@Override
 	public <T> CompletionStage<List<T>> reactiveGet(Class<T> entityClass, Object... ids) {
-		checkOpen();
-		for (Object id : ids) {
-			if ( id == null ) {
-				return failedFuture( new IllegalArgumentException( "Null id" ) );
+		return serialized( () -> {
+			checkOpen();
```
yrodiere, line 230: Note moving the `checkOpen` call this way means any error reported by this check will be reported in the returned future, instead of being directly thrown by the method. Which I don't think is the intent?

Also... You can probably move the check directly into `serialized`? I'm guessing close to every single call needs it.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveQueryProducer.java
```diff
@@ -42,6 +42,10 @@ public interface ReactiveQueryProducer extends ReactiveConnectionSupplier {
 
 	<T> CompletionStage<T> reactiveFetch(T association, boolean unproxy);
 
+	default <T> CompletionStage<T> internalReactiveFetch(T association, boolean unproxy) {
```
yrodiere, line 45: I'm guessing this was introduced to avoid deadlocks? I.e. address [that comment on the issue](https://github.com/hibernate/hibernate-reactive/issues/2494#issuecomment-3252725205):

> Note the suggested solution could... perhaps... possibly... cause big trouble if some session operations call other session operations. I'm not sure, but it's something to be careful about.

If so, +1. But we likely don't want to expose this as API, which is what you did here.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SessionOperationSerializationTest.java
```diff
@@ -0,0 +1,504 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Timeout(value = 10, timeUnit = MINUTES)
+@EnabledFor(POSTGRESQL)
+public class SessionOperationSerializationTest extends BaseReactiveTest {
+
+	private static final int ENTITY_COUNT = 100;
```
yrodiere, line 40: Was this enough to trigger a problem before your patch?

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SessionOperationSerializationTest.java
yrodiere, line 1: Nice to have a test. What about documentation?

Was there something in the docs saying "you have to wait until an operation is finished before triggering another on the same session"? If so, we should probably soften the language. I still wouldn't recommend doing what we do in this test, but still we can say that calling multiple operations on the same session without waiting for them to finish will simply result in them being executed in the order they were requested, one after the other, and not in parallel.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveSession.java
```diff
@@ -90,6 +91,10 @@ default CompletionStage<Void> reactiveLock(Object entity, LockMode lockMode){
 
 	<T> CompletionStage<T> reactiveGet(Class<T> entityClass, Object id);
 
+	default <T> CompletionStage<T> internalReactiveGet(Class<T> entityClass, Object id) {
+		return reactiveGet( entityClass, id );
+	}
```
yrodiere, line 96: Same comment as elsewhere, this probably shouldn't be API?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/ReactiveSession.java
```diff
@@ -155,6 +160,8 @@ default <T> CompletionStage<T> reactiveFind(Class<T> entityClass, Object id, Loc
 	boolean isDirty();
 	boolean isOpen();
 
+	<T> CompletionStage<T> serialized(Supplier<CompletionStage<T>> operation);
```
yrodiere, line 163: This probably shouldn't be API?

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -197,6 +198,14 @@ private void threadCheck() {
 		InternalStateAssertions.assertCurrentThreadMatches( associatedWorkThread );
 	}
 
+	@Override
+	public <T> CompletionStage<T> serialized(Supplier<CompletionStage<T>> operation) {
+		final CompletionStage<T> result = previousOperation
+				.thenCompose( ignored -> operation.get() );
+		previousOperation = result.handle( (v, e) -> null );
+		return result;
```
yrodiere, line 206: I'd clarify the intent, because without context, this code looks like a very bad attempt at providing thread safety (which it isn't).

```suggestion
		// Make sure a second operation triggered before the first finished can still work:
		// we will only start the second operation when the first one finished.
		// Note we're not using locks because we still don't support parallel calls;
		// just sequential calls that fail to wait for a previous (async) operation to finish.
		final CompletionStage<T> result = previousOperation
				.thenCompose( ignored -> operation.get() );
		previousOperation = result.handle( (v, e) -> null );
		return result;
```

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
```diff
@@ -258,62 +262,62 @@ public <T> Uni<T> find(EntityGraph<T> entityGraph, Object id) {
 
 	@Override
 	public Uni<Void> persist(Object entity) {
-		return uni( () -> delegate.reactivePersist( entity ) );
+		return uni( delegate.reactivePersist( entity ) );
```
yrodiere, line 265: Not sure why you did this.

The previous implementation would execute the operation every time someone subscribes to the uni.

The new implementation will execute the operation once, then return the same result every time subscribes to the uni.

It's a rather big change in behavior, and it seems unwarranted?

yrodiere: Note the fact you're now passing a completion stage supplier to `serialized()` does not change my point above: the completion stage returned by `serialized` will still only execute once.

We *could* take advantage of `serialized()` to remove some of the suppliers, but only if `serialized()` returned a `Uni`. Which I don't think we want to do right now (it's be a very extensive change).

hibernate-reactive-core/src/main/java/org/hibernate/reactive/session/impl/ReactiveSessionImpl.java
```diff
@@ -197,6 +198,14 @@ private void threadCheck() {
 		InternalStateAssertions.assertCurrentThreadMatches( associatedWorkThread );
 	}
 
+	@Override
+	public <T> CompletionStage<T> serialized(Supplier<CompletionStage<T>> operation) {
```
yrodiere, line 202: This solves the problem but introduces two risks:

1. "Too much serialization": An operation that is serialized depends on a sub-operation that will itself be serialized -- creating a cyclic dependency, resulting in a deadlock.
2. "Too little serialization": A session operation forgets to call `serialized()`, resulting in that specific operation potentially running concurrently to another (serialized) one.

I think risk 1 is probably already addressed by tests: any code path that relies on that sub-operation would deadlock, which would fail our tests, so assuming sufficient coverage we'll likely find all such problems.

The second risk is harder to address. Right now I don't have an idea... Static checks, maybe? I suppose we could have some automated rule that checks every `Session` method returning a completion stage wraps it using `serialized`. It might be more trouble than it's worth, though.

hibernate-reactive-core/src/main/java/org/hibernate/reactive/mutiny/impl/MutinySessionImpl.java
```diff
@@ -66,6 +66,10 @@ <T> Uni<T> uni(Supplier<CompletionStage<T>> stageSupplier) {
 		return factory.uni( stageSupplier );
 	}
 
+	<T> Uni<T> uni(CompletionStage<T> stage) {
+		return Uni.createFrom().completionStage( stage );
```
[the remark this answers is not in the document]
yrodiere: Good bot. I am Yoann and I approve this message.

hibernate-reactive-core/src/test/java/org/hibernate/reactive/SessionOperationSerializationTest.java
DavideD: Good point, I will update it

PR #4168, opened 2026-09-18
Update JDKs for workflows

DavideD:


.github/workflows/build.yml
```diff
@@ -255,9 +255,9 @@ jobs:
           # even if we don't use these features, just enabling them can cause side effects
           # and it's useful to test that.
           - { name: "21", java_version_numeric: 21, jvm_args: '--enable-preview' }
-          - { name: "25", java_version_numeric: 25, jvm_args: '--enable-preview' }
```
marko-bekhta, line 258: maybe let's keep 25 as it is an lts ?

DavideD: I don't mind but do we still need to test with 25? Aside for the 21, I usually just keep the latest 3.

DavideD: Nevermind, let's keep the 25

DavideD: > maybe let's keep 25 as it is an lts

Better to write it as LTS, I couldn't understand what ITS was :D

marko-bekhta: 🙂 yes! sorry 🙈 I'll keep that in mind for the next time 🙂

